PR #37, opened 2017-07-09
Added Italian translation

JackHack96:
Hi! I've added an Italian translation for PulseEffects.

It would also be nice to have a tutorial about PulseEffects 👍 (maybe I could make one, when I'll finish with summer exams)


wwmm:
Thanks!

It would be good to have some kind of tutorial. Right now I am lacking imagination to do that so feel free to add one when you have time :-)

PR #57, opened 2017-08-25
Add Meson support and improve translation system

AsavarTzeth:
Adding Meson support greatly simplifies the entire build process and it kind of is a 
requirement for a simplified and streamlined workflow for the translators.

The only consequence of the change is that new languages must be added to
`po/LINGUAS` for Meson to install it. However, it will overall still be simpler. This
can easily be added to the wiki (it will be).

AppStream appdata is also included and fully integrated with the existing translating
system. As is the renamed desktop entry file. Now all translations can happen in one
file and the same format.


AsavarTzeth:
For your convenience, here is a modified PKGBUILD that you can base your own on:

```
# Maintainer: Wellington <wellingtonwallace@gmail.com>
pkgname=pulseeffects
pkgver=2.3.3
pkgrel=1
pkgdesc="Limiter, compressor, reverberation, stereo equalizer and auto volume effects for Pulseaudio applications"
arch=(any)
url="https://github.com/wwmm/pulseeffects"
license=('GPL3')
depends=(python python-gobject python-cairo gtk3 gst-plugins-good gst-plugins-bad
         gst-python swh-plugins python-numpy python-scipy)
makedepends=('python-setuptools' 'meson')
options=(!emptydirs)
source=("$pkgname-$pkgver::https://github.com/wwmm/pulseeffects/archive/v$pkgver.tar.gz")
sha256sums=('')

build() {
  cd "$srcdir/$pkgname-$pkgver"
  # Remove any potential residual build files
  rm -rf _build
  meson _build --prefix=/usr
}

package() {
  cd "$srcdir/$pkgname-$pkgver/_build"
  env DESTDIR="$pkgdir" ninja install
}

# vim:set ts=2 sw=2 et:
```

wwmm:
Hi! Many thanks for adding support for meson. Now I will finally learn how to use it! :-)

Is there still need for setuptools files like setup.py and the manifest?

AsavarTzeth:
As far as I know the manifest is not needed. As for for `setup.py`, you can safely remove it. I only left it because I did not know if you wanted to keep it.

wwmm:
Ok,  I will remove them. 

Do you know a meson tutorial for the kind of things you implemented in PulseEffects?

Thanks for the example pkgbuild :-)

AsavarTzeth:
The official documentation is quite good.
http://mesonbuild.com/Quick-guide.html

Now the bigger problem is with the translations. I actually kind of cheated and added everything manually and tested that it works. I am unsure what tools could be used to automate updates of the POT file.

Maybe you could ask someone like @TingPing. I got my own inspiration from his projects.

Here is the official AppStream documentation on how to integrate translations. Unfortunately it does not include Meson, nor does it have examples for everything.

https://www.freedesktop.org/software/appstream/docs/sect-Quickstart-Translation.html

wwmm:
Thanks! :-)

TingPing:
> I am unsure what tools could be used to automate updates of the POT file.

`ninja PulseEffects-pot`

Also I think you set the wrong name there you probably wanted lower-case for the project name. You also generally don't store `.pot` files in git.

meson.build
```diff
@@ -0,0 +1,23 @@
+project('PulseEffects',
+  version: '2.3.3'    
+)
+
+py3_mod = import('python3')
+i18n_mod = import('i18n')
+
+message('Checking dependencies')
```
TingPing: I think these messages are pretty silly, its not like any of these actions can fail.

wwmm:
Hi TingPing. I don't  have much experience with translations. If I remove the .pot file from the project how would translators proceed to get a .pot file? Would they run `ninja PulseEffects-pot` or build the project with meson and then search for the .pot file in _build ?

TingPing:
@wwmm So in 2017 the common thing is to have a third party translation service like transifex.com or zanata.org where you upload the .pot file, translators translate things in a webui and then you download the .po files.

If you really want to avoid these services and you want people to just throw .po files at you via github then maybe it makes sense to keep it in tree but I think this is a less friendly method in general.

meson.build
AsavarTzeth: Well it is my first time using Meson. I did not fully understand what the messages was for. Part of me had them there to act as comments.

AsavarTzeth:
I agree actually. Most of my time spent as a translator in the past has been via these kinds of services. The user experience is similar to Poedit, but also brings instant online collaboration. It also does solve the need to have a GitHub account and doing a pull request.

PR #58, opened 2017-08-26
Add some dependency checking to build

AsavarTzeth:
Meson now has some dependency checks. Unfortunately not all dependencies could get added.

This is due to the fact that Meson uses pkg-config and not all projects ship the required files. Additionally it lacks a simple way of checking for python modules, which naturally do not ship these files.

The main net positive is that the build system clearly checks your gstreamer version, which should prevent any future issue threads regarding this.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -8,6 +8,13 @@ i18n_mod = import('i18n')
 message('Checking dependencies')
 py3 = py3_mod.find_python()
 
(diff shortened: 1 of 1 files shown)


AsavarTzeth:
Dependency checking in Meson is based upon pkg-config. To get the names of the dependencies you need to specify in the `meson.build` file you may run:

```
pkg-config --list-all
```

I recommend using `grep -i` to filter it down.

PR #62, opened 2017-09-01
Fix translations not working in a prefixed installation

AsavarTzeth:
Localisations (translations) would not work when PulseEffects was installed into a /app prefix (or any other for that matter). The issue is that `gettext` localisations were not properly initialized and thus it would always use the default locale directory: `/usr/share/locale`.

*Note that is the python gettext module default, not the original gettext.*

It took a lot of research and reading and in the end all documentation is written with C in mind. Documentation often also assumed that GNU Autotools was the build system. The main point though was that the recommended way of providing a different locale directory is via the build system.

In the end I found a method, which works. It does not feel completely proper, but I could not find another way of solving this. Other methods would mean either hardcoding a list of often used directories or a complicated search function (possibly based on xdg-data-dirs). Not sure, but the latter would possible become an attack vector.

Credit goes to the [Piper Project](https://github.com/libratbag/piper) for part of the solution.

There are also a few other optimizations in this pull request. I found them as I was initially hacking on the code.


AsavarTzeth:
Closing this for now. I will focus on the bigger translation update issue first. I will probably restructure this pull request as well.

PR #65, opened 2017-09-02
Fix i18n support

AsavarTzeth:
This fixes internationalization support when PulseEffects is installed in a prefix, as is the case with the Flatpak. Without it translations will only work if localization files are installed under the python `gettext` default locale dir: `/usr/share/locale`

See commit messages for more details.


AsavarTzeth:
I also have an idea as to how you may be able to do a bug fix release now that the master branch is already in use for the other changes. I would try:

1. Checkout v.2.3.4
2. Make a new branch based on that, call it maintenance or something; does not matter
3. Push that branch to GitHub
4. Set that branch as the target for this pull request
5. When you feel ready make a bug fix release based on this new branch

I just patched these changes in myself downstream. During testing I did something similar to this and I don't think there were any merge conflicts. If there was it is really minimal.

After all of this you may also merge the new maintenance branch into master.

wwmm:
Hi,

I released a bugfix 2.3.5 release based on 2.3.4 and these commits. No problems with Arch pkgbuild process and everything is working fine here. I hope the same goes with the flatpak package.

AsavarTzeth:
Strange, I did not have any merge conflicts. Maybe it would have worked better with `git cherry-pick` which I did use initially.

Anyhow the end result looks and works well and the Flatpak which has been updated to use v2.3.5.

I also noticed that you initially missed some parts of the merge conflict. In case you are handling these manually, might I make a suggestion? I found using [Meld](http://meldmerge.org) to be a great tool for this. It has a special mode for version control just for these situations.

wwmm:
Feel free to do any suggestion :) I always used git on my own in private repositories and this is the first one where I actually interact with other people. So there is much I don't know. Specially about localization.

I cherry picked the commits because using the merge on v2.3.4 would add the new pipeline code(or at least did when I tested it following the manual instructions that github showed). And as you were having noise with it I decided it would not be a good idea to do a bugfix release with new code.

AsavarTzeth:
Well you still use git, you just don't handle the merge conflicts in a text editor. So if anything you replace the text editor with Meld only while in the middle of a merge conflict.

You can see this [branch](https://github.com/AsavarTzeth/pulseeffects/tree/backport), which is where I did it without conflicts. That branch is based on 803b917c197e120be7405d3d1a4cec928233b708, then I think I cherry picked commits like you did and had no conflitcts. Although the order did matter; naturally I had to do the oldest first.

Now I don't feel too comfortable with the more advanced features of git. I should probably play around more and fail a lot on purpose.

AsavarTzeth:
By the way, if this is the first time you handle localization's I would actually say you are doing quite well. Some people just don't care at all.

PR #66, opened 2017-09-02
Add croatian translation

muzena:


wwmm:
Thank you for the translation :-)

po/hr.po
```diff
@@ -0,0 +1,279 @@
...
+#: data/com.github.wwmm.pulseeffects.desktop.in:4
+msgid "Audio Enhancer"
+msgstr "Zvučni poboljšivač"
+
+#: data/com.github.wwmm.pulseeffects.desktop.in:6
+msgid "limiter;compressor;reverberation;equalizer;autovolume;"
+msgstr "graničnik;kompresor;odjek;ekvalizator;automatska glasnoća zvuka;"
+
+#: data/com.github.wwmm.pulseeffects.desktop.in:9
+msgid "pulseeffects"
```
AsavarTzeth: Just an heads up, that last "pulseeffects" string is the icon name and must not be translated. Currently I don't know how to exclude this. We might have to do something similar to [this](https://github.com/gnome-mpv/gnome-mpv/blob/master/data/io.github.GnomeMpv.desktop.in#L6)

po/hr.po
```diff
@@ -0,0 +1,279 @@
...
+msgstr "pulse efekti"
```
AsavarTzeth: Just an heads up, that last "pulseeffects" string is the icon name and must not be translated. Currently I don't know how to exclude this. We might have to do something similar to [this](https://github.com/gnome-mpv/gnome-mpv/blob/master/data/io.github.GnomeMpv.desktop.in#L6).

muzena: @AsavarTzeth 
Yes this icon names is a lot in GNOME modules in po/pot files, they just add comment to not translate like this in gnome-mpv

PR #69, opened 2017-09-03
german translation update

Bleuzen:
I 'translated' some things back to englisch, because I don't find a 'perfect' german translation for it. We have also some english words / vocabulary in german that we use.
So I thought again about the translation and optimized it.

I also translated for example "Window" back, because this was shown in the Auto Volume section and the other parameters are not translatable. (Because they are hard coded strings in the UI?)
In my opinion this looks bad, when in a row 1 of 4 words is german and the other 3 are english, so thats why I let it in english.

But I translated the main things and this should be 'ok' and make PulseEffects 'usable' for german users.

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -8,7 +8,7 @@ msgstr ""
 "Project-Id-Version: PulseEffects\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2017-09-02 12:17-0300\n"
-"PO-Revision-Date: 2017-09-02 21:09+0200\n"
+"PO-Revision-Date: 2017-09-03 14:46+0200\n"
 "Language-Team: \n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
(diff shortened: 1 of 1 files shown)


wwmm:
There shouldn't be untranslatable strings in the interface. I will check if there was a regression

wwmm:
I forgot to add a few files to POTFILES.in. I will add them and update the template

wwmm:
In other to support turning plugins on/off I did a bid refactoring and in this process each plugin gained its own glade file. Of course I forgot to add all these new files to the POTFILES.in XD It would be really nice if we could add a whole folder as argument in this file. Everything that is in the ui folder have to be translated.

wwmm:
Talking about test how can I test translations when running from the source code? I think I saw a way to specify the `.mo` path when I first added localization support but I can't remember how to to that. THis is the first project where I have to think about localization and it is being a little hard for me to know if I am breaking things or not

Bleuzen:
@AsavarTzeth How can I test it? Is there a better way then moving the .mo file to /usr/share/locale ... ?

AsavarTzeth:
It depends on where you have installed it. If you installed it from source, using Meson and did not change any default values, then you should install the .mo files under `/usr/local/share/locale/language_code/LC_MESSAGES/PulseEffects.mo`

PR #92, opened 2017-10-14
i18n and appdata updates

AsavarTzeth:
Some updates and fixes to translations and appdata. I thought I would do this now, so it is ready for any upcoming release.


wwmm:
I did not know about `PulseEffects-update-po`. It is definitely a good idea to run it. Thanks!

I don't know if there is anything we can do about the window aspect ratio.

AsavarTzeth:
The aspect ratio screenshot recommendation is probably the most annoying one. If you are interested you can find the relevant AppStream recommendations [here](https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html#qsr-app-screenshots-info).

On occasion I do get lucky and the app is close enough to 16:9 by default.

Sometimes the app will remember size changes on exit. In order to support this it usually stores width and height in dconf or a config file. Then I can simply change those values manually and restart the app.

I guess that is one approach that could work. I mostly mentioned it in the commit message for documentation purposes though. For example if anyone else would update the screenshots in the future, they would hopefully check the last commit (one can hope).

Issue #30, opened 2017-06-26
Better installation instructions for Ubuntu 


thenewnano:
I know this is not an issue but would save me almost an hour if it was added into the readme
`git clone https://github.com/wwmm/pulseeffects.git`
`cd pulseeffects`
`sudo python3 setup.py install`
`sudo cp -r share/* /usr/share/`
`sudo glib-compile-schemas /usr/share/glib-2.0/schemas/`

probably a  sudo apt install "the list of the dependencies" if they are  missing.

wwmm:
Hi I have updated the readme with these steps.

thenewnano:
I just installed this on my main DJing computer, new ubuntu installation and had to check around for all dependencies, this can be added to the guide also recommendation to run pulseeffects from terminal as it will throw error messages that are helpful, I was missing swh-pugins at the last step somehow :
 
sudo apt install swh-plugins gir1.2-gst-plugins-bad-1.0 python-cairo python-gobject python-gobject-2 ladspa-foo-plugins python3-cairo python3-scipy python3-numpy python3-nameparser

wwmm:
Hi,

wwmm:
Hi

As I do not use ubuntu it is complicated for me to know all the details involved to have PulseEffects running in this system. I also don't know if the readme is the best place for a step by step specific for each distro. Would the wiki page be a good place for this? Maybe with a link to it in the readme... In any case I never used this wiki function that github offers so I do not know how good it is for a collaborative guide writing

thenewnano:
@wwmm  I was intending to write there, I mean even if you use Ubuntu you still can't cover all other distros, and distros upgrade and stuff change, I read about your application in OmgUbuntu blog, but still I would like you to enable the Wiki and we users will maintain it. 
swh-plugin for instance is something I saw in your comments for the Gentoo users, so all distros somehow have cross connections and in a Wiki page we will sort it out. 
I know many people use other solutions for DJing but I still prefer the simplicity of PulseAudio as the defauld soluion in my distro than JACK and other solution, I just don't need that performance threre.

thenewnano:
Maybe you open up the wiki for public and see if it works itself out https://help.github.com/articles/changing-access-permissions-for-wikis/

wwmm:
Ok.  I changed the wiki permissions. It should be public now :-)

PR #107, opened 2017-11-06
Add Debian/Ubuntu package building specs and instructions

mikhailnov:


mikhailnov:
Sorry for the mess in commits and branches, I'm new to git((
So, this pull request:
1) adds debian package building specs
2) adds information about compiling and installing PulseEffects and about build the deb package to README.md
3) adds Russian localisation of this program

wwmm:
No problems, I am not a git guru :-)

I think that distribution files like the ones in the debian folder should not be in this git repository but in one in the distribution site like Arch Linux does. But for now it can be here

README.md
```diff
@@ -36,25 +36,39 @@ output at the same time it applies them for applications output:
...
 ### Flatpak
 
 [Flatpak](https://flatpak.org) packages support multiple distributions and are sandboxed.
 
 Stable releases are hosted on [Flathub](https://flathub.org):
 
 ```
-flatpak remote-add --if-not-exists flathub https://flathub.org/repo/flathub.flatpakrepo
-flatpak install flathub com.github.wwmm.pulseeffects
+$ flatpak remote-add --if-not-exists flathub https://flathub.org/repo/flathub.flatpakrepo
```
AsavarTzeth: Any reason you added $ here but not on your own additions?

README.md
```diff
@@ -36,25 +36,39 @@ output at the same time it applies them for applications output:
...
+$ sudo ninja install
+```
+
 ### GNU/Linux Packages
 
 - [Arch Linux](https://aur.archlinux.org/packages/pulseeffects/)
 
+Ubuntu >=17.10 repository:
+```
+wget -q -O- http://repo.dumalogiya.ru/keys/mikhailnov_pub.gpg | sudo apt-key add -
```
AsavarTzeth: If $ is to be used, did you miss it here?

README.md
```diff
@@ -36,25 +36,39 @@ output at the same time it applies them for applications output:
...
 10. Spectrum Analyzer (Gstreamer)
 
-## Installation
+## Build and install
 
 Users upgrading from 1.x to 2.x will have to rebuild their presets. Since
 version 2.0.0 PulseEffects uses a different format. This
 change was necessary to support presets for microphone processing.
 
+```
```
AsavarTzeth: I fundamentally disagree with adding this information to the top and mixing build and regular install instructions. Issues with this section and why it should  be removed:

- We should not expect most users to build and install this.
- This replicates information already present in the wiki.
- This information is incomplete, lacking instructions for installing dependencies etc which, again, is already present in the wiki.

Minor note for the future, in English grammar the title should be "Build and Install".

AsavarTzeth:
A lot of the additional documentation should have been added to the wiki. This goes against earlier changes I made based on comments you made in #30.

I recommend removing most of the changes to the docs here and adding it to the wiki. Let's keep it consistent.

See my code comments for more issues. In the future, feel free to use me as a reviewer for documentation, grammar and build system changes. Basically anything that isn't distribution specific packaging or the PulseEffects code base.

README.md
```diff
@@ -36,25 +36,39 @@ output at the same time it applies them for applications output:
...
+$ meson _build --prefix=/usr
+$ cd _build
+$ sudo ninja install
+```
+
```
AsavarTzeth: I fundamentally disagree with adding this information to the top and mixing build and regular install instructions. Issues with this section and why it should  be removed:

- We should not expect most users to build and install this.
- This replicates information already present in the wiki.
- This information is incomplete, lacking instructions for installing dependencies etc which, again, is already present in the wiki.

Minor note for the future, in English grammar the title should be "Build and Install".

AsavarTzeth:
By the way, if you want to decrease the mess of commits in the future you can simply squash them. See the [Git documentation](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History) if you wish to learn.

Rewriting history is fine as long as you have not shared your current commits or already merged them. Basically, you are best off squashing them locally on your machine, then pushing it to GitHub.

README.md
mikhailnov: I wanted to show that the commands are expected to be ran not under root, but did not follow the same style in all places

mikhailnov:
Having package building specs in the main repo is very convenient to my mind. For example, I want to make some changes into the program or create a fork, but I dont want to create a mess in my OS, so I just take the specs and rebuild the package.
Having Flatpk specs here would also be great, because at the moment I have to package PulseEffects for Ubuntu 16.04 which does not meet several depencies' versions, but PulseEffects has to be with the Russian localisation that I've added to there, and now I have to look for Flatpak rebuilding instructions and sources, manually put other PulseEffects sources there etc.

README.md
AsavarTzeth: Interesting is that what the dollar sign means in these instances? I am quite experienced with Linux and Open Source and have seen the $ sign used but never knew why. I assumed it was there to make it harder for people to just copy/paste, since that can be dangerous.

Is there any documentation, or other reading material I have missed, that you can link to on this?

Of course you are right in this case then, since flatpak is best used without root or sudo.

AsavarTzeth:
Just in case there was a misunderstanding. I have nothing against the adding of the "building specs". If @wwmm accepts it, I obviously agree. I simply try to maintain clarity and consistency in the docs and the wiki since I have done a lot of work on that previously.

Regarding the [flatpak manifest](https://github.com/flathub/com.github.wwmm.pulseeffects/blob/master/com.github.wwmm.pulseeffects.json), we could possibly add it as a submodule in this repo, but then it would have to be updated every time it changes on flathub. Another way is of course improving documentation or adding information in the wiki. Simply adding the file is not an option since it is bound to be outdated quickly.

mikhailnov:
$ is a standart command prompt fot non-root users and # is for root in bash and other shells.
Just execute 
``` sudo -i ```
You login prompt will be changed to # from $.

So, if the command starts with #, users understand that the command is expected to execute as root.

mikhailnov:
We can probably make a link from this repo to https://github.com/flathub/com.github.wwmm.pulseeffects as a Git submodule (I have never used them) 
https://git-scm.com/book/en/v2/Git-Tools-Submodules

README.md
mikhailnov: $ is a standart command prompt fot non-root users and # is for root in bash and other shells.
Just execute 
``` sudo -i ```
You login prompt will be changed to # from $.

So, if the command starts with #, users understand that the command is expected to execute as root.

AsavarTzeth:
Now that you mention it, that prompt seems obvious. I will try and remember that for the future. I strongly suspect that a lot of people are unaware of this so in some situation it is probably a good idea to complement with written instructions.

Imagine git submodule like a repo in your repo, or like a hyperlink to another repo. It is great if you need to include other sources that may have incompatible licenses. That works because it is just a link and not actually part of your repo. Whenever the main repo is cloned you have to pull down all submodules too.

The bad news in this case is that the link is locked to a specific commit. You have to pull all changes manually inside the submodule directory commit it to the main repo. Since I don't have commit rights here, there is no way for me to ensure it stays up to date.

mikhailnov:
Yes, I've already understood it and deleted the comment about git submodules

wwmm:
I see no problems in keeping build specs here. I only would like not to see many folders like "debian", "opensuse", "fedora", "gentoo", etc in the repository root. As they are not part of PulseEffects code it would be confusing. If they are all inside a folder with a name like `distributions packages` I see no problems in having them here.

AsavarTzeth:
That is is generally my opinion as well. Based on my slightly outdated Debian packaging skills though, I do believe the native Debian packaging tools expect it to be named "debian" and located in the root. There are most likely ways to work around this.

A common directory name for the purposes you mentioned is `dist`, as a short.

wwmm:
I think that in this case the user could just move it to the root before starting the build.

`dist` is a better name indeed.

mikhailnov:
Made a PPA https://launchpad.net/~mikhailnov/+archive/ubuntu/pulseeffects and updated the wiki page https://github.com/wwmm/pulseeffects/wiki/Package-Repositories#debian--ubuntu

PR #110, opened 2017-11-07
Fix issues with documentation

AsavarTzeth:
The recent introduction of debian build files also modified the
documentation in less than ideal ways.

This fix solves this by staying consistent with previous changes that
moved distribution specific instructions to the wiki. Specifically a
wiki page for community maintained distribution packages is introduced
and linked to.

Build instructions is removed altogether since it has long since had a
dedicated wiki page. Some instructions were incomplete anyhow or located
in the wrong section.

Showing incomplete build instructions at the top is bad, because it
decreases the likelihood of a user reading the complete wiki page, thus
missing information. This in turn leads to an increase in support issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,29 +36,23 @@ output at the same time it applies them for applications output:
 9. Output Limiter (Ladspa Fast Lookahead Limiter)
 10. Spectrum Analyzer (Gstreamer)
 
-## Build and install
+## Installation
 
 Users upgrading from 1.x to 2.x will have to rebuild their presets. Since
 version 2.0.0 PulseEffects uses a different format. This
 change was necessary to support presets for microphone processing.
 
-```
-$ meson _build --prefix=/usr
-$ cd _build
-$ sudo ninja install
-```
-
 ### GNU/Linux Packages
 
 - [Arch Linux](https://aur.archlinux.org/packages/pulseeffects/)
 
-Ubuntu >=17.10 repository:
-```
-wget -q -O- http://repo.dumalogiya.ru/keys/mikhailnov_pub.gpg | sudo apt-key add -
-echo "deb http://repo.dumalogiya.ru/aptly/public artful main" | sudo tee /etc/apt/sources.list.d/dumalogiya-artful.list
-sudo apt update
-sudo apt install pulseeffects
-```
+#### Community Packages
+
+These are community maintained repositories of distribution packages. You can
+find more information about these in the
+[wiki](https://github.com/wwmm/pulseeffects/wiki/Package-Repositories#package-repositories).
+
+- [Ubuntu](https://github.com/wwmm/pulseeffects/wiki/Package-Repositories#ubuntu-1710-and-newer)
 
 ### Flatpak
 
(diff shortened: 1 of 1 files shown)


AsavarTzeth:
I you would like, feel free to mention me in any pull request you would like help with a review of changes to the documentation or the build system. I would personally much prefer finding issues and solving them before they are merged.

wwmm:
Thanks. When I reviewed I had all my attention focused on possible code changes and I did not give the needed attention to documentation changes. From now on I will mention you in this kind of pull requests :-)

mikhailnov:
I think that the information about Flatpak packages should also be added to https://github.com/wwmm/pulseeffects/wiki/Package-Repositories

mikhailnov:
Added instructions how to build a Debian pckage: https://github.com/wwmm/pulseeffects/wiki/Installation-from-Source

mikhailnov:
Added Flatpak to
https://github.com/wwmm/pulseeffects/wiki/Package-Repositories
https://github.com/wwmm/pulseeffects/wiki/Installation-from-Source

I will add Flatpak rebulding instructions after I manage to do it myself (or maybe you can add them?)

AsavarTzeth:
I did consider adding the detailed information about the Flatpak packages to the wiki but decided not to. At least not to that page, which I made with a specific use case in mind.

My reasoning is this:

- The purpose of moving parts to the wiki is to keep the README minimal. As part of the original cleanup I moved distribution specific instructions to the wiki.
- Flatpak is not distribution specific so I feel we can spare some precious space for it. It is only an issue when we start doing it for every distribution. If we accept one distribution specific way it sort of obligates us accept all others.
- Flatpak is quite new and a lot of people are unfamiliar with it. Having a quick overview on the homepage helps mitigate this. As some issue threads reveal, not everyone thinks of checking the wiki. As some issue threads reveal, not everyone thinks of checking the wiki.
- At best we are talking about at best saving 2-3 lines of space, just for a case of potential correctness. Splitting the information from the section in the README will probably do more harm than good. If the installation procedure required a lot documentation I would reconsider this point.

Further more, I made that new wiki page with traditional third-party distribution repositories in mind. Some of the security risks and the possibility of a bad package affecting the rest of the system does not apply to Flatpak.

AsavarTzeth:
Actually while I am the maintainer the flatpak build is provided by Flathub. It is they who build and distribute it and it is them the user trusts. Also see above why I don't think it belongs, at least on that page.

AsavarTzeth:
The flatpak build process differs, but the best way is probably to use `flatpak-builder` which is documented [here](http://docs.flatpak.org/en/latest/flatpak-builder.html). It is quite extensive and we should not replicate that here. Linking to the the official docs, is probably wiser.

AsavarTzeth:
Okay, I expanded the [Flatpak section](https://github.com/wwmm/pulseeffects/wiki/Installation-from-Source#flatpak) in the Installation-from-Source wiki page. I think it is clear enough, but since I am somewhat biased on this topic, I could be wrong.

There is a fine balance between clarity and not reproducing documentation to be kept though.

mikhailnov:
Yes, I did not mean moving it away from README. That page is about repositories, and if the user wants to get PulseEffects easily and quickly installed to his system, he goes go that wiki page. In many cases both Flatpak and not Flatpak packages may satisfy him.
Removing Flatpak from README will be bad, and not mentioning it on the page about repositories breaks the logic, because not every user will read the README up to the end.

mikhailnov:
Thank you, now I briefly understand how to rebuild the flatpak package, that's what I wanted and missed.
Added information about cloning the git repo with flatpak sources: https://github.com/wwmm/pulseeffects/wiki/Installation-from-Source

mikhailnov:
Does Flathub automatically rebuild the flatpak package when a new release of PulseEffects gets available on Github?

AsavarTzeth:
Not by itself no. I need to update the release tag and commit and push it. Flathub will then re-build any module that has changed.

That isn't much work though. Two line changes, a commit and push.

PR #114, opened 2017-11-08
Debian preparation

mikhailnov:
Add debian/watch to make Debian building system to be able to track new releases automatically


mikhailnov:
oh, it seems that I've understood how to split different changes to different pull requests. Will try to do it properly next time.

PR #115, opened 2017-11-08
Update hr.po

muzena:


muzena:
@wwmm 
Can you somehow fix this, on Applications section, I think it is caused by 'Promjena frekvencije' and 'Latencija međuspremnika' the distance between the translated strings (Format, Frekvencija, Kanali...) is too large, and this cause too large Pulseaudio window, after window resize it takes half of the screen.
Like in picture:
slika zaslona 2017-11-08 15-40-16

wwmm:
The gtk grid had its columns set to homogeneous. This makes them have the same width. I made some changes to the interface and hopefully this will not happen anymore.

PR #116, opened 2017-11-08
fix Debian dependencies - 2 

mikhailnov:
additional fixes for https://github.com/wwmm/pulseeffects/issues/111

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,9 @@
+pulseeffects (3.0.7.3) artful; urgency=high
+
+  * Additional fixes for https://github.com/wwmm/pulseeffects/issues/111
+
+ -- Mikhail Novosyolov <mikhailnov@dumalogiya.ru>  Mon, 08 Nov 2017 22:21:00 +0300 
+ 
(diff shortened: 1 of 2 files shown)


AsavarTzeth:
May I suggest something? At this rate, perhaps it is better that you wait longer and do more testing in your fork before you do a pull request.

No offence meant, but I just don't want to see @wwmm spending too much time managing pull requests and issues related to this.

mikhailnov:
yes, I agree with you. Arch dependencies confused me because those packages are splitted to many more in Debian.

AsavarTzeth:
There is no shame in making a misstake or being confused :)

Just take your time. It is always better to catch mistakes or issues earlier and get it right. It will save not only your time, but others as well and serve you well in the long run.

Another suggestion is, when you get the time, to perhaps try and get this into Debian. That way you will get advice and reviews by the Debian community in the process. I am sure they will be able to offer a great deal of help.

mikhailnov:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=881224 (I tried submitting it yesterday, but Debian bug submitting system did not accept it for some reasons)

PR #117, opened 2017-11-08
Fix debian build and runtime dependencies

mikhailnov:
https://github.com/wwmm/pulseeffects/issues/111


AsavarTzeth:
I tested this myself on Ubuntu 17.10 and there are some minor issues that should probably be fixed but it does build and install correctly. At this point it does launch without dependency issues at least.

What is your plan for the changelog versioning I wonder? I would personally fix it right away, but I realize it will affect existing installs. There should not be too many yet though.

mikhailnov:
Which issues did you find?
My plan is to have debian/changelog versioning the same as the upstream's

PR #126, opened 2017-11-22
add more sliders and scroll functionality to equalizer

swag4swag:
Some of my EQ presets (migrating from Peace/APO) use 20-25 bands. However, just adding an extra 15 bands would destroy usability on small screens. I attempted to improve usability on small screens and add more bands by putting the GTKBox in a scrolling viewport. The window's width can be made many pixels smaller despite the extra EQ bands.


wwmm:
Hi! Thanks for your contribution :-) I see no problems in merging it. I will just need some time to do a few tweaks over it and then test. With 30 bands the default frequencies will have to be updated/extended. And on the load presets class it will be necessary to add fallback values to all the new parameters. Otherwise people presets will be broken as they do not have those keys in their file.

swag4swag:
Thanks a lot! Looking forward to it.

wwmm:
I have just merged it. Please test it and let me know if there is any problem :-)

PR #132, opened 2017-11-30
Available in gentoo

soredake:


wwmm:
@AsavarTzeth

AsavarTzeth:
The link is for an official distribution package, so it looks good to me.

PR #139, opened 2017-12-13
update debian/chengelog with newer versions

mikhailnov:
also added the result of running ```mk-build-deps --install debian/control``` to .gitignore


mikhailnov:
\+ Fix https://github.com/wwmm/pulseeffects/issues/137

wwmm:
In case you have not done it yet add Calf Studio plugins and the Lilv library as dependencies. They are not mandatory for PE to work but users wanting to try the exciter and the bass enhancer will not be able to do so without these packages

mikhailnov:
Ok. calf-plugins and liblilv-0-0 will be added.

wwmm:
It is not the calf that uses ladspa but the one that uses LV2. This is the package in Arch Linux [https://www.archlinux.org/packages/community/x86_64/calf/](https://www.archlinux.org/packages/community/x86_64/calf/)

PR #151, opened 2018-01-20
New dependencies for debian/control

mikhailnov:
Added zam-plugins for the maximizer plugins, but I still get 
```00:19:02.249 - PulseEffects - WARNING - Stereo Spread plugin was not found. Disabling it!```
calf-plugins is a dependency, but it's version is 0.60 in Debian/Ubuntu vs 0.90 in Arch, is that plugin really from this package of a newer version? What is the file name of this plugin?
If yes, than it probably cannot be fixed easily right now and you can merge it, if there are other ideas, please don't merge it right now.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,8 +1,9 @@
-pulseeffects (3.1.6) artful; urgency=low
+pulseeffects (3.1.6-2) artful; urgency=low
 
   * update to version 3.1.6 upstream
+  * add zam-plugins as a new dependency
   
- -- Mikhail Novosyolov <mikhailnov@dumalogiya.ru>  Tue, 14 Dec 2017 20:02:00 +0300 
+ -- Mikhail Novosyolov <mikhailnov@dumalogiya.ru>  Sun, 21 Jan 2018 00:27:00 +0300 
 
 pulseeffects (3.1.3-3) artful; urgency=low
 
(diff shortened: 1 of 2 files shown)


mikhailnov:
Also tried ```lives-plugins``` https://packages.ubuntu.com/search?searchon=contents&keywords=spread&mode=exactfilename&suite=artful&arch=any but  it does not help (however, they may include the needed plugin but with another name, as a variant)

mikhailnov:
PPA https://launchpad.net/~mikhailnov/+archive/ubuntu/pulseeffects should be online soon.

wwmm:
Remove the GStreamer cache in `~/.cache/gstreamer-1.0`. After that lv2 plugins will be found. I reported this bug to GStreamer developers and it should be fixed in GStreamer 1.13.1

mikhailnov:
It did not help

wwmm:
It is possible that the multispread plugin (stereo spread) have been added in the latest calf. I think I saw this somewhere

wwmm:
According to their changelog it was added in the 0.90 version [https://github.com/calf-studio-gear/calf/blob/master/ChangeLog](https://github.com/calf-studio-gear/calf/blob/master/ChangeLog)

mikhailnov:
ok

PR #165, opened 2018-02-13
script to build and publish packages for the PPA

mikhailnov:
I suggest to have debian/build-ppa.sh upstreamed to make it easy for a new person to maintain the PulseEffects repository in case I stop doing it for some reason. Automation of publishing packages on Launchpad is not as obvious as it could be. Probably, it can also be used for mainline Debian packages.

There was a PPA with Pulseffects http://www.omgubuntu.co.uk/2017/06/install-pulse-effects-ubuntu-ppa but then the maintainer stopped updating it

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ debian/*
 !debian/pycompat
 !debian/rules
 !debian/source/format
+!debian/build-ppa.sh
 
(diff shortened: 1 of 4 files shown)


mikhailnov:
Also updated the link in Readme

PR #175, opened 2018-02-22
Emphasize that PulseEffects is equalizer

teneri66:
It might be not convenient for the ordinary user to find pulseeffects application from KDE/Gnome. I suggest we emphasize that PulseEffects is equalizer mostly. Eventually we will benefit once PulseEffects becomes a part of default set of applications installed with major distributions. Users will try to type in "equalizer" and will be able to find PulseEffects because "PulseEffects" is too technical (relates to pulseaudio), not everyone will know it / remember it.

### DIFF
--- a/data/com.github.wwmm.pulseeffects.desktop.in
+++ b/data/com.github.wwmm.pulseeffects.desktop.in
@@ -1,6 +1,6 @@
 [Desktop Entry]
 Name=PulseEffects
-GenericName=Audio Enhancer
+GenericName=Audio Enhancer and Equalizer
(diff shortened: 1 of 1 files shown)


wwmm:
Are you sure it is necessary? I do not have kde to test but at least in Gnome typing `equalizer` in the application search field finds PE with no problems. I think the desktops should use the `keywords` in the desktop file besides what is written in `name` and `generic name`

teneri66:
In KDE it works only when I type the whole word "equalizer". Even if having "equalize" it does not show me any results. If I type "enhan" I see PulseEffects.
That's why I think it's important.

PR #189, opened 2018-03-22
Round Frequency and Intensity to 2 digits in GUI

sedrubal:
#### Before:
screenshot_20180322_194019
#### After:
screenshot_20180322_193840


wwmm:
Thanks! :-)

PR #193, opened 2018-03-27
Debug messages

Bleuzen:
Hey
I made PulseEffects to be able to print debug messages. Debug messages are only printed if the PULSEEFFECTS_DEBUG env var is set.
I think this could be useful if a user has a problem (or someone wants to know more about what PE is doing). So you can print some debug messages without spamming the console output to much for other users.

This does not contain much helpful debug messages yet. I just implemented some examples. But you can now print debug messages whenever you want. Maybe you should implement the more messages in the future, because you know better where debug messages would be useful.

For now this is just an idea. I would also understand if you don't like and merge it.
(Btw. this is the first time I did something in Python, so I'm not 100% sure if I did everything right. For me it works, but maybe it's better if you also test if everything still works.)

Edit: Oh and I forgot one thing. I also closed the file object after writing for the autostart file. I think it is better to close such things after usage to get them away from ram when not needed anymore. Anyway, only a very tiny tweak I did beside the debug thing ;)

### DIFF
--- a/PulseEffects/application.py
+++ b/PulseEffects/application.py
@@ -47,6 +47,22 @@ class Application(Gtk.Application):
         self.add_main_option('load-preset', ord('l'), GLib.OptionFlags.NONE,
                              GLib.OptionArg.STRING, help_msg, None)
 
+        self.init_log()
+
+        debug_logging = os.environ.get('PULSEEFFECTS_DEBUG')
+        if debug_logging:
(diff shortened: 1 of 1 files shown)


wwmm:
Thanks. Sooner or later some kind of fine-grained control of log messages would have to be implemented. 

About the file closing Python closes it automatically when opening the file with the statement `with open`. But there is no problem in calling the close method explicitly.

PR #197, opened 2018-04-03
Use Gtk.FileChooserNative

### LINKED ISSUE #195: [flatpak] Impossible to import presets
Hi
I installed pulseeeffects on Arch Linux (Gnome 3.28) via flatpak/flathub.
It's impossible to import presets because dialog to open files shows no file in my home and bookmarks in sidebar are missing.

captura de tela de 2018-03-29 13-41-35

AsavarTzeth:
Gtk.FileChooserNative will use a file chooser dialog that is native to
the desktop environment (when supported). It will also work in sandboxed
environments, such as Flatpak.

Mainly this solves the sandbox problem. Users should now be able to
import presets, no matter how they install the application.

Fixes #195

### DIFF
--- a/PulseEffects/presets_manager.py
+++ b/PulseEffects/presets_manager.py
@@ -208,23 +208,23 @@ class PresetsManager():
                 self.save_preset(path)
 
     def on_import_preset_clicked(self, obj):
-        dialog = Gtk.FileChooserDialog(_('Import Presets'), self.app.window,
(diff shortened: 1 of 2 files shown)


AsavarTzeth:
By the way, there is no rush to make a release. I am already patching this downstream.

wwmm:
Ok. I still have a few things to do. No idea yet when there will be a release

PR #252, opened 2018-06-28
Update Flatpak install instructions

AsavarTzeth:


wwmm:
I didn't notice that. Where is this information?

wwmm:
I will try to find opensource alternatives

AsavarTzeth:
Not sure what you mean? Its covered in detail in the [commit log](https://github.com/flathub/com.github.wwmm.pulseeffects/commits/master). Other than that, where else would it be but here?

I'm just updating your docs to reflect some changes I had to do to the Flatpak.

wwmm:
I did not know that lsp had such restrictions. For some reason I thought it was on the same ground as Calf in this regard. I prefer to use opensource plugins.

I think that the gate plugin from Calf that we already have can be used as an expander. We would not have the `Up` mode but I don't think it is the end of the world. The delay may be a little harder to replace but I will keep looking

AsavarTzeth:
Its okay, I got it now :)

Yeah that description on their web page is a bit missleading. I also think the lsp-plugins author himself actually think the limited source code release is open source.

Thats a problem with custom licenenses, that you write yourself. Easy to make misstakes.

I was considering opening an issue upstream, suggesting they release the aource code release under a BSD or MIT license. Dual licensing is a much better solution.

wwmm:
In any case if the expander is closed source this license change would probably improve only the delay plugin situation. In the next PE release I will remove the expander and put Calf gate in his place. As the gate is already used in the microphone effects pipeline it is a very easy move. A few lines of code.

The delay plugin can be replaced by a standard stereo delay plugin that I think it should be easy to find. The user will have to calculate by himself the corresponding delay for each distance from the listener to the speaker but in the end the correction can be done.

So if you were planning to update the flatpak package wait a little for PE 4.1.2. I will probably add calf multiband gate in this release too but it is already available when compiling calf anyway.

AsavarTzeth:
Agreed. I also doubt they would get the 2500$ to release the code. That is for all plugins up to that goal. Its too bad there is no option to just fund that plugin alone.

I will wait with the update. Take your time.

PR #278, opened 2018-07-21
POSIX shell /bin/sh instead of /bin/bash

mikhailnov:
Do not require bash for the script which is fully compatible with minimal POSIX shell

(`shellcheck` can be used to check the scripts, it checks for POSIX compatibility and no bashisms if the shebang id `#!/bin/sh`, `Geany` (build->lint) can be a GUI for `shellcheck`)
deepinscreenshot_chromium-browser_20180721174526

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,4 +1,4 @@
-pulseeffects (4.1.9-2) unstable; urgency=low
+pulseeffects (4.1.9-3) unstable; urgency=low
 
   * updated Russian translation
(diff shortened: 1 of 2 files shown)


wwmm:
No problems. The file `environmental_variables.sh` is only a shortcut I use in development. It is not installed int the user computer. So it does not make much difference which shell it uses. In any case it is good to know it is not restricted to bash.

PR #281, opened 2018-07-21
Initial upload of Russian help

mikhailnov:
(I have not not yet tested if it builds because of meson)


wwmm:
It seems ok
screenshot from 2018-07-21 13-38-01

mikhailnov:
Is help accessible from PulseEffects UI?

wwmm:
Not yet. I still have to see how people make the call to yelp. What I am doing for now is run `yelp index.page`

mikhailnov:
I first looked for a button , checked all menus and then tried to click F1

wwmm:
Ok

wwmm:
Now there is a help button and the F1 shortcut works.

I did not find a way to test this without installing. that is why I created a aur package that builds from git and added it to the git repo.

As far as I could see everything is fine

mikhailnov:
You've broken debian buildings, including debian/build-ppa.sh and instructions in wiki. Direcotory 'debian' must be in the root of the source code, but not in dists/. Otherwise it has to be `mv`-ed before building.

mikhailnov:
I mean building deb package from source:
https://github.com/wwmm/pulseeffects/wiki/Installation-from-Source#debian--ubuntu

mikhailnov:
https://github.com/wwmm/pulseeffects/pull/283 fixes this

wwmm:
It would be more interesting but it doesn't  seem to be easy to do without adding one help button to each plugin. What would be a little tiring.

I also don't  know if I can really open a section directly. If you take a look at how the help is called in the [source](https://github.com/wwmm/pulseeffects/blob/221f1b38121760e7f9f1861ab5a92095affe0111/src/application.cpp#L153) you will see just a `help:pulseeffects`. Yelp finds the `index.page` on its own. In the gtk projects I looked I did not see direct links to a section.

PR #289, opened 2018-07-29
v4.2.1 deb and UI changes

mikhailnov:
* v4.2.1 debian package
* `gstreamer1.0-autogain-pulseeffects` package
* updated Russian localization
* you forgot to mansion `libebur128-dev` as a new build dependency 

* changes to UI: I tried to make the new 'Auto Gain' UI better understandable; did not update any .po files except ru.po, they should be updated after this change

deepinscreenshot_pulseeffects_20180729145707
deepinscreenshot_pulseeffects_20180729145418

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ level.
 - Sometimes PulseEffects was not paused when there was no application playing
 audio. Fixed that.
 
+### Note for Packagers
+- 	libebur128-dev is now a build time dependency.
+
 ## [4.2.0]
 ### Added
 - We finally have plugins documentation and a few basic explanations about some
--- a/data/ui/autogain.glade
+++ b/data/ui/autogain.glade
@@ -143,7 +143,7 @@
                 <property name="can_focus">False</property>
                 <property name="halign">center</property>
                 <property name="valign">center</property>
-                <property name="label" translatable="yes">Momentary</property>
+                <property name="label" translatable="yes">Influence of measurements with momentary integration time</property>
               </object>
               <packing>
                 <property name="left_attach">1</property>
@@ -156,7 +156,7 @@
                 <property name="can_focus">False</property>
                 <property name="halign">center</property>
                 <property name="valign">center</property>
-                <property name="label" translatable="yes">Short Term</property>
+                <property name="label" translatable="yes">Influence of measurements with short integration time</property>
               </object>
               <packing>
                 <property name="left_attach">2</property>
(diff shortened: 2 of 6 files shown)


wwmm:
I cherry picked the commits that were not related to the auto gain. The interface gets too verbose with these changes. It also defeats the purpose of having written the help in the manual. And different number of lines between neighbor labels is too striking to the eyes. 

If for some reason you think the user is not going to open the help or be able to find the auto gain entry this could be added as a tooltip shown when the mouse is over the widget.

wwmm:
I am not against make the ui more user friendly whenever it is reasonable to do so and I understand it is not optimal to the user with less technical knowledge. But no audio equipment or plugin out there(be it for windows, linux or mac) has so much text in its interface. I agree with them in that decision. Although not perfect tooltips are a decent middle ground here for most of the users.

mikhailnov:
Plugins are made for professional audio engineers, not other people. As pulseeffects is not made for professional audio processing and is opensource, we can try making it friendlier...
Tooltips do not solve the problem 'looking at the UI and do not understand anything', but still better than anything

mikhailnov:
Maybe move on/off button into the bottom to prevent UI from being verbose on window resize?

mikhailnov:
(it will be a bit strange)

wwmm:
It will be very strange XD

mikhailnov:
or it's possible to make a general title for multiple controls and leave shorts names of each control, it will use free space and prevent text duplication (the same text is duplicated in my variant)

mikhailnov:
e.g.
Title for 3 controls: `Influence of measurements with different integration time`
control 1: `Momentary`
control 2: `Short term`
control 3: `Integrated value`
and move control 4 `Target` to separate line with text `Target` on the left from control, not above it

mikhailnov:
`Weights` will be a word that just gives a general title for the group and does not help to better understand what they do

mikhailnov:
Here are some suggestions from other people:
https://vk.com/wall-114916478_314164?reply=314174

wwmm:
I still think it is better to have less text. To me the ui only gets more confusing to read with long lines of text there. Another problem of having long texts is that there isn't space to do this in all of them. The multi band versions of the gate and the compressor are examples of interfaces that are already crowded. It seems odd to me to have a few plugins with lots of text and others only with keywords.

The proposal to have the enable button to the right is not bad but I prefer to have this button in the same position in all plugins interface.

I will stick to short texts
screenshot from 2018-07-29 14-35-55

mikhailnov:
Well, the meaning of 'Weights' is clear only after reading help... Weights in what is unclear

Issue #295, opened 2018-08-01
Limit effects to audio device/Device-specific profiles


jonas2515:
It would be great if PulseEffects would support device-specific profiles or could turn on and off depending on the output device.

This would be especially useful for people using PulseEffects to adjust the sound of internal speakers on laptops. As soon as a headphone is being used, those adjustments aren't needed anymore and have to be manually turned off at the moment.

wwmm:
As there is a similar issue opened with more discussion I will close this one. Anyone comeing here should follow #398

Issue #288, opened 2018-07-28
pulseaudio avoid-resampling=yes not working when pulseeffect in use.


Neo595:
Hello,
the option avoid-resampling in pulseaudio allows to let the soundcard to use native sampling if supported avoiding to systematically resampled stream.
To reproduce the issue :
1) verify it works : stop pulseeffect, play file (VLC) at 88.2K or 96K, you can see that sampling output follows file sampling rate.
2) if you start pulseeffect at that moment, it still worked at native sampling rate (Resampler = copy)
3) play another file with different sampling rate (e.g 88.2K instead of 96K) then the output stream is resampled from 88.2K to 96K, sampling rate is not updated. pulseffect input sampling is not updated with new native sampling rate (stick to sampling rate at launch, I guess). Resampler= speex-float10

38github:
I have started loving using avoid-resampling and I am certain, though without ABX logs, that I am hearing better sound by avoiding the resampling. I also love Pulseffects because I can EQ some really cheap in-ear phones to go from very mid-rangey to sounding actually quite good. The problem is though what has been mentioned that the sample rate is "held hostage" by Pulseeffects. I really hope someone comes up with a solution. If there is anything a non-programmer like me can do just let me know (testing?).

wwmm:
Last month I contacted Pulseaudio's developers and if one of pull requests is accepted(I think it probably will be) we should be able to reconfigure the null sink sampling rate in the next Pulseaudio version. Here you can see the discussion [https://lists.freedesktop.org/archives/pulseaudio-discuss/2018-September/030512.html](https://lists.freedesktop.org/archives/pulseaudio-discuss/2018-September/030512.html).

For now all we can do is waiting.

Neo595:
Hello, 
Very good news, Arun is a lead pulseaudio developer. The "avoid-resampling" feature has been developed by him. Wait & see...
From discussion, it seems that if the stream is declared as "compressed", it will activate "pass through" behavior so sample rate is kept, maybe an indirect way to get the good property ?
Regards.

wwmm:
Good news! I implemented this feature in the master branch. But in order to test you will need Pulseaudio 13(not release yet) or compiling Pulseaudio from sources. People willing to try will find a step by step to compile Pulseaudio in a post I made in #350. It is not easy but it would be good to have feedback. At least for me it is working.

Neo595:
Hello, 
I've compiled from source as described;
[...]

PR #301, opened 2018-08-19
Update com.github.wwmm.pulseeffects.appdata.xml.in

har9862:
* Added releases info for the last couple of releases
* Added OARS

### DIFF
--- a/data/com.github.wwmm.pulseeffects.appdata.xml.in
+++ b/data/com.github.wwmm.pulseeffects.appdata.xml.in
@@ -56,6 +56,36 @@
       <caption>Calibration</caption>
     </screenshot>
   </screenshots>
+  <releases>
+    <release version="4.2.8" date="2018-08-14">
+      <description>
+        <p>Added:</p>
+        <ul>
+          <li>The latency introduced by the Gstreamer pipeline is displayed in the headerbar.</li>
+        </ul>
+      </description>
+    </release>
+    <release version="4.2.7" date="2018-08-12">
+      <description>
+        <p>Added:</p>
+        <ul>
+          <li>The adapter plugin used internally by PE reports the latency that it may introduce(proportional to the block size) to GStreamer. Latency is only added when the block size s larger than the number o samples per buffer in Pulseaudio buffer.</li>
+        </ul>
+        <p>Fixed:</p>
+        <ul>
+          <li>Fixed a regression that could cause severe noises when multiple audio apps were playing at the same time.</li>
+        </ul>
+      </description>
+    </release>
+    <release version="4.2.6" date="2018-08-11" />
+    <release version="4.2.5" date="2018-08-09" />
+    <release version="4.2.4" date="2018-08-05" />
+    <release version="4.2.3" date="2018-07-31" />
+    <release version="4.2.2" date="2018-07-29" />
+    <release version="4.2.1" date="2018-07-28" />
+    <release version="4.2.0" date="2018-07-24" />
+  </releases>
+  <content_rating type="oars-1.1" />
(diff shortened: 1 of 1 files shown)


wwmm:
Thanks!

AsavarTzeth:
I actually have a more extensive version of this, which I have maintained for a little while now. In time I did plan to ask you if you wish to maintain it upstream.

This though, causes issues for me. I have to ask, are you planning on maintaining this yourself from now on? It has to be included prior to tagging each release. Otherwise this just made my maintenance of the Flatpak a great deal lot harder.

Now as for the changes themselves I see multiple issues. First of have you verified this against `appstreamcli validate --pedantic` and `appstream-util validate-strict` (or at least `appstream-util validate`)? It seems mostly correct, but there are a few issues.

Lastly, the OARS data seems completely invalid. There is no data at all, just an empty tag? Again `appstreamcli` seems to react on this.

I recommend you look at at my patches:

- https://github.com/AsavarTzeth/pulseeffects/commit/99ce9cb8621ccfa20d7e97a9f49c351572c3eba1
- https://github.com/AsavarTzeth/pulseeffects/commit/d61a54c18e8c686cef07369ae50381947d0b22e8

Those have been in use for a while now on Flathub and it would probably be wiser to just use those instead. That is if you wish to maintain this yourself, which would certainly be welcome.

wwmm:
I never thought about this file before receiving this pull request. So I was not aware that it had to be validated by appstream tools. I was also not aware you had been patching the flatpak package with this. Sorry :) If you have more extensive versions it is better to use the ones you have

Is flatpak the only one making use of appstream data or things like gnome and kde software center also make use of it? If there is use outside of flatpak it makes sense to maintain this info here.

Although I don't really think it is a problem to update this file and the changelog before each release I still wonder if there is a way to somehow to automate the creation of one of the files. Something like creating a custom meson build target that executes a script that would generate one of the files.

AsavarTzeth:
Well I feel it is also my fault for never taking the time to mention it. In a way this pull request was a great way to force my hand.

All frontends pretty much use AppStream appdata, i.e GNOME Software, KDE Discover and the Flathub webb frontend.

Now with the release information specifically the advantage is tjat flatpak is then able to display the version.

About automation, I have tried many many times. The thing that is very difficult to do in a safe and reliable way with XML. Based on my attempts, it is possible, as long as you skip the description.

The problem with the description isn't so much XML as reliably parsing the current changelog. Adding a more parsable file, only for release descriptions, sort of defeats the idea. It just moves the issue.

Anyhow the current changelog isn't really suitable, since its a bit to technical and mentions some things that end users don't need to care about. It also fails some appstream-glib specific validations like the 100 char max per list item.

So if you're happy with just version and date (the most important parts), yes I could automate it.

I even made a script way back for this, which I could adjust for meson.

I could also toss in some appstream validation tests that run with meson.

AsavarTzeth:
Well for what it is your changelog is mostly fine and informative to myself. I am just saying it may not all be something you would show in GNOME Software, or for comparison, Google Play or the Apple App Store.

I tried adjusting it for the audience, for a while. But it got too tedius; well at least with daily releases :P

I don't think descriptions would be overwritten. The automated part basicly would just add one or two lines to the top of the releases (the order matters). You could then add things in between the opening and closing tags.

As long as it first checks if a release already exists it should be fine. I will verify this when I get home, which is soon.

wwmm:
Hahaha That is what I thought. It would be tedious to rewrite everything for non technical users :D

I think that we will have less often releases for a while. I don't have any plans in my mind and at this moment I don't know if it is even possible to do what is asked in #295 and #288. So the next updates will happen only if there is a bug to fix or if people send updated translations(or updated things like appstream info)

PR #303, opened 2018-08-21
Add automation, testing and releases for appdata

AsavarTzeth:
Here you go. It turns out we were very lucky. For whatever reason my xml style sheet (which was essential for this) was located on a single computer in the trash bin, which I have not emptied in a long time.

I have added the mentioned appdata validation tests, as well as tests for the desktop entry file. Tests are run within the standard test target. Details are documented in the 83f6cd1 commit message.

I also added a script that you can run to add the minimal required release information to the appdata file. It also offers great foundations if we want to automate more similar things with the same script, instead of adding a bunch of inconsistent separate tools.

I did my best to test it. I tried failing every command on purpose and I ran the script both from within the repository and outside. It should work from wherever and however you run it.

As of now all the strictest AppStream validation tests passes, that is if you run it against the final file (not the .in file). And yes, I made sure it always passes after running the new script.

The only bad news is, I don't think you can create a custom Meson target and add the script to that. I also cannot find any existing target that matches this use case. You will just have to run it directly. Anyhow, in this case a target only makes sense if we had more than one command added to it. If a need for more actions comes up we could, as mentioned earlier, just integrate it into the existing script.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,9 @@ obj-*-linux-gnu
 # binary language files
 *.mo
 
+# patches
+*.patch
+
 \.python-version
 
 # exclude debian building logs and other junk
--- a/data/com.github.wwmm.pulseeffects.appdata.xml.in
+++ b/data/com.github.wwmm.pulseeffects.appdata.xml.in
@@ -37,11 +37,11 @@
   <screenshots>
     <screenshot type="default">
       <image>https://raw.githubusercontent.com/wwmm/pulseeffects/master/images/appdata-screenshot-01.png</image>
-      <caption>Applications</caption>
+      <caption>Set the volume and turn on/off effects</caption>
     </screenshot>
     <screenshot>
       <image>https://raw.githubusercontent.com/wwmm/pulseeffects/master/images/appdata-screenshot-02.png</image>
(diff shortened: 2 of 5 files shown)


wwmm:
I am fine with running a bash script.

AsavarTzeth:
Oh you're right! It should increment the version or maybe better parse it from the main meson.build file.

wwmm:
It happens haha :)

Parse it from the meson files sounds like a good idea. Now that things like the app version in the about window and the plugins version are automatically updated based on the version in the meson file I always check it before a release

AsavarTzeth:
I have some questions regarding release dates. I used to rely upon the tag timezone, which be -0300, or -0200 for some past versions (I assume you moved west?).

Since I now need another source I am cosidering using just using `date --utc +%F`.

Now two questions:

1. Is UTC okay with you? Or do you really need configurable timezone support?

2. If you do need timezone support, which one should I set and test agsinst?

wwmm:
I did not move. Maybe it was Brazilian summer schedule. I do not have a preference regarding date format.

AsavarTzeth:
Ah right, dayligt savings. I suppose that is also one argument for UTC.

PR #305, opened 2018-08-21
v4.2.8 deb

mikhailnov:
* https://github.com/wwmm/pulseeffects/issues/304 - temporary workaround, otherwise I can't build git code
* gstreamer1.0-adapter-pulseeffects package
* new build dep: appstream-util
* fix https://github.com/wwmm/pulseeffects/issues/306


wwmm:
As there is some changes to the appstream validation (`data/meson.build`) I will wait for the @AsavarTzeth  feedback before accepting the commit.

mikhailnov:
PulseAudio 12 is now in the PPA (https://github.com/wwmm/pulseeffects/wiki/Package-Repositories/adecce6a1aeaf36d85680b73294cb137cc9c5b96)

AsavarTzeth:
Could you just not build the `test` target? I mean at the moment these are the only tests run and they are only really useful to upstream anyhow. Building the `test` target is not the default, so you might look if Debian tools override this.

As for the changes, I strongly disapprove. You are lowering validation testing to the lowest possible standard for everyone, most importantly upstream because of an issue that somehow only affects Debian packaging so far. Instead I suggest you look at GNOME Builder and how that is solved on Debian, since they use the same tests, or don't run this target until you find the real cause of the problem.

AsavarTzeth:
In the end, if this is the only practical way for you to solve this at the moment I think you should just carry the validation patches downstream, especially if they are "temp." changes. That way it won't negatively affect anyone else.

AsavarTzeth:
After some further research and considerations I suppose using `--nonet` would not hurt. It seems to only affect validation of screenshots.

Since screenshots tend to only be updated by me, it just means I will have to keep doing validations manually before I update them. Once committed it isn't changing.

Other reasons in favor of `--nonet`:

- Since the screenshots are hosted by GitHub, validation would fail if it is down.

- I now realize `flatpak-builder` also builds without network access by default. I missed this during testing.

I still disagree on using `validate-relax`. It seems to me we should find the real cause of your issues with the validation of the Russian localalization. Issue: https://github.com/wwmm/pulseeffects/issues/304

For now, or until more distributions report similar issues, I think you should just patch in `validate-relax` downstream.

wwmm:
@mikhailnov could you make another pull request without the changes in meson.build? This way we could update all the other changes that have nothing to do with appstream without the need to cherry-pick many commits :-)

AsavarTzeth:
By the way whats rhe deal with that merge commit? You should rebase against upstream, not pull/merge it. Or am I missing something?

wwmm:
I never did a pull request to anyone. So I don't know if there is a proper way for that...

AsavarTzeth:
I could share how I do it but it would be a long comment...

Essentially what you want is a clean log that is easy to follow. Once merged it is too late to fix it, without rewriting history.

Rebasing (and squashing) makes it cleaner to you and all other downstreams.

For example, do we really need to know that patch files were added? Only to then be removed? That is relevant to this downstream but could be squashed.

Try a fewer commits, describing the end result, from the perspective of upstream instead.

AsavarTzeth:
I found a decent guide that explains rebasing & squashing, in the context of pull requests:
- https://github.com/servo/servo/wiki/Beginner%27s-guide-to-rebasing-and-squashing.

There are tons of other guides if you find this one insufficient. Also while I am no git expert, I will do my best to answer any questions.

AsavarTzeth:
By the way, you got me curious about your packaging, so I have to ask:

How are you building your packages? Or rather based upon what source (repo,commit,tag,tarball)? I know it can't be upstream GitHub releases, since your debian files are not updated here until post release.

For that reason I assumed you had a stable branch in your fork, which should now based upon `v4.2.8`. However I cannot find any stable branch in your fork. Is it kept locally on your machine or somewhere else?

All I can see is a **master** branch, based upon **upstream/master**. Are you building directly from this?

mikhailnov:
Yes,  I merged wwmm/pulseeffects/master to mikhailnov/pulseeffects/master, looked at the commits history , saw that after branching v4.2.8 there were only changes in the build system and no changes in PE itself. So I built directly from master to prepare debian specs for future releases. 

At the moment I don't know what to do with validation. If I just make a PR without changving validation, then the package will be unbuildable. Disabling tests completely is also a bad idea. I will start from trying to build with strict validation on Ubuntu 18.10, where appstream-util is the newest (http://packages.ubuntu.com/appstream-util).

24 августа 2018 г. 9:59:47 GMT+03:00, Patrik Nilsson <notifications@github.com> пишет:

AsavarTzeth:
Yes, I just noticed the branch was in fact based upon `v4.2.8` orginaly. I still consider that merge a bad idea though. I fail to see any good reason for it. If you wanted something specific, you do a cherry-pick. If you wanted to do a pull request you do another branch and rebase on master.

AsavarTzeth:
Give me some time and I will try to fix all the issues in your git repo. I could then share the commands and explain how you can fix it yourself.

Once fixed you should not be having validation issues at least until `v4.2.9` and your pull requests will be a lot neater. Preferably we will have found the root cause before `v4.2.9.

AsavarTzeth:
*I'm sorry if my communication skills really have been poor lately. I did not really answer you. My mind was elsewhere. I'm giving it another try.*

> So I built directly from master to prepare debian specs for future releases.

That's okay, but it would be better if you used `git rebase` and a separate testing branch. Using merge in this fashion on a branch you plan on merging back upstream messes up the history.

> At the moment I don't know what to do with validation. If I just make a PR without changving validation, then the package will be unbuildable.

Well the best thing would be if we could reproduce the issues. We could then maybe report it as an `appstream-util` bug. Until then why not use patch files, like you did initially? Alternatively, you could just keep your own temporary changes in separate branch.

PR #309, opened 2018-08-23
Fix add-appdata-release utility

AsavarTzeth:
The initial script had a fundamental issue; the application version and
release dates were based on git tags. Since this script needs to be run
before a release is tagged, that tag will never include an up-to-date
appdata file.

Versions are now parsed from the main meson.build file. It works well
but there is one limitation, the order of the key/value pairs "version"
and "meson_version" matters. If the order of those keys changes this
script would also require adjustments.

Dates are now provided by the `date` command, using the UTC setting.
This offers consistency since it also is not affected by day light
savings.

An added benefit of these changes is that less external commands are
now used.

### DIFF
--- a/data/com.github.wwmm.pulseeffects.appdata.xml.in
+++ b/data/com.github.wwmm.pulseeffects.appdata.xml.in
@@ -57,6 +57,14 @@
     </screenshot>
   </screenshots>
   <releases>
+    <release version="4.2.8" date="2018-08-14">
+      <description>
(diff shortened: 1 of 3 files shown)


AsavarTzeth:
I removed the insertion of the `<description />` tag. At first I used it so that it would be simpler for you to add any descriptions you might want.

However I realize if you, for some reason, don't add a description for a given release it would just look messier and use three lines for no good reason.

PR #343, opened 2018-09-24
fixed a choppy playing issue

qunxyz:
this choppy issue is very annoying, don't know why nobody complaint it

### DIFF
--- a/src/pipeline_base.cpp
+++ b/src/pipeline_base.cpp
@@ -23,8 +23,6 @@ void on_message_error(const GstBus* gst_bus,
     util::debug(pb->log_tag + debug);
 
     if (err->message == std::string("Internal data stream error.")) {
-        pb->set_null_pipeline();
-
         // As far as I know only a bad latency or buffer value causes this error
         // in PE pipeline
 
@@ -154,7 +152,6 @@ void on_buffer_changed(GObject* gobject, GParamSpec* pspec, PipelineBase* pb) {
         /*when we are playing it is necessary to reset the pipeline for the new
          * value to take effect
          */
-        pb->set_null_pipeline();
         pb->update_pipeline_state();
     }
 }
@@ -164,7 +161,6 @@ void on_latency_changed(GObject* gobject, GParamSpec* pspec, PipelineBase* pb) {
         /*when we are playing it is necessary to reset the pipeline for the new
          * value to take effect
          */
-        pb->set_null_pipeline();
         pb->update_pipeline_state();
     }
 }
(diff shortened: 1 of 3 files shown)


wwmm:
I disagree. If we do not put the pipeline in the null state when there is no application actually playing audio we will waste cpu because effects will be applied to silence. In PulseEffects first versions things were like this and it was really bad for people on weak cpu or on laptops. People started to complain and that is why we alternate states today.

Although in my machines I almost never can hear this choppy I agree that at some point it will happen when alternating the pipeline state between the null and the playing state. After all putting the pipeline in the null state means freeing allocated resources. When we tell it to play it has to reallocate everything. But applying effects to silence while wasting cpu in the process is not a real fix.  Maybe well chosen buffer and latency values in the settings menu can help with this. Or maybe the fix should be done at a lower level. In Gstreamer or even Pulseaudio. Maybe if we put the pipeline in the paused state the choppy won't happen. But we will may have other side effects... Maybe there are different ways to structure the GStreamer pipeline so that this doesn't happen... 

Another reason for not accepting the pull request is that it breaks other parts of the code. Some settings like the buffer size or the `pulsesrc` plugin latency only take effect when the pipeline is put in the null state and then back in the playing state. GStreamer just ignores the value if we try to change it when playing. When you removed `set_null_pipeline` from `update_pipeline_state` you broke that.  So when the user changes any of these parameters on the fly he will think that the change is taking place. But in reality it is not. In my opinion a real fix for this should be done in GStreamer.

qunxyz:
ok. got it. but the choppy issue has to be happen, cause set_null_pipeline take time. and set_null_pipeline not just make playing choppy, it seems cause gnome choppy too. I'll try other way, see what I can got.
ps. thanks for your great job, i really like this app

wwmm:
This seems to be one of those issues whose intensity are hardware dependent. I use gnome here and nothing strange happens when the pipeline alternates states.

I will investigate which changes we need to make to alternate between the paused and the playing state instead of the null. I know that we can't just do that as things are right now. Depending on the blocksize chosen by the user the peadapter plugin may accumulate a lot of data. If this data is not flushed in the paused state the user may listen to old audio data once the pipeline is restarted after some time. If this would be as annoying as the choppy is hard to say.

wwmm:
I made a few changes to the function `update_pipeline_state`. They are already in the master branch. As far as can see there is no side effect. Please test if there were any change to the choppy

qunxyz:
better now. but got some cracking noises, I think it's ok.

qunxyz:
and I figured out what cause gnome choppy, my system problem, sorry

PR #381, opened 2018-11-22
Update the Croatian translation

flipwise:
The Croatian translation is in a pretty bad state. There were many false translations (= same terms being used everywhere) and some missing translations. I tried to fix it to the best of my knowledge, but an expert in the field of acoustics needs to have a look at this in order for the translation to be adequate. In this state it's still so-so.


wwmm:
thanks!

Issue #389, opened 2018-12-16
No sound, continuous "Internal data stream error" with "streaming stopped, reason not-linked (-1)"


vmatare:
I'm now opening a separate issue for this, since it is related, but not identical to #380. So far I've been trying to understand what exactly the logged messages mean. Here they are again in full, for reference:
```
(pulseeffects:26217): pulseeffects-DEBUG: 01:03:10.132: sie: NULL -> READY -> PLAYING
(pulseeffects:26217): pulseeffects-DEBUG: 01:03:10.132: sie: READY -> PAUSED -> PLAYING
(pulseeffects:26217): pulseeffects-DEBUG: 01:03:10.132: sie: sampling rate: 44100 Hz
(pulseeffects:26217): pulseeffects-DEBUG: 01:03:10.132: sie: PAUSED -> PLAYING -> VOID_PENDING
(pulseeffects:26217): pulseeffects-DEBUG: 01:03:10.132: sie: total latency: 0 ms
(pulseeffects:26217): pulseeffects-DEBUG: 01:03:10.132: sie: pulsesrc latency [us]: 10000
(pulseeffects:26217): pulseeffects-DEBUG: 01:03:10.132: sie: pulsesrc buffer [us]: 200000

(pulseeffects:26217): pulseeffects-CRITICAL **: 01:03:10.156: sie: Internal data stream error.
(pulseeffects:26217): pulseeffects-DEBUG: 01:03:10.156: sie: /var/tmp/portage/media-libs/gstreamer-1.14.4/work/gstreamer-1.14.4/libs/gst/base/gstbasesrc.c(3055): gst_base_src_loop (): /GstPipeline:pipeline/GstPulseSrc:source:
streaming stopped, reason not-linked (-1)
(pulseeffects:26217): pulseeffects-DEBUG: 01:03:10.157: sie: NULL -> VOID_PENDING
```
From what little I've read about gstreamer so far, it seems to me that for some reason, the entire pipeline ends up being unconnected. So I've produced a dot rendering of [the pipeline](https://github.com/wwmm/pulseeffects/files/2683175/pulseeffects-pipeline.pdf), and what really looks suspicious to me is those two `proxypad` thingies at the start and end. Shouldn't there be something related pulseaudio, like a `GstPulseSrc` resp. `GstPulseSink` or something?

[...]

PR #390, opened 2018-12-16
fix issue #389

vmatare:
Not entirely sure whether ba9d3d9 is the best way to fail here, I'm not very familiar with GTK. But you definitely want to trigger things like e.g. Fedora's `abrtd`, since if this happens, it's likely to be a packaging bug.

### DIFF
--- a/include/pipeline_base.hpp
+++ b/include/pipeline_base.hpp
@@ -67,6 +67,8 @@ class PipelineBase {
   void set_caps(const uint& sampling_rate);
   void init_spectrum_bin();
   void init_effects_bin();
+
+  static GstElement *get_required_plugin(const gchar *factoryname, const gchar *name);
 };
 
 #endif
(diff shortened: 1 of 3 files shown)


wwmm:
Thanks! Even if it is not the best way to handle this it is still an improvement. The plugins loaded in `pipeline_base.cpp` are mandatory. If they are not installed PulseEffects is unusable anyway. The plugins that can be enable/disabled by the user are optional and in their case a `throw` would make no sense.

vmatare:
BTW, in the ChangeLog you write "fix crash". I wouldn't call these things crashes because a crash is something very loud and obvious, like a segfault or an uncaught exception, i.e. something that makes the app exit and triggers an error handling routine outside of the app itself. In most cases that would be an OS signal handler and/or something like coredumpctl or abrtd.

In this case however, no such thing happened, the problem merely triggered a log message. What the patch does is in fact to **make pulseeffects crash** since that is precisely what you want to happen in a situation where the app is so fundamentall broken that it can impossibly work.

PR #391, opened 2018-12-16
make the spectrum awesome

vmatare:
This commit adds settings that make the spectrum view much more useful.

Most importantly, its update frequency is now configurable from 1 to 200
Hz. Additionally, the magnitudes can now be scaled with an exponent:

final_mag = (scale * mag) ^ exponent

This way we can actually visualize what is happening. The defaults are
set so that the original behavior is unchanged, however the sweet spot
is at very different values:

Sampling Freq: 60 Hz (The usual screen refresh rate, no point in going
                      higher than that)
Scale: 1.2
Exponent: 3.0

With these settings, things actually become visible. Scale and exponent
are of course highly dependent on the type of audio material. For added
pleasure, also double the number of points and the height ;-)

### DIFF
--- a/data/schemas/com.github.wwmm.pulseeffects.gschema.xml
+++ b/data/schemas/com.github.wwmm.pulseeffects.gschema.xml
@@ -158,6 +158,9 @@
         <key name="show-spectrum" type="b">
             <default>true</default>
         </key>
+	<key name="spectrum-fill" type="b">
+	    <default>false</default>
+	</key>
         <key name="spectrum-n-points" type="i">
             <range min="2" max="600"/>
             <default>150</default>
@@ -171,6 +174,18 @@
         <key name="spectrum-height" type="i">
             <default>100</default>
         </key>
+	<key name="spectrum-scale" type="d">
+	    <range min="0.0" max="10.0"/>
+	    <default>1.0</default>
+	</key>
+	<key name="spectrum-exponent" type="d">
+	    <range min="0.0" max="10.0"/>
+	    <default>1.0</default>
+	</key>
+	<key name="spectrum-sampling-freq" type="i">
+	    <range min="1" max="200"/>
+	    <default>10</default>
+	</key>
         <key name="enable-all-apps" type="b">
             <default>false</default>
         </key>
--- a/data/ui/application.glade
+++ b/data/ui/application.glade
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.22.1 -->
+<!-- Generated with glade 3.20.4 -->
 <interface domain="pulseeffects">
   <requires lib="gtk+" version="3.20"/>
   <object class="GtkAdjustment" id="buffer_in">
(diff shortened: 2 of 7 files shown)


vmatare:
I'd also argue that at the current default settings, the spectrum isn't really useful. There's a huge difference between the default 10 Hz and, say, 30 Hz. So I'd advocate 30 Hz as the new default since it doesn't even increase CPU load a lot. However it might need some additional testing on low-end systems. The default scaling values could also be changed to something with an exponent > 1 since that greatly improves the visibility of individual instruments, voices etc. at no additional computational cost.

wwmm:
Interesting. I will make some tests and see what is causing negative values. But I have already seen something else that needs to be extended. In PulseEffects we have 2 pipelines. One for apps output 
 and another for microphone output. Using Pulseaudio vocabulary it is one for `sink inputs` and another for `source outputs`. So in the function `on_spectrum_sampling_freq_set` we have to to the same for the pipeline that handles microphone. In other words it is also necessary to call:
```
app->sie->spectrum_interval = guint64(1000000000. / spectrum_sampling_freq->get_value());
app->sie->update_spectrum_interval();
```

wwmm:
In my computer I doubt I would have a problem with higher sampling frequencies. But based on issues opened along the time I'd say that there is quite a number of PulseEffects users on weak hardwares like raspberry. Based on this I think that keeping the `10 Hz` frequency will avoid a flood of opened issues about too much cpu usage.

vmatare:
Oh right, I hadn't thought of all those raspberries. I'll do the source output thing a little later, hope there's no rush in merging this...

wwmm:
There isn't :-) I am taking a look at the negative values. This change to the `on_message_element` callback fixes it
```
if (min_mag < pb->spectrum_mag[n]) {
  pb->spectrum_mag[n] = (min_mag - pb->spectrum_mag[n]) / min_mag;
} else {
  pb->spectrum_mag[n] = 0.0f;
}
```

wwmm:
I didn't write about this anywhere but in PulseEffects source there is a `.clang-format` file I use to force code style through Atom plugins. There is also a `.clang_complete` file to do code completion using clang.

vmatare:
Huh. So the `spectrum` plugin actually returns magnitudes lower than the threshold? That would be a bug, since (Quote from `gst-inspect-1.0 spectrum`):
```
  threshold           : dB threshold for result. All lower values will be set to this
```
But whatever, in that case the `min(1, max(0, mag))` thing is required as a workaround, anyways.

wwmm:
I found something else that it has to be considered when doing a scaling. When we hover the mouse over the spectrum it shows the `frequency` and `intensity` of the corresponding position. This is done in the method `ApplicationUi::on_spectrum_motion_notify_event`. As things are right now the value shown by the mouse will not correspond to the bar height when scaling

wwmm:
And yes. It seems there is a bug in the spectrum plugin threshold

wwmm:
You probably have already seen it but in the file `pipeline_base.hpp` there are a few parameters that are hardcoded. The parameter `spectrum_nbands = 1600` controls the number of bands that the spectrum plugin uses when calculating the FFT. As you want to see more details in the spectrum you may want to play a little with it. The higher the number the higher the resolution in the low frequency region. But the cpu usage may increase considerably. The 1600 value seemed reasonable to me. But it may not be to everybody

vmatare:
Oh thanks, that's a good hint, I was actually looking at that bottom region with a little bit of disappointment. I'll take a look and see if it's worth more widgets ;-)

wwmm:
The downside is that in order to cut in half the bin size we have to use `3200` bands. My ryzen can handle this but I couldn't hardcode such a high value. People on weak processors would suffer.  And as it took me many months to improve other things like the dynamic pipeline I forgot to expose this parameter.

wwmm:
As I am with time to do the little fixes I mentioned above I decided to merge it. And as I will add some things to the interface it is better to add these ones first

PR #409, opened 2019-01-04
realtime_kit: Port to GDBus

jtojnar:
There is no need to use libdbus, GLib contains its own implementation.

Builds but I did not test this yet.

Also this is mostly 1:1 port to low-level GDBus API, in the future, it could be simplified using higher-level APIs.

### Relevant docs
- https://dbus.freedesktop.org/doc/dbus-specification.html#basic-types
- https://dbus.freedesktop.org/doc/api/html/group__DBusConnection.html
- https://developer.gnome.org/gio/stable/GDBusConnection.html
- https://developer.gnome.org/glibmm/2.53/classGio_1_1DBus_1_1Connection.html

### DIFF
--- a/include/realtime_kit.hpp
+++ b/include/realtime_kit.hpp
@@ -1,7 +1,7 @@
 #ifndef REALTIMEKIT_HPP
 #define REALTIMEKIT_HPP
 
-#include <dbus-1.0/dbus/dbus.h>
+#include <giomm/dbusproxy.h>
 #include <iostream>
 
 #define RTKIT_SERVICE_NAME "org.freedesktop.RealtimeKit1"
@@ -19,7 +19,8 @@ class RealtimeKit {
  private:
   std::string log_tag;
 
-  DBusConnection* bus;
+  Glib::RefPtr<Gio::DBus::Proxy> proxy;
+  Glib::RefPtr<Gio::DBus::Proxy> properties_proxy;
 
   long long get_int_property(const char* propname);
 
--- a/src/meson.build
+++ b/src/meson.build
@@ -104,7 +104,6 @@ pulseeffects_deps = [
 	dependency('boost', version: '>=1.65', modules:['system','filesystem']),
 	dependency('sndfile'),
 	dependency('threads'),
-	dependency('dbus-1'),
 	dependency('gstreamer-audio-1.0')
 ]
 
--- a/src/realtime_kit.cpp
+++ b/src/realtime_kit.cpp
@@ -8,23 +8,23 @@
 RealtimeKit::RealtimeKit(const std::string& tag) {
   log_tag = tag + "rtkit: ";
 
-  DBusError error;
-
-  dbus_error_init(&error);
-
-  if (!(bus = dbus_bus_get_private(DBUS_BUS_SYSTEM, &error))) {
+  try {
+    proxy = Gio::DBus::Proxy::create_for_bus_sync(Gio::DBus::BusType::BUS_TYPE_SYSTEM,
+                                                  RTKIT_SERVICE_NAME,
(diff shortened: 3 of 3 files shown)


jtojnar:
Hmm, I am getting

```python
(pulseeffects:4157): pulseeffects-WARNING **: 04:02:34.380: sie: rtkit: The connection is closed

(pulseeffects:4157): pulseeffects-WARNING **: 04:02:34.380: sie: rtkit: The connection is closed
fish: “env G_MESSAGES_DEBUG=pulseeffec…” terminated by signal SIGSEGV (Address boundary error)
```

wwmm:
I am testing it now and I get the same here. The crash happens in the method `get_int_property`. For some reason at the time `r = bus->send_message_with_reply_sync(m, -1);` is called the connection is closed

jtojnar:
I guess I need to move reading the reply to the try block but I am still not sure how it would have worked before if similar issue was encountered. And also why the connection is getting closed.

wwmm:
I tried to create the connection with `bus = Gio::DBus::Connection::get_sync(Gio::DBus::BUS_TYPE_SYSTEM);` and it goes ahead. But crashes in other sections with these messages:
```
(pulseeffects:8010): GLib-GIO-WARNING **: 11:44:25.199: g_dbus_message_set_serial: Attempted to modify a locked message

(pulseeffects:8010): GLib-GIO-WARNING **: 11:44:25.199: g_dbus_message_set_serial: Attempted to modify a locked message

(pulseeffects:8010): pulseeffects-WARNING **: 11:44:25.199: sie: rtkit:  : Expected value of type i or x but received (v)

(pulseeffects:8010): pulseeffects-WARNING **: 11:44:25.200: sie: rtkit:  : Expected value of type i or x but received (v)

(pulseeffects:8010): GLib-GIO-WARNING **: 11:44:25.203: g_dbus_message_set_serial: Attempted to modify a locked message
(pulseeffects:8010): pulseeffects-DEBUG: 11:44:25.203: sie: rtkit: changed GstQueue:eq_queue_L thread real-time priority value to 4
```
But it at least was able to set the priority in one of the threads...

jtojnar:
> Expected value of type i or x but received (v)

This I understand. I saw in [Bustle](https://gitlab.freedesktop.org/bustle/bustle) that rtkit sends `@v <@x 200000>` instead of just `@x 200000` one would expect. That explains why the iterator code was so complicated.

jtojnar:
I cleaned it up a bit with [`call_sync`](https://developer.gnome.org/glibmm/2.53/classGio_1_1DBus_1_1Connection.html#a0453f36932222ac59684548c5e7e7a87) method, no longer creating the `Message` manually. But am still no wiser why the connection gets closed.

If I dump the closed status at the top of the `get_int_property` method, it shows `0` but the dbus call still fails few lines down.

```
util::warning(log_tag + "get_int_property: bus.closed=" + std::to_string(bus->is_closed()));
```

jtojnar:
Another simplification with the use of high-level `DBus::Proxy` API. Now it crashes with some broken assertions.

jtojnar:
Finally got it to work – I was forgetting about the fact that the reply is boxed in a tuple.

But I do not really like the need to explicitly use `org.freedesktop.DBus.Properties` proxy. I would much prefer to just use [`g_dbus_proxy_get_cached_property`](https://developer.gnome.org/gio/stable/GDBusProxy.html#g-dbus-proxy-get-cached-property) but it just returns `NULL`, contrary to what the [constructor](https://developer.gnome.org/glibmm/2.53/classGio_1_1DBus_1_1Proxy.html#af3d7e5ea04b9f27862067372f42f2887) would lead me to believe.

PR #417, opened 2019-01-23
Fix pulseeffects.svg in Qt environments

### LINKED ISSUE #308: Default icon rendered incorrectly in Qt environments
Hey, so I know PE got a nice blue new icon recently, but on two of my computers that I tried it's showing up completely white.

image

I have the flatpak PE version 4.2.7 installed here. On my other computer I'm using the latest version compiled from AUR. 

When I click on 'About PulseEffects', the right icon is shown there, however.

DE is KDE Plasma.

noahadvs:
Fixes #308 

There was an unnecessary reference to linear gradient A in linear gradient B. It was unnecessary because B had its own colors defined. The icon works correctly when the reference to A is removed.

I also optimized the SVG to reduce the amount of disk space it uses by about 24%.


wwmm:
Thanks!

PR #427, opened 2019-02-09
Add link to archlinux git version (aur)

Bleuzen:
I made an aur package to make it easier to test the latest git version. This PR just adds a link to the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ Version `1.12.5` or higher
 
 ### GNU/Linux Packages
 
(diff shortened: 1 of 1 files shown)


Bleuzen:
Ups, I just noticed that there is already a PKGBUILD in the git repo here. Sorry, didn't see it. This make it a little less needed.
But I still think an aur package is a good thing to have.

wwmm:
There is no problem. When PE was moved to the official repo I considered to make an aur pkg for git. But I did not want one more thing to maintain XD

The pkgbuild in the source is mostly used by myself when doing tests before a new release.

PR #428, opened 2019-02-09
97% complete german translation!

BlobCodes:
I couldn't find a good replacement for "Crystalizer", so I instead translated the name of the effect in Viper4AndroidFX.
There it is called "Spectrum extension" which could be translated to "Frequenzerweiterung".
I hope that's what it means :D

There are only 9 texts to translate left, but most of them are abbreviations (like BWC, LRX, MT) so they are probably the same in german.

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -7,8 +7,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PulseEffects\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2019-02-09 13:40-0200\n"
-"PO-Revision-Date: 2019-02-09 16:15+0100\n"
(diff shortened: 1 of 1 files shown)


BlobCodes:
It would be very good if someone could tell me what those three abbreviations mean.

wwmm:
They are explained in the equalizer author page http://lsp-plug.in/?page=manuals&section=para_equalizer_x32_lr

The Crystalizer plugin increases dynamic range https://en.wikipedia.org/wiki/Dynamic_range IT doesn't change the frequency spectrum.

wwmm:
Sorry. Not all of the abbreviations are in the page above. Only the ones from the equalizer. The ones from Calf plugins can be found here https://calf-studio-gear.org/doc/index.html

wwmm:
If I am not mistaken `S/C` is probably related to side chain

PR #429, opened 2019-02-09
Add spectrum border option

BlobCodes:
I didn't know how to complete the language files so they were left as it is. (Maybe someone who knows how to do it could create a push request?)
With this option the spectrum bars don't have a border between them. 
I think this looks very cool.
screenshot_20190209_222650


BlobCodes:
By default it's the way it was before.
screenshot_20190209_232406

BlobCodes:
I just noticed a bug:
The spectrum stops drawing when you move your mouse over it and have a high amount of bars.
I think I shouldn't get the value of the switch in the for loop for every bar :laughing:

BlobCodes:
That bug is now fixed.
Aside from that language-file thing, everything is ready!

wwmm:
It is way better this way! Thanks! :-)

Honestly I think I will remove the old way and let only this one :-)

wwmm:
The overlap between the rectangles changes a little with their amount and the window width. Not a problem with your patch. It happens without the correction too. I will try to see if there is a way to do this that does not depend on these factors

wwmm:
I found out why the space between bars was changing with window size. It was the division between two integers `width / n_bars`. After making `width` a float everything is fine. That was the problem with my old code. Now the bar width has to be reduced for the borders to be visible XD

PR #438, opened 2019-02-19
it_IT updated and full help translation

M0Rf30:
As in title. Cheers man


M0Rf30:
Can you provide some text area in PulseEffects GUI for translator credits? Thanks in advance

wwmm:
I think it can be added to the about window like it is already done for authors and artists. Which name should I put there?

M0Rf30:
Gianluca Boiano

PR #442, opened 2019-02-22
it_IT: update for latest changes

M0Rf30:


wwmm:
I did a few small changes to the help of the crystalizer plugin. This is probably the last modification with impact on translations before the next release in 1 or 2 days

PR #519, opened 2019-05-29
Improve `norewind` detection

joonatoona:
If the pulse version _might_ be 13, try creating a sink with `norewind`, and if that fails just create it normally.

### DIFF
--- a/include/pulse_manager.hpp
+++ b/include/pulse_manager.hpp
@@ -171,7 +171,8 @@ class PulseManager {
 
   std::shared_ptr<mySinkInfo> load_sink(std::string name,
                                         std::string description,
-                                        uint rate);
+                                        uint rate,
+                                        int version = -1);
 
   void load_apps_sink();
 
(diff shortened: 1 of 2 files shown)


wwmm:
Interesting idea. Thanks! :-) I think I will take it a little further ahead. Maybe it is better just trying to load with norewind without any version check and if it fails we try again with the old argument string. This should work for everybody without the need for any check.  While people are on Pulseaudio 12.2 or olders trying to load with norewind will always fail but as time goes by it will be the other way around.

It would be handy if Pulseaudio had an api to check supported module arguments. But as it doesn't trying again after failing seems to be the way to go.

joonatoona:
There's always a way :-P

You should _definitely_ not even think about using this, but I created a monstrosity that can check for supported arguments.

```
bool has_argument(const std::string &name) {
    static void *handle =
        dlopen("/usr/lib/pulse-12.2/modules/module-null-sink.so", RTLD_LAZY);
    if (!handle) return false;
    static auto get_usage = reinterpret_cast<const char *(*)(void)>(
        dlsym(handle, "module_null_sink_LTX_pa__get_usage"));
    if (!get_usage) return false;

    std::string usage(get_usage());

    return usage.find(name + "=") != std::string::npos;
}
```

wwmm:
hahaha XD It doesn't look like a good idea indeed XD

PR #522, opened 2019-06-02
Add nullptr check for server_info.channel_map

DeX77:
Pulseaudio API states that pa_channel_map_to_pretty_name will return NULL for unknown pretty names.
This results in a SEGV for me.
Fix by checking if pa_channel_map_to_pretty_name eventually is NULL.


wwmm:
Thanks! I saw this at some point when reading Pulseaudio docs but I forgot to add the check.

PR #586, opened 2019-10-18
partial sync with Debian

mikhailnov:
P.S. @hosiet , I've merged some of your changes which do not break git history by a large diff


hosiet:
Looks good to me. :ok:

PR #605, opened 2020-01-07
[WIP] German translation update

BlobCodes:
This is an update for the german translation of pulseeffects!
po/de.po is ready, the docs are at 40%
I'll bring that up to 100% and remove the WIP from the title soon :smile:


wwmm:
Thanks! :-)

PR #620, opened 2020-01-22
Update appdata to include new versions

corecontingency:
Flathub was listing the incorrect version due to the appdata file being out of date. Remember to update `/data/com.github.wwmm.pulseeffects.appdata.xml.in` when releasing a new version! :smiley:


wwmm:
Thanks! I forgot to do that. We have a script for this in the util folder but having to remember to run it is not optimal. It would be good to make meson run it somehow after a compilation.

PR #625, opened 2020-02-08
Fix mistakes in Gate plugin help

parras:
The gain reduction applies when the signal is *below* the threshold.
See https://calf-studio-gear.org/doc/Gate.html


wwmm:
Thanks!

PR #662, opened 2020-04-21
exception handler on json parsing

### LINKED ISSUE #658: Exception handler when loading presets
It occurred some time ago: I noticed that loading a preset PulseEffects got close immediately.

Afterwards I figured out the crash was because I copied an empty json file by mistake.

Anyway, I think an exception handler is missing when loading the preset and would be good to add it, avoiding to get the application crashing and maybe to inform the user that something went wrong when PulseEffects was trying to load the preset.

Digitalone1:
This resolves #658.

Successfully compiled and tested. Applying an empty file or a corrupted json, default presets are loaded with no plugin enabled.


wwmm:
Thanks! :-)

Issue #657, opened 2020-04-19
Add more directory for PulseEffects to look for presets other than ~/.config/PulseEffects


kevinsmia1939:
Hello,

I'm thinking of making a package for installing PulseEffects presets so presets can be install from linux package manager.

The presets are simply move to ~/.config/PulseEffects, however, most (if not all?) linux package manager don't allow package to be install on user directory.

How about adding /usr/share/PulseEffects for installing presets?

Thanks.

Digitalone1:
Good idea. The proper place would be `/etc/pulseeffects` and, like pulseaudio, the local config folder takes the highest priority.

Digitalone1:
@wwmm Thanks for merging the PR.

Anyway, are you sure to take in consideration only `/etc/xdg/` folder for system configuration? Most of the distributions uses `/etc/`, just like mpv. Maybe we can add `/etc/` to the array and reorder to eliminate duplicates if the environment variable already set it.

I can write the code for this.

wwmm:
I am fine with adding more paths. No problems.

Digitalone1:
Can be closed.

PR #664, opened 2020-04-24
support for system configs

Digitalone1:
PresetsManager class modified to support system global presets. Related to #657.

Search for presets located under folders returned by  `Glib::get_system_config_dirs`. I thought it was `/etc`, but is `/ect/xdg` or what's specified by `$XDG_CONFIG_DIRS` environment variable as reported in [xdg specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html).

It seems not really used, if we want to default to /etc/PulseEffects, then we can specify a vector with `/etc`.

If there are duplicates, user preset is chosen over system one. System presets can't be removed, but if one is saved, a copy is created inside user folder.

Successfully compiled and tested, at least on my system.

### DIFF
--- a/src/presets_manager.cpp
+++ b/src/presets_manager.cpp
@@ -6,9 +6,9 @@
 #include "util.hpp"
 
 PresetsManager::PresetsManager()
-    : presets_dir(Glib::get_user_config_dir() + "/PulseEffects"),
-      input_dir(Glib::get_user_config_dir() + "/PulseEffects/input"),
-      output_dir(Glib::get_user_config_dir() + "/PulseEffects/output"),
+    : user_presets_dir(Glib::get_user_config_dir() + "/PulseEffects"),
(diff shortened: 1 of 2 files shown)


src/presets_manager.cpp
```diff
@@ -415,14 +475,30 @@ void PresetsManager::autoload(PresetType preset_type, const std::string& device)
...
   if (preset_type == PresetType::output) {
-    input_file = output_dir / boost::filesystem::path{name + ".json"};
+    conf_dirs.push_back(user_output_dir);
+    conf_dirs.insert(conf_dirs.end(), system_output_dirs.begin(), system_output_dirs.end());
+
+    for (auto& dir : conf_dirs) {
+      input_file = dir / boost::filesystem::path{name + ".json"};
+      if (boost::filesystem::exists(input_file)) {
+        return true;
+        break;
```
Digitalone1, line 490: break after return, a leftover, please remove it.

Issue #578, opened 2019-09-27
Suggest plugins to install when not found


tannercollin:
To avoid the confusion I had in #577, I'd like to suggest giving the user feedback when a plugin isn't found.

A warning / error above the toggle slider would work:

![Screenshot from 2019-09-26 22-06-18](https://user-images.githubusercontent.com/6752449/65741627-fc63bb00-e0a9-11e9-8388-1497ca1a7b59.png)

It should at least tell me which plugin is needed. Bonus points for where I can get it and which type (JACK, lv2, LADSPA, etc). Docs on how to install the plugin would be great too.

Currently the UI makes no distinction between an effect that's "off" and a plugin that's missing.

Digitalone1:
That would be useful, I had issue in the past understanding which package was needed for a specific plugin.

wwmm:
I think that the credits are enough to help the user. Dealing with the way each distribution handles package names would be a nightmare.

tannercollin:
I think something to distinguish between an effect that's off and a plugin that's not installed would be handy. You could also display the plugin name instead of a distro's package name.

PR #671, opened 2020-05-02
Plugin credits

Digitalone1:
Added a line in the UI as credit to the original author of the plugins, except Crystalizer and Convolver provided by PulseEffects. 

Related to #578. This can help users to figure out which package to install if the plugin is disabled.

Translated in Italian, need to be translated in other languages.

Successfully compiled and tested.


wwmm:
Good idea. But in the particular case of the autogain plugin the message is a little misleading as libebur128 is not an autogain. It is level measurement plugin I used to make an autogain. I think that in this case `Based on` will be better than `Provided by`

PR #674, opened 2020-05-04
Update CHANGELOG.md

kevinsmia1939:
Fix /etc/PulseEffects typo


Digitalone1:
Sorry, that was my mistake.

@wwmm add also

> **Fixed**
> PulseEffects no longer crashes when an empty or corrupted preset is provided

Issue #560, opened 2019-08-22
[Feature Request] Reset only one tool/plugin (for example only Compressor settings)


Mlocik97-issues:
Reset only one tool/plugin (for example only Compressor settings)

it should be fine, if I can reset settings only in one tool/plugin.

wwmm:
One reset button for each plugin will require some work because we have many plugins. So I don't know when I will have time to implement this feature. But one day it will.

While this is not done you can use dconf-editor to view each key default value.

wwmm:
![](trash:///Screenshot%20from%202019-08-22%2011-22-28.png)

wwmm:
![Screenshot from 2019-08-22 11-22-28](https://user-images.githubusercontent.com/16225172/63522870-80d18580-c4cf-11e9-877b-a91bbc7e00f8.png)

Digitalone1:
@Mlocik97-issues @wwmm This has been implemented and can be closed.

PR #679, opened 2020-05-13
Reset a single plugin individually

Digitalone1:
This pull request implements #560 

Some templates in plugin_preset_base have been rewritten and added inside plugin_ui_base to apply default settings. A reset button has been added to each plugin UI.

A reset method has been added to each plugin UI class, except equalizer that already had one. UI classes didn't know which pipeline they belong to, so in order to apply the correct default preset, a PresetType property has been added to plugin_ui_base. In reset method this property is read to get the proper section.

The PresetType is initialized to "output" and inherited from plugin_ui_base, so it is set to "input" for all UI inside soe_ui.

Successfully compiled and tested on my system. A single plugin is reset properly to its default settings. Each plugin can be reset even if it is disabled.

I don't know if this was the simplest way to do this. I know you @wwmm don't have time implement these features, so at least I tried.

Testing this I found a weird issue. Laxity key inside Deesser is not updating. If I reset the plugin or everytime a preset is applied from the menu, laxity option stays always the previous value. This issue in not related to my modifications, I discovered it's happening in Arch upstream repository version too. Maybe it's something wrong with the plugin. Please check it out.

### DIFF
--- a/data/ui/autogain.glade
+++ b/data/ui/autogain.glade
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.22.2 -->
+<!-- Generated with glade 3.36.0 -->
 <interface domain="pulseeffects">
   <requires lib="gtk+" version="3.20"/>
   <object class="GtkImage" id="image1">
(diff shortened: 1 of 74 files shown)


wwmm:
>I don't know if this was the simplest way to do this

As this reset implementation works I did not pay much attention to the approach at the time. Now that I adding the `audiornnoise` plugin that I really looked at this in detail. Yes, there is a simpler way to reset those keys https://github.com/wwmm/pulseeffects/blob/808b94113e49a24c847d3802b54d693bda6ddb99/src/rnnoise_ui.cpp#L53. Calling settings->reset("key_name") is better. I will try to remember to simplify each plugin reset function before the next release.

wwmm:
But if you want feel free to take a look at this while #728 is being solved. I won't touch the reset functions until I finish that.

Digitalone1:
Didn't think about that at the time. If I have time, I'll try to rewrite all the reset methods.

Digitalone1:
@wwmm in your version there's not the preset type. Isn't it needed? Without won't be reset both input and output preset?

Digitalone1:
Okay, I confirm only the pipeline where you push the reset button is reset, so also preset_type in plugin_ui_base is unneeded. I will remove it.

wwmm:
The correct settings path is set when the class is instantiated. That is why there is no need to know which is the pipeline.

Issue #684, opened 2020-05-21
Convolver: Display a more intuitive frequency response?


torokati44:
Please excuse the personal anecdote:
I wanted to filter out some constant annoying beeps from a regular conference call. Given that PulseEffects has a Convolver plugin, I thought I can do this easily with it.
I went with this instead of the Equalizer or the Filter plugins because I wanted to silence just a few really specific and quite narrow bands.

There are a number of great filter design apps online, but in the end I went with SciPy to produce the impulse response of my choice, using this snippet to compute the kernel, export it into an .irs file, and plot the weights and the spectrum: https://gist.github.com/torokati44/4bdcba8ba512ac9e6ee7363eeec6af49

There might be some mathematical nonsense in it, especially right after the FFT, but the plots make intuitive sense to me, and I can verify (by looking at the output spectrum with a noise source, or listening to a sine sweep or the original problematic audio) that it indeed does exactly what I designed it to do.

These are the plots produced by the script above, in time and frequency domain:

![impulse_response](https://user-images.githubusercontent.com/288816/82615227-3a2dca80-9bca-11ea-93dd-695be4de237e.png)

![frequency_response](https://user-images.githubusercontent.com/288816/82615260-46b22300-9bca-11ea-8c2b-71d0763e38e9.png)

And these are what I'm seeing in PulseEffects instead:

![impulse_response_pe](https://user-images.githubusercontent.com/288816/82615229-3ac66100-9bca-11ea-806c-e4630b701d7a.png)

![frequency_response_pe](https://user-images.githubusercontent.com/288816/82615262-46b22300-9bca-11ea-8d8f-c8c9c82cdf39.png)

The difference between the first ones might be just because of the quite small number of curve points (I think I found a maximum of around 200 hard-coded in the plugin UI code) producing aliasing (also, the logarithmic scale of the Y axis); but the second one in PulseEffects is mostly meaningless to me. Maybe I just don't know what I should be looking for, I'm not that experienced in audio processing.

For this kind of use case at least, I think showing something similar to my second plot in Spectrum mode, with a bit more resolution, would make more sense. What do you think? Would you accept a PR that either changes the current Spectrum mode, or adds a third option?

wwmm:
Maybe I made a mistake when I used GStreamer's function to calculate the FFT. I will take a lookt. In any case feel free to make a PR improving this section

torokati44:
From what I've seen so far, there are indeed some issues with the spectrum calculation.

[...]

PR #685, opened 2020-05-22
Improvements to the Spectrum calculation in the Convolver plugin

torokati44:
Aims to solve #684.

I'm not sure about the commit that adds a square root - or, rather, if it does not conflict somehow with the normalization on the very next line.

The spectrum mentioned in the issue above now looks like this:
image

Please note that I couldn't test this as it is right now; only based on v4.7.1, because current master does not build on my machine with Boost 1.69 due to a missing header.

Also, the mouse info text now also seems to be completely wrong - both the frequency and the gain -, but that is a separate issue.

### DIFF
--- a/include/convolver_ui.hpp
+++ b/include/convolver_ui.hpp
@@ -50,7 +50,7 @@ class ConvolverUi : public Gtk::Grid, public PluginUiBase {
   boost::filesystem::path irs_dir;
 
   bool mouse_inside = false, show_fft_spectrum = false;
-  unsigned int max_plot_points = 200;
+  unsigned int max_plot_points = 1000;
   float mouse_intensity = 0.0F, mouse_time = 0.0F, mouse_freq = 0.0F;
   float min_left = 0.0F, max_left = 0.0F, min_right = 0.0F, max_right = 0.0F;
   float max_time = 0.0F;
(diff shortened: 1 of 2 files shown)


wwmm:
In order to show more points a spinbox with such control in the window will also be necessary. Even on my computer(Ryzen 3700x and Radeon RX 5700 XT) I can feel a subtle slowdown in the window when a impulse file with at least 1000 points is loaded.

wwmm:
But this is something that I can add later. Let's focus on the make the calculation correct.

torokati44:
> In order to show more points a spinbox with such control in the window will also be necessary. Even on my computer(Ryzen 3700x and Radeon RX 5700 XT) I can feel a subtle slowdown in the window when a impulse file with at least 1000 points is loaded.

Or, if the canvas control has such option, maybe the plot can be cached? It doesn't change that often.

wwmm:
I have no idea if a gtk cairo canvas can cache. It will probably be easier to add a control for the number of points and choosing a small default value. I will let the final decision on the number of points to the users.

wwmm:
But even if it can be cached once a new impulse file is loaded the canvas has to be reset to draw the new curve. The real solution for this would be to have a really smart drawing algorithm that "knows" when there is no gain in drawing more points because the user won't be able to see the difference and the performance hit is too high. But unfortunately I do not know how to do that.

PR #688, opened 2020-05-23
Update boost version requirement in meson.build

torokati44:
So it is consistent with the the actual requirement, and with the README.

This prevents builds that failed even after meson could configure the project successfully - without detecting the insufficient Boost version.


wwmm:
Thanks! I forgot that.

PR #689, opened 2020-05-23
Convolver: Fix mouse cursor info text

### LINKED ISSUE #687: Convolver: Info text on plots for mouse position is bogus
With the exception of the time in non-Spectrum mode, most information does not really correspond to the cusros position, when compared to the line drawn on the plot.

In the time-domain plot, negative values seem to be displayed in positive (maybe that computation still corresponds to a log scale?). Additionally, it is in scientific notation, which is unnecessary and inconvenient in my opinion.

In Spectrum mode, the displayed frequency does not match the line plot at all, and the gain is again, nonsense, and in scientific notation.

torokati44:
Aims to fix #687.

I'm not exactly sure about the last commit, I just didn't understand the logic behind the old calculations,
so I simplified them to these, which "look good enough". Feel free to skip that commit.

### DIFF
--- a/src/convolver_ui.cpp
+++ b/src/convolver_ui.cpp
@@ -607,11 +607,11 @@ void ConvolverUi::update_mouse_info_L(GdkEventMotion* event) {
   if (show_fft_spectrum) {
     mouse_freq = static_cast<float>(event->x) * fft_max_freq / width;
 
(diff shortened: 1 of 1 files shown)


wwmm:
GStreamer's example on the spectrum plugin usage computes the frequency axis that way. I did not put much thought into it. Just did the same. In any case the difference will be negligible.

Issue #626, opened 2020-02-09
[Feature Request] Add ability to blacklist active streams


SGOrava:
When adding something to blacklist one has to write full name (or maybe some wildcard would work?) which is tiring.
It would be very useful to have optional list of active stream as dropdown menu in blacklist dialog (also keep the text field to write custom rule).

Best regard.

Digitalone1:
@wwmm I was thinking to add a button for every app in applications list tab to add the app name in the blacklist. Do you think it's a good idea?

Maybe I'll investigate the code and see what could be done.

I was also thinking to improve the function for adding entries in the blacklist in a way that can't be added duplicate entries.

wwmm:
I do not see a problem. In case there is an icon capable of passing the "blacklist idea" without a label it would be better. But if there isn't then the usual button with label will be the way.

Each app in that list is an instance of this class https://github.com/wwmm/pulseeffects/blob/master/src/app_info_ui.cpp. In order to edit the blacklist an addiotanl variable pointing to the schema containing the blacklist has to be created https://github.com/wwmm/pulseeffects/blob/master/data/schemas/com.github.wwmm.pulseeffects.gschema.xml.

Digitalone1:
@wwmm I'm working on it, but during tests I get no audio while an app is blacklisted. Is this normal?

I also tried installing upstream version from Arch Linux repository and it's the same. I never used the blacklist, so I didn't test it so much, anyway I thought the blacklist would redirect the stream to the default sink without processing it, but it's playing silence on my system.

I tried everything. Deleted gstreamer cache, resetted PE, git and upstream version, but with blacklist there's no sound at all.

Maybe there's something wrong with my Pulseaudio configuration?

Digitalone1:
@wwmm I tested it on another system and it's not working the same. I think it's completely broken.

wwmm:
@Digitalone1 I did a few tests on my computer. There is a problem but it is not entirely on our side. Pulseaudio has a feature where it automatically moves the application to the last sink it was using. When we add the app to PE blacklist we do not try to move it anymore. But Pulseaudio is moving it to our sink again once the app starts. If the user has only this app running there is nothing telling PulseEffects to put its pipeline in the playing state as the only app playing was blacklisted. That is why there is no sound.

As we can't control this Pulseaudio feature I think that the only way to improve this is turn the app switch in our application list off before moving it to the blacklist

wwmm:
[...]

PR #692, opened 2020-05-27
Blacklist button to app entry inside applications list tab

Digitalone1:
Added a blacklist button as the title says. Related to #626 

Parts of code for adding/removing entries to/from blacklist are made unique as static method.

Right after blacklisting by the button, the application is removed from pipeline so pulseaudio can remember the last sink and apply it when the app is started and it's already in the blacklist.

@wwmm as you said, actions inside blacklist config ui can't be effective immediately, so my idea has been to left a message to inform the user that he/she has to restart/reconnect the player/microphone. I'd leave that ui so the user can see which apps are in the blacklist, otherwise he/she can't remember or can't remove one already added.  

Successfully compiled and tested for output stream, please @wwmm test for input one, I don't have a mic at the moment.

### DIFF
--- a/data/ui/app_info.glade
+++ b/data/ui/app_info.glade
@@ -1,7 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.22.1 -->
+<!-- Generated with glade 3.36.0 -->
 <interface domain="pulseeffects">
   <requires lib="gtk+" version="3.20"/>
+  <object class="GtkImage" id="blacklist_icon">
+    <property name="visible">True</property>
+    <property name="can_focus">False</property>
+    <property name="icon_name">action-unavailable</property>
+  </object>
   <object class="GtkImage" id="mute_icon">
     <property name="visible">True</property>
     <property name="can_focus">False</property>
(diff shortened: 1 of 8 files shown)


wwmm:
Everything is fine in the input devices section. By the way you do not need a mic to test it. Just run the command `gst-launch-1.0 pulsesrc ! queue ! pulsesink`. Pulseaudio will use the monitor of the current default sink as input device if no mic is found.

Digitalone1:
> Everything is fine in the input devices section. By the way you do not need a mic to test it. Just run the command `gst-launch-1.0 pulsesrc ! queue ! pulsesink`. Pulseaudio will use the monitor of the current default sink as input device if no mic is found.

Can you write also to command to disable this? Thanks.

wwmm:
I did not understand. Disable what? The `gst-launch-1.0` command?

Digitalone1:
> I did not understand. Disable what? The `gst-launch-1.0` command?

I'm not an expert of pulseaudio. That thing enabled something I think, so I was asking something else to disable. :)

wwmm:
Can you take a picture of it? I have no idea about what it could have enabled XD

Digitalone1:
No, wait, you misunderstood. :)

If you say that command will use the monitor of the current default sink as input device if no mic is found, is there another command to revert back this behavior?

wwmm:
You can give a device as argument like this `gst-launch-1.0 pulsesrc device=PulseEffects_mic.monitor ! queue ! pulsesink`. In this case it will record from the sink `PulseEffects_mic` monitor. Unless you pass a device as argument Pulseaudio will use the default source.

Usually once a mic is plugged Pulseaudio makes it the default source automatically.

PR #696, opened 2020-06-02
New blacklist logic

Digitalone1:
Hello @wwmm 

I thought about removing the add button from blacklist config ui, but I noticed that if you have two instance of the same app running and blacklist one of them and then close it, the other one will remain inside the pipeline, so when you start the same app in a new session, it won't play because PA remembers it connected the last time to PE and PE does not remove it from the pipeline. So I thought this behavior is not very optimal.

Then I inspected the code and changed something. First of all, I removed the blacklist check inside `PulseManager::newapp` so an app is always added to PE lists. The check remains in `PulseManager::changed_app` anyway.

When it comes to `on_app_added` method of SinkInputEffects/SourceOutputEffects class, we check if the app is connected. If true, the app is removed from pipeline when blacklisted. If it's not connected, it's moved to pipeline if it's not in the blacklisted and "process all output/input" is checked. 

So this initial state has to be reflected into the `app_info_ui`. The class has been modified to handle settings and save "enabled" and "blacklisted" state. This way, if the app is blacklisted, it will be shown inside the app list tab, but blacklist button is toggled. If it starts already disabled, the switch reports the correct state. The class has also another boolean to save the previous state before the blacklist has been toggled, so it can be restored when the button is untoggled.

This behavior seems more coherent. The blacklist works when you add the app from the blacklist config ui. No matter how the app left the pipeline, whether it's connected or not, if you have it in the blacklist, it will start disabled. On the other hand, the app will start enabled if you don't have it in the blacklist and you left it enabled or process all output is checked.

I know someone will complain because app blacklisted will show in the list, but they will not be processed by the pipeline and this seems the only reasonable way to not having apps pointing to PE without getting processed and getting no sound.

Please, test on your system. I did on mine and seems everything okay. If there's an easy method to do this, let's discuss.


wwmm:
I though I would have time to test this today. I didn't XD. I will try to take a look at ti tomorrow. In any case one solution for the people that will complain would be to have an additional setting in the blacklist menu like this `Show Blacklisted Applications in the Main Window`. If this is off by default  I think we could make everybody happy.

Digitalone1:
> an additional setting in the blacklist menu like this `Show Blacklisted Applications in the Main Window`. If this is off by default I think we could make everybody happy.

It could be an additional feature.

Digitalone1:
Hello @wwmm, I'm testing some ways to hide blacklisted apps from app list, but I'm experiencing some weird issues.

First of all, If I test PE in a different location (i.e. /tmp/pe), the executable can't locate new options specified inside `com.github.wwmm.pulseeffects.gschema.xml` file. So I'm forced to test overwriting existing PE files installed by pacman.

Another weird thing is that if I make a modification to configuration glades ui, I see this modifications even if I install the previous pulseeffects-git of upstream package. I thought schema files were overwritten by pacman during the reinstall, or am I wrong?

At last, just for curiosity, modifications to gsettings schema are not saved in schema xml files. These files contains only enumerations and default values. So do you know where real configuration data is stored? dconf specific files?

wwmm:
I think you have to install to a path where gsettings searchs for schemas. Try to install in `/usr/local`. It will probably be necessary to uninstall the repository package first.

>Another weird thing is that if I make a modification to configuration glades ui, I see this modifications even if I install the previous pulseeffects-git of upstream package. I thought schema files were overwritten by pacman during the reinstall, or am I wrong?

Modifications to the widgets in the interface or to the gsetting xml files?

>At last, just for curiosity, modifications to gsettings schema are not saved in schema xml files. These files contains only enumerations and default values. So do you know where real configuration data is stored? dconf specific files?

I don't know where it saves these settings. Removing `~/.config/dconf/user` did not reset PE settings. As there is a `dconf-service` running maybe user settings are managed in a system folder.

Digitalone1:
> Modifications to the widgets in the interface or to the gsetting xml files?

Widgets in the interface.
Installed upstream version and I'm seeing this:

<details>
<summary>Screenshot</summary>

Schermata del 2020-06-12 16-36-11

</details>

Digitalone1:
Maybe I found why. If you see [here](https://www.archlinux.org/packages/community/x86_64/pulseeffects/), the installation does not install glades files. So where glades ui are loaded from?

If I make a change to UI for test and then I get back to stable version, I see the previous testing changes.

wwmm:
Glade files are embedded to the executable file and accessed through `gresources`:
https://github.com/wwmm/pulseeffects/blob/master/data/pulseeffects.gresource.xml
https://github.com/wwmm/pulseeffects/blob/d63b8662c1785ae51e9ccc9d555bcd58d9c6b53e/data/meson.build#L90

wwmm:
This process happens when meson generates the executable

Digitalone1:
Found the culprit. Ninja installed other copies under `/usr/local/` while I thought it was overwriting existing files installed by pacman.

PR #706, opened 2020-06-13
Option to show/hide blacklisted apps

Digitalone1:
Added a switch to show/hide blacklisted apps in Applications List.

Other things fixed, blacklist management methods made static, updated Italian translation and Spectrum config UI centered vertically like other config tabs.

Successfully compiled and tested on my system.

@wwmm, this needs to be installed overriding existing files otherwise new gsetting config won't be detected.

### DIFF
--- a/data/schemas/com.github.wwmm.pulseeffects.gschema.xml
+++ b/data/schemas/com.github.wwmm.pulseeffects.gschema.xml
@@ -206,6 +206,9 @@
         <key name="custom-source" type="s">
             <default>""</default>
         </key>
+        <key name="show-blacklisted-apps" type="b">
+            <default>false</default>
+        </key>
         <key name="blacklist-in" type="as">
             <default>[]</default>
         </key>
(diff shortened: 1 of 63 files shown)


wwmm:
I had to make a small modification because of the error `g_object_set_qdata_full: assertion 'quark > 0' failed`. The reason was trying to initialize the `settings` variable statically. For some reason I do not know there are a few functions in the glib ecosystem that can not be used that way. A glib loop must be running when they are called. The easiest solution was to make the variable `settings` local to each static function. As they won't be called often doing a connection to gsettings in every call should not cause performance problems.

wwmm:
By the way I could test without removing or overwriting my PE install. All that have to be done in this case is running `glib-compile-schemas data/schemas/` inside PE source folder before executing the binary created in the meson build folder. The steps would be like this:

```
cd pe_sources
glib-compile-schemas data/schemas/
meson _build
cd _build
ninja
cd src
source ../../util/environmental_variables.sh
./pulseeffects
```

Digitalone1:
No problem. 

Another way could be making blacklist_settings_ui a singleton.

PR #734, opened 2020-07-22
New parameters in Limiter and Gate

### LINKED ISSUE #732: Limiter Auto-Leveling
I discovered that calf limiter has a bool option called auto-level:

```
auto-level          : Auto-level
                        flags: leggibile, scrivibile, controllabile
                        Boolean. Default: false
```

Looking the limiter code, it's specified that the plugin is doing *automatic makeup gain*, so we reduce the output level accordingly binding both `output-gain` and `limit` to the same gsettings key.

But from version 0.90.2 this new option was introduced that disables auto-level by default. See changelog [here](https://freshcode.club/projects/calf).

```
Limiter: Allow toggling auto-leveling. . . . Move all limiter auto-level toggles to header.
```

So this behavior is not needed anymore and we should expose this toggle in the UI (along with the output gain) to let the user enable auto leveling, right?

Besided, I was looking the gsettings schema and sourceoutput limiter has other unused options I don't know:

```
        <key name="autovolume-state" type="b">
            <default>false</default>
        </key>
        <key name="autovolume-window" type="d">
            <range min="1" max="1000"/>
            <default>1000.00</default>
        </key>
        <key name="autovolume-target" type="i">
            <range min="-99" max="-4"/>
            <default>-12</default>
        </key>
        <key name="autovolume-tolerance" type="i">
            <range min="1" max="4"/>
            <default>1</default>
        </key>
        <key name="autovolume-threshold" type="i">
            <range min="-99" max="0"/>
            <default>-50</default>
        </key>
```

Was the limiter UI using a different plugin in the past?

Digitalone1:
Added `auto-level` and `output-gain` in Limiter, plus `input-gain` in Gate. Fixes #732 

@wwmm I added a margin between input and output bars in Gate. I'm seeing it good on my theme, but please check if it's showing fine on yours. If no, I'll undo the changes, otherwise it could be replicated in other plugin ui since I think that separating input from output bars is better.

### DIFF
--- a/data/schemas/com.github.wwmm.pulseeffects.sinkinputs.gate.gschema.xml
+++ b/data/schemas/com.github.wwmm.pulseeffects.sinkinputs.gate.gschema.xml
@@ -44,6 +44,10 @@
             <range min="0" max="18"/>
             <default>9</default>
         </key>
+        <key name="input" type="d">
+            <range min="-36" max="36"/>
+            <default>0.0</default>
+        </key>
         <key name="makeup" type="d">
             <range min="0" max="36"/>
             <default>0.0</default>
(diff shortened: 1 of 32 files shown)


Digitalone1:
I applied margins to output and input bars with a better solution that should work with all themes. Please, check.

wwmm:
It is fine on the default gtk theme.

Issue #588, opened 2019-11-06
Importing generic EQ presets


Shurichan:
[left4.txt](https://github.com/wwmm/pulseeffects/files/3813097/left4.txt)
I believe that this generic format originated from Eq APO? Turns out that importing these to PA manually every time is a massive pain!
I looked into using a script to insert the EQ data to the PA json files, but parsing the files seems to be beyond me.
PA already appears to aim for comparability with APO, so this would be a great add IMO.

polygonalenippel:
I second this.

An often used tool is AutoEQ: https://github.com/jaakkopasanen/AutoEq/
It outputs several formats, one of which is a parametric EQ which is very similar to what you posted:
```
Filter 1: ON PK Fc 52 Hz Gain -6.6 dB Q 0.97
Filter 2: ON PK Fc 139 Hz Gain -1.8 dB Q 3.07
Filter 3: ON PK Fc 171 Hz Gain -0.4 dB Q 2.41
Filter 4: ON PK Fc 210 Hz Gain 4.8 dB Q 2.00
Filter 5: ON PK Fc 522 Hz Gain 0.5 dB Q 2.47
Filter 6: ON PK Fc 2066 Hz Gain -0.9 dB Q 0.71
Filter 7: ON PK Fc 2432 Hz Gain -3.3 dB Q 2.40
Filter 8: ON PK Fc 3665 Hz Gain 9.0 dB Q 2.17
Filter 9: ON PK Fc 6467 Hz Gain -5.2 dB Q 1.48
Filter 10: ON PK Fc 19607 Hz Gain -7.1 dB Q 0.32
```
Right now there is no way of importing this. You could use a simple perl script or `jq` and `awk` perhaps to parse it into a PE config json, but yeah.

I suggest adding a button in the Equalizer screen to import such file and parse it into the current preset.

Some useful links:
AutoEQ implementation: https://github.com/jaakkopasanen/AutoEq/blob/aa25ed8e8270c523893fadbda57e9811c65733f1/frequency_response.py#L675
APO config reference: https://sourceforge.net/p/equalizerapo/wiki/Configuration%20reference/

Note: the preamp is basically the Input slider of the equalizer

Also, maybe it's a good begin to support just PK Fc filters since that's what AutoEQ does as well: https://github.com/jaakkopasanen/AutoEq/issues/102

Shurichan:
The newest standalone version of the used LSP Parametric Equalizer plugin does have support for importing a REW file.
[...]

PR #742, opened 2020-08-05
APO filter preset support

polygonalenippel:
Addresses #588

### DIFF
--- a/include/equalizer_ui.hpp
+++ b/include/equalizer_ui.hpp
@@ -3,6 +3,29 @@
 
 #include "plugin_ui_base.hpp"
 
+enum FilterType : const unsigned int {
+  PEAKING = 1,
+  LOW_PASS = 1 << 1,
+  LOW_PASS_Q = 1 << 2,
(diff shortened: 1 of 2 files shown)


include/equalizer_ui.hpp
```diff
@@ -56,7 +79,15 @@ class EqualizerUi : public Gtk::Grid, public PluginUiBase {
 
   void on_import_apo_preset_clicked();
 
+  bool parse_apo_filter(const std::string& line, struct ImportedBand& filter);
+
   void import_apo_preset(const std::string& file_path);
+
+  std::unordered_map<std::string, FilterType> const FilterTypeMap = {
```
polygonalenippel, line 86: This is ugly but gets the job done. If anyone has a better idea on how to better parse this godforsaken spec, I'll gladly accept it.

wwmm:
I think it is ok. I doubt we will ever have full compatibility to APO presets. What we have now is way better than have no support.

PR #747, opened 2020-08-08
Change order of plugins in Readme to same order, as default order in PE.

Mlocik97:


Mlocik97:
Btw. I think, We can delete old files, for example screenshot from PE that were created years ago.

https://github.com/wwmm/pulseeffects/tree/master/images

wwmm:
>We can delete/replace old files, for example screenshot from PE that were created years ago.

We can not delete as some are used in the readme and others by the flatpak package. The only option is replacing by newer ones.

PR #767, opened 2020-08-28
Code optimizations

### LINKED ISSUE #622: Improve construction of numbered strings
Numbered strings are constructed in [the function “gst_pecrystalizer_class_init”](https://github.com/wwmm/pulseeffects/blob/d8e6aee0575274c82d9640b9374cdec997ed09c0/src/crystalizer/gstpecrystalizer.cpp#L151 "Update candidate").
I suggest to improve this implementation detail.
* Add two local variables for storing generated strings.
* Append desired data to the corresponding string object instead of using the plus operator together with temporaries.

Digitalone1:
Fixes #622


elfring:
I find the pull request comment questionable.

PR #768, opened 2020-08-31
UI optimizations

Digitalone1:
I revised all plugin ui to apply same style to levelbars. Some plugins, like deesser, were redesigned, most of the times homogeneous property was applied to modify space between spinbuttons to fit the window horizontal length.

Also settings ui have been reworked a bit.

@wwmm let me know if there's something you don't like.


wwmm:
>let me know if there's something you don't like

It is good. No problems.

PR #793, opened 2020-09-28
Show sign on global level meter

Digitalone1:
I wanted to test std::showpos on global level meter and I liked it.

Not really needed, but the plus sign shows better that the clipping is happening above 0, combined with the saturation icon.

Anyway, I didn't like the fact that std::showpos shows the positive sign also on zero, which is unsigned. So I modified it.

Then I made some adjustments on step increments that I previously forgot.


wwmm:
Interesting. I did not know about this function.

PR #799, opened 2020-10-07
Positive sign on gain/intensity EQ/Cryst; locale formatted values

Digitalone1:
- scale drawn value have been hided in equalizer and crystalizer
- a new label has been added to show the gain/intensity property in equalizer/crystalizer
- make use of `level_to_str`/`level_to_str_showpos` functions to format values where possible, adding also a float version to both
- show positive sign for gain/intensity values in equalizer/crystalizer
- widget values are formatted according to locale (i.e on my Italian system, decimal separator is the comma) while `level_to_str` strings weren't (always point separator), so they have been modified to return a locale value also
- band columns in equalizer/crystalizer ui have been modified to expand in homogeneous columns having the same space between them like widgets in all other plugin ui

The only issue I encountered is with gain/intensity values not been updated when the initial value is 0. I don't know why this was happening, but I added a workaround initializing the label in the band constructor.

### DIFF
--- a/data/ui/crystalizer.glade
+++ b/data/ui/crystalizer.glade
@@ -150,7 +150,7 @@
           <object class="GtkGrid" id="bands_grid">
             <property name="visible">True</property>
             <property name="can-focus">False</property>
-            <property name="halign">center</property>
+            <property name="hexpand">True</property>
             <property name="vexpand">True</property>
             <property name="column-spacing">12</property>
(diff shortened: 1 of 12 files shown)


Digitalone1:
Noticed now also latency in app info ui is not formatted according to locale. I'll make another commit.

Digitalone1:
Made also for latency/buffer in `app_info_ui`

PR #805, opened 2020-10-21
Regex optimized

Digitalone1:
I tried to write regex in `parse_apo_filter` with possessive quantifiers to avoid backtracking. Then I added the case insensitive option to make them more robust.

I made some tests, seems it's working. Please @wwmm, test you too.


wwmm:
As far as I could see it is working.

PR #824, opened 2020-11-13
Fixed Loudness plugin and UI changes

### LINKED ISSUE #819: Reference Signal in Loudness Compenstation Plugin is useless, Input value is not saved to preset
I think, we can remove that reference signal thing, not only PE has already testing signals feature with much more settings, but also, reference signal from Loudness Plugin doesn't work, if there is nothing in input (no App is giving audio to PE).

Input value is not saved to preset.

Limiter docs are not updated (doesn't contain autoleveling feature, but still has things from older version (4.7.x)

Digitalone1:
* Input value saved to preset
* Reference Signal removed (test signals can be used in its stead)

Fixes #819

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
   property has been applied to get the same space between band sliders.
 - App info and plugin interfaces made scrollable: now PulseEffects window does not change its size when switching beetween
   plugins.
+- Loudness plugin: removed reference signal button; input value is now saved to preset.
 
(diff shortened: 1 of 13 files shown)


Digitalone1:
Added other commits. 

One to fix [this](https://github.com/wwmm/pulseeffects/issues/783#issuecomment-692872029) issue. Now preset label is reduced when too long after checking on a max length.

The other to make a new blocklist row disposition and max width set.

PR #827, opened 2020-11-17
Simplyfied reset method in plugin UI classes

Digitalone1:
Implemented suggestions mentioned in #679


Digitalone1:
Opps, I forgot to remove the previous templates. I'll make another commit.

PR #830, opened 2020-11-21
RNNoise redesigned and UI changes

Digitalone1:
Redesigned RNNoise UI a bit: model list frame is only visible when one or more models are added.

Then used ellipsize to limit string labels, so deleted the custom methods I previously added.

### DIFF
--- a/data/ui/application.glade
+++ b/data/ui/application.glade
@@ -215,7 +215,9 @@
                 <property name="visible">True</property>
                 <property name="can-focus">False</property>
                 <property name="label" translatable="yes">Presets</property>
+                <property name="ellipsize">end</property>
                 <property name="single-line-mode">True</property>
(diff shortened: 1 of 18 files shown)


wwmm:
I don't  know if it is only me but the file name label after the buttons is a little weird XD

Digitalone1:
> I don't know if it is only me but the file name label after the buttons is a little weird XD

Can be changed. It seemed better to me when the file name too long is truncated to the right. 

Schermata del 2020-11-21 21-42-43

Can be modified if you want.

wwmm:
I prefer the old widget order XD

wwmm:
The problem with the current order is that my eyes always go to the "remove button" first instead of the file name label XD

Digitalone1:
I'll change it.

PR #836, opened 2020-11-22
Moved APO button and documentation updated

Digitalone1:
Apo preset button is moved into equalizer preset section.
Documentation and Italian translation updated.

### DIFF
--- a/data/ui/equalizer.glade
+++ b/data/ui/equalizer.glade
@@ -109,15 +109,7 @@
     <property name="step-increment">0.1</property>
     <property name="page-increment">1</property>
   </object>
-  <object class="GtkAdjustment" id="nbands">
-    <property name="lower">1</property>
-    <property name="upper">30</property>
-    <property name="value">30</property>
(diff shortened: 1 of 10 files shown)


wwmm:
It makes sense as the button action is related to presets. But I wonder if we are going to create wrong expectations in the users. In other places the import button adds something to the list that is next to it. But in this case that is not going to happen because that preset list only shows the presets embedded in PE binary.

Digitalone1:
> It makes sense as the button action is related to presets. But I wonder if we are going to create wrong expectations in the users. In other places the import button adds something to the list that is next to it. But in this case that is not going to happen because that preset list only shows the presets embedded in PE binary.

I thought about this. If you want I can re-move the button in the previous section. Or we can modify the description in `Apply APO Profile`. Maybe in that case it's clear that you apply the profile, but don't import it inside the list.

wwmm:
>Or we can modify the description in Apply APO Profile

I did not think about this. It should be enough.

PR #851, opened 2020-12-13
Locale following pipewire development

Digitalone1:
As the title says.

Some work still has to be done when porting pipewire branch, but at least this has quite the same behaviour.


Digitalone1:
`c++: fatal error: Killed signal terminated program cc1plus`

Successfully compiled two times on my system. I also installed from my repository with custom PKGBUILD `makepkg -sri`

PR #853, opened 2020-12-14
Equalizer: added allpass filter and changed gain slider

Digitalone1:
As the title says.

@wwmm with this changes and the previous fixes to equalizer and locale, maybe a new release before Christmas is needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+### Added
+- Added Allpass filter in Equalizer plugin.
+
 ### Fixed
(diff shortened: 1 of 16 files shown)


wwmm:
>maybe a new release before Christmas is needed

It makes sense. Do you intend to do more changes before that? If not I can release it today.

The pipewire branch is already able to apply effects to streams output :D. I did not work on the mic pipeline yet so no mic effects for now.

Digitalone1:
No, but perhaps some other issues will came out. At least you can release in the weekend.

wwmm:
>At least you can release in the weekend.

Ok. I have no reason to rush it.

PR #857, opened 2020-12-21
Changes for legacy version

Digitalone1:
Updated changelog and PKGBUILD


Digitalone1:
PKGBUILD has been modified replacing libpulse with pulseaudio, so the legacy version will require it rather than Pipewire.

Digitalone1:
@wwmm After the creation of the legacy branch, please update the PKGBUILD with the url of the new branch.

Issue #882, opened 2021-01-23
Pulseeffects stopped working on Manjaro KDE


das22:
Hi

Apologies if this was already posted but I didn't notice this exact issue. I've used Pulseeffects without issue (Manjaro KDE, 5.10 kernel) for a year or so but for a few weeks now it doesn't work. Specifically, the version 4.8.3 from the official Manjaro repo doesn't start at all (no GUI) and when I try to lunch from the terminal, I get:

`terminate called after throwing an instance of 'std::runtime_error'
  what():  locale::facet::_S_create_c_locale name not valid
Aborted (core dumped)
`

I've tried installing a 4.8.4 from the AUR and GUI opens up but doesn't "see" sound to process (e.g. if I have audio running from a browser or video player, the app no longer shows up in Pulseeffects). It also doesn't show up as an audio device next to the usual mic/speakers shown by the system's Audio app. 

From googling around, I get the impression this is pipewire related (I see it installed on my system though pipewire-pulse mentioned in some discussions isn't) but I don't understand enough to fix it on my own. Is there a solution?

Thank you!

wwmm:
What is the output of the command `locale` in your system? Recently a user had the same error #878 and the cause was a bad locale configuration.

PipeWire became our server in version `5.0.0`. Anything before that uses Pulseaudio and not PipeWire.

wwmm:
If you install `pipewire-pulse` Pulseaudio will be removed and your system will be on PipeWire. In this case you will need PE `5.0.0`. But I think that you have a locale problem.

das22:
Thanks for the quick reply! Here's my locale output. Its a mess because of using a German laptop (with German keyboard layout) while wanting to retain English as system language so maybe this is indeed the source of the issue. Any idea what/how to fix it?

`locale: Cannot set LC_CTYPE to default locale: No such file or directory
locale: Cannot set LC_MESSAGES to default locale: No such file or directory
locale: Cannot set LC_ALL to default locale: No such file or directory
LANG=en_DE.UTF-8
LC_CTYPE="en_DE.UTF-8"
LC_NUMERIC=en_GB.UTF-8
LC_TIME=en_GB.UTF-8
LC_COLLATE="en_DE.UTF-8"
LC_MONETARY=en_GB.UTF-8
LC_MESSAGES="en_DE.UTF-8"
LC_PAPER=en_GB.UTF-8
LC_NAME=en_GB.UTF-8
[...]

Issue #881, opened 2021-01-23
Pulse effects doesn't show any applications


kitty-alt:
Pulseeffects (git version) on Manjaro no longer shows any audio sources, and therefore doesn't intercept any audio and apply effects, or do anything. But hey at least you made the ui labels more "inclusive"...

wwmm:
>But hey at least you made the ui labels more "inclusive"...

In case you did not notice this kind of comment is too offensive for someone seeking help...

kitty-alt:
> > But hey at least you made the ui labels more "inclusive"...
> 
> In case you did not notice this kind of comment is too offensive for someone seeking help...

It's offensive to be annoyed that the latest iteration of a very good application doesn't seem to function anymore, but has refactored its wording for other people who find common terminology offensive? *I apologize.* That aside, [pulseeffects-git](https://aur.archlinux.org/packages/pulseeffects-git/) does seem to be non functional on arch/manjaro currently, at least on my system. This might be out of scope for issues here as it may just be a packaging related issue.

wwmm:
You have the right to be annoyed when an application does not work. No problems with that. But disdain the work made for free by volunteers is not right... Specially when the mentioned work has been done 4 releases ago and has nothing to do with the latest releases...

Since version `5.0.0` we use Pipewire instead of Pulseaudio. My bet is that your system is using Pulseaudio instead of Pipewire as sound server. In that case install `4.8.4` or edit the pkgbuild so it gets the code from our legacy Pulseaudio branch https://github.com/wwmm/pulseeffects/tree/pulseaudio-legacy. This branch will receive bug fixes for critical bugs when necessary.  So if PipeWire is not for you yet staying on the legacy branch is fine.

Digitalone1:
> But hey at least you made the ui labels more "inclusive"...

Please, delete this sentence.

If you're still on plain PulseAudio, `makepkg -sri` on PKGBUILD provided in the branch above by wwmm, or the one in [my repo](https://github.com/Digitalone1/pulseeffects-1) updated on the last rnnoise commit.

If Manjaro will switch on Pipewire, change to `pulseeffects-git` from AUR.

wwmm:
>or the one in my repo updated on the last rnnoise commit

I think it is possible to do a pull request to the legacy branch. If you want to do it feel free to go ahead.

Digitalone1:
[...]

PR #883, opened 2021-01-23
Updates to Legacy Branch

Digitalone1:
Related to #881 and #882

### DIFF
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -2,8 +2,9 @@
 # Contributor: Filipe Laíns (FFY00) <lains@archlinux.org>
 # Contributor: Wellington <wellingtonwallace@gmail.com>
 
+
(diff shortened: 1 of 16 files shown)


wwmm:
Let me know when you think that the legacy branch got enough changes to justify a new release. It seems to me we can wait a little longer for that.

Digitalone1:
Excluding a little cleanup on header directives, moving the locale definition to util and changing two methods to templates, the only substantial modification is the rnnoise directive without extern keyword. 

If the new rnnoise version installed on the system requires the normal directive, a release is needed, otherwise not.

wwmm:
>If the new rnnoise version installed on the system requires the normal directive

I don't think it does. The C preprocessor will just ignore the `ifndef` in older PulseEffects versions if a newer rnnoise header is used.

PR #915, opened 2021-03-09
Update build dependencies

guyru:
See https://github.com/wwmm/pulseeffects/issues/914


wwmm:
It seems fine to me. But as @mikhailnov is the one using and maintaining these files I would like to have his opinion.

mikhailnov:
Sorry for not responding for long. I see a new 4.x release has been made. That's good. This PR should have probably remove libpulse-dev. Let me try to package both 4.x and 5.x and send PRs.

tilgovi:
I started to give this a shot today and got stuck. There's no libglibmm-2.6-dev or libgtkmm-4.0-dev.

tilgovi:
It also doesn't appear that meson 0.57.0 exists in Debian yet. If it's not strictly necessary, dropping back to 0.56.0 and c++2a might work for the Debian build.

wwmm:
>It also doesn't appear that meson 0.57.0 exists in Debian yet. If it's not strictly necessary, dropping back to 0.56.0 and c++2a might work for the Debian build.

If I am not mistaken the reason was having the support for the c++20 build flag https://github.com/mesonbuild/meson/pull/8085. @vchernin was there any requirement on Flatpak for Meson 0.57?

vchernin:
Using meson >= 0.57 indeed solved the Flatpak issue. I recall nothing worked until I updated meson, so I'm not sure if it's possible to compile without it.

There's maybe some other way I'm unaware of but for Flatpak temporarily bundling meson did work.

tilgovi:
Ah, I meant to suggest that if it's possible to compile for Debian with a lower version, we could patch the version for Debian builds only. It's okay if a newer version is needed for Flatpak builds.

wwmm:
>Ah, I meant to suggest that if it's possible to compile for Debian with a lower version, we could patch the version for Debian builds only.

Oh! I see. If Meson 0.56 has the patch adding support for c++20 it should work.

tilgovi:
I'm not that well versed in C++, but 0.56 had support for c++2a as a target. I don't know whether that would functionally mean c++20 or not at this point, or if the flags that propagate to the build result in something meaningfully different.

eli-schwartz:
> I'm not that well versed in C++, but 0.56 had support for c++2a as a target. I don't know whether that would functionally mean c++20 or not at this point, or if the flags that propagate to the build result in something meaningfully different.

Well, see e.g. the GCC manpage:

```
    c++20
    c++2a
        The 2020 ISO C++ standard plus amendments. Support is experimental, and could change in incompatible
        ways in future releases. The name c++2a is deprecated.
```

And, https://en.wikipedia.org/wiki/C%2B%2B20#Compiler_support

So passing either one has the same effect on newer versions of GCC, but 2a also works on older versions of GCC (i.e. GCC 9). If I understand correctly, the final name is implemented as support matures, and the deprecated name is kept for compatibility.

The same aliases exist for c++1z (c++17) and c++1y (c++14) and c++0x (c++11).

...

However, MSVC on Windows, if that were a concern of yours, does NOT include these preview names. Does it matter?

tilgovi:
Ubuntu 22.04 has Meson 0.61.2, so I gave another shot at this. Unfortunately, neither Ubuntu nor Debian has libsigc++3, only libsigc++2. More problematic is that they don't have rnnoise at all (see https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=980839). It looks like this was optional back in pulseeffects v4, but it's not in easyeffects v6, so I'm a bit blocked.

tilgovi:
I can push a branch with what I have, though, if it's helpful.

tilgovi:
I'll try to add patches to a debian patch queue to drop the rnnoise plugin and see if that works.

tilgovi:
I've got a work in progress on this, as I've just mentioned in #1000. Can we close this and consolidate discussion there? I'd love any help.

wwmm:
>Can we close this and consolidate discussion there?

Yes

PR #924, opened 2021-03-17
presets manager was loading wrong values from blocklists

PeeK1e:
The presets_manager.cpp was loading the wrong values from blocklists specified in the preset .json files. 
The output-blocklist was loading the input-blocklist and vice versa.
Fixed that by just swapping out the conditions in the switch-case.

I also added an entry for people using CLion/Jetbrain IDE's in the .gitignore, so the .idea folder created by the IDE will be ignored

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,3 +52,6 @@ debian/*
 
 ppastats
 .autobuild
+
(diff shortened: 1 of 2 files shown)


wwmm:
Thanks!

PR #934, opened 2021-03-28
Make key const.

kevincox:
Otherwise there is an error in C++ mode for converting a string literal to a non-const string.

```
../src/plugin_base.cpp: In constructor 'PluginBase::PluginBase(std::string, std::string, const string&, const string&)':
../src/plugin_base.cpp:146:30: warning: ISO C++ forbids converting a string constant to 'gchar*' {aka 'char*'} [-Wwrite-strings]
  146 |   on_state_changed(settings, "state", this);
      |                              ^~~~~~~
```

### DIFF
--- a/src/plugin_base.cpp
+++ b/src/plugin_base.cpp
@@ -23,7 +23,7 @@
 
 namespace {
 
-void on_state_changed(GSettings* settings, gchar* key, PluginBase* l) {
+void on_state_changed(GSettings* settings, const gchar* key, PluginBase* l) {
   if (l->plugin_is_installed) {
     int enable = g_settings_get_boolean(settings, key);
 


wwmm:
Thanks! I do not remember having seen this warning. Strange.

In case you have any intention of taking a deeper look at our code have in mind that the next big release will not use GStreamer in the effects chain. I am replacing it by native Pipewire filters in the GTK4 branch. So changes in this part of the code will not carry on to the next release.

kevincox:
I noticed it because I was building on NixOS (https://github.com/NixOS/nixpkgs/pull/117913) and the build was failing. It is odd that it is failed because the error message says that it is a warning, but maybe it was being build in a strict C++ mode?

So for now I'm not poking too deep, just trying to make it build.

wwmm:
I don´t know. I really do not remember seeing that warning when compiling on Arch Linux. Maybe it is the compiler version.

PR #958, opened 2021-05-03
Display icon in window list and alt+tab menu

FreaxMATE:
This PR allows the pulseeffects icon to be displayed in window lists, window selectors, the alt+tab menu, ... (MATE Desktop Environment, possibly in  others as well)

**Before:**
before_panel

before

**Ater:**
after_panel

after

Possibly fixes https://github.com/wwmm/pulseeffects/issues/546


wwmm:
Thanks! For some reason GNOME does not need this property to be set. So I was not aware people were having this issue.

PR #997, opened 2021-07-07
Update german translation

BlobCodes:
I haven't tested these because Fedora 34 can't compile easyeffects (glibmm-2.68 not available yet).
Should work however.. (Still, it would be appreciated if someone could test it out)

Btw: I think the following text needs to be changed to mention PipeWire and not PulseAudio:
> #: data/com.github.wwmm.easyeffects.appdata.xml.in:12
> msgid ""
> "Because EasyEffects uses the default PulseAudio sound server it will work "
> "with most, if not all, applications you use. All supported applications are "
> "presented in the main window, where each can be enabled individually."

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -8,42 +8,40 @@ msgstr ""
 "Project-Id-Version: EasyEffects\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2021-07-06 14:44-0300\n"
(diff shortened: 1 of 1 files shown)


wwmm:
>Btw: I think the following text needs to be changed to mention PipeWire and not PulseAudio:

I agree. As I usually do not look at the files related to flatpak I did not notice this. Thanks!

PR #1032, opened 2021-07-17
First attempt to port LSP MBC

Digitalone1:
As the title says


wwmm:
I will change some of the names in the schema to match the ones in the lsp file `/usr/lib/lv2/lsp-plugins.lv2/mb_compressor_stereo.ttl`. At the end of the day it is the name in that file that matters and not the ones in lsp user docs.

wwmm:
I will also number the bands from 0 to 7 like it is done in the same file

wwmm:
>I will also number the bands from 0 to 7 like it is done in the same file

Never mind. The file names the `enable` parameter as `cbe_1` and it does not start from 0. I thought this had 8 bands. But there is only 7 `cbe_` parameters...

wwmm:
But things like `sidechain-source` and `sidechain-mode` in mb_compressor_stereo.ttl start from zero... That is confusing...

Digitalone1:
Bands are from 0 to 7, but the first is always enabled and the split frequencies are 7 (they were 3 for calf 4 bands compressor).

wwmm:
Ok. The native plugin interface is a little misleading in this regard. Looking at it it seems that all the bands can be enabled/disabled.
I also did not see where the native interface is setting the split-frequency for each band. I can see the LCF and HCF there. But this seems something else.

wwmm:
Now I see that the file has parameters named like `Compressor enable 0`. This must be what is in the native interface. Do the bands can be turned on/off independently of the compressors? Not what I was expecting.

Digitalone1:
> I also did not see where the native interface is setting the split-frequency for each band. I can see the LCF and HCF there. But this seems something else.

In split tab you can set the split frequencies. Then frequencies can be modified by LCF and HCF (if LCF and HCF are disabled, their values are assumed equal to the split frequencies).

Digitalone1:
> Now I see that the file has parameters with human names like `Compressor enable 0`. This must be what is in the native interface. Do the bands can be turned on/off independently of the compressors? Not what I was expecting... And I am not sure it makes sense...

I think this is an additional enable parameter. Each band can be enabled individually (except band 0 always enabled), then the "compressor enable" option can enable/disable the compression for that specific band (this means you can disable the compression for band 0, even if the band itself is always enabled). I forgot to add this parameter to schemas, I will do it.

wwmm:
I wonder why he had to take such a complicated approach. Why not keep all the bands enabled and just enable/disable the compressors? Could it be because of cpu usage? It feels redundant to have both controls.

Digitalone1:
This branch of mine can be deleted since I'm working on another one.

PR #1043, opened 2021-07-25
Update AppStream screenshots

vchernin:
This updates the AppStream screenshots to the ones used in the [readme](https://github.com/wwmm/easyeffects/blob/master/README.md), which should be reflected in most Linux GUI software managers. The [Flatpak](https://github.com/flathub/flathub/pull/2429) will also benefit from this.

I'll try to revisit the captions if I or anyone has ideas for screenshot captions.

[AppStream screenshot documentation](https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html#qsr-app-screenshots-info).
`appstreamcli validate --pedantic` complains about the lack of captions, but those are optional (a pedantic complaint).
[I used this tool to help](https://www.freedesktop.org/software/appstream/metainfocreator/#/guiapp).

As a side note, what do you think of the AppStream description? I noticed a typo ("tools" > "tool"), is everything else in the description still accurate, or is there a better place to source a description?

### DIFF
--- a/data/com.github.wwmm.easyeffects.appdata.xml.in
+++ b/data/com.github.wwmm.easyeffects.appdata.xml.in
@@ -17,26 +17,15 @@
   <url type="homepage">https://github.com/wwmm/easyeffects</url>
   <url type="bugtracker">https://github.com/wwmm/easyeffects/issues</url>
   <url type="help">https://github.com/wwmm/easyeffects/blob/master/README.md</url>
-  <screenshots>
+  <screenshots>    
     <screenshot type="default">
-      <image>https://raw.githubusercontent.com/wwmm/easyeffects/master/images/appdata-screenshot-01.png</image>
-      <caption>Set the volume and turn on/off effects</caption>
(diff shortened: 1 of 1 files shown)


wwmm:
These descriptions?
```
<p>EasyEffects is an advanced audio manipulation tools. It includes an equalizer, limiter, compressor and a reverberation tool, just to mention a few. To complement this there is also a built in spectrum analyzer.</p>
    <p>Because EasyEffects uses the default PipeWire sound server it will work with most, if not all, applications you use. All supported applications are presented in the main window, where each can be enabled individually.</p>
    <p>Besides manipulating sound output, EasyEffects is able to apply effects to an input device, such as a microphone. This is, for example, useful in audio recording, but it also works well during voice conversations.</p>
    <p>When EasyEffects is launched it will conveniently remember the configuration used in the last session. It is also possible to save all the current settings as profiles.</p>
```
You are right. There is a type there. But I think that the message the descriptions is trying to pass is still accurate.

As I do not use flatpak I usually do not look at our appstream files with the attention they deserve. If you find anything that needs improvement feel free to suggest changes.

PR #1056, opened 2021-07-30
Implement LSP Multiband Compressor

Digitalone1:
Second attempt to replace Calf MBC with LSP MBC. This time we can do it.

As now, it remains only to link preset keys.


wwmm:
Solid work. I am still quite busy but after a quick look I have only 2 concerns

1. The plugin interface height seems too big. I know that there is a scroll there but we still have to find a place for each band signal meters. Maybe we will still need to use a tab for each band somewhere like in the single band compressor. As I have a big monitor at my desktop I could just maximize the window. But I am concerned about users on laptop screens.
2. Sometimes there is a segmentation fault. Looking at `gdb` output the cause is the line with `int row = listbox->get_selected_row()->get_index();`. I am not sure I will have time today to understand why.

PR #1059, opened 2021-07-31
LSP MBC interface fixed

Digitalone1:
> Solid work. I am still quite busy but after a quick look I have only 2 concerns
> 
>     1. The plugin interface height seems too big. I know that there is a scroll there but we still have to find a place for each band signal meters. Maybe we will still need to use a tab for each band somewhere like in the single band compressor. As I have a big monitor at my desktop I could just maximize the window. But I am concerned about users on laptop screens.

I added an expander. It seems better than the stackpage because when you change band, you still see the main compress options. With stackswitcher, if you switch to sidechain options and change band, then return to the original band, you can't see the compressor options.
 
>     2. Sometimes there is a segmentation fault. Looking at `gdb` output the cause is the line with `int row = listbox->get_selected_row()->get_index();`. I am not sure I will have time today to understand why.

Maybe the signal is emitted when the listbox return a null pointer. I made a safer function for row switching. 

Please test on your system.

Tomorrow I'll try to link the presets and test with audio reproduction. There's still to link the band gain and "band end freq" in the UI.

### DIFF
--- a/data/ui/multiband_compressor.ui
+++ b/data/ui/multiband_compressor.ui
@@ -9,122 +9,120 @@
         <property name="orientation">vertical</property>
 
         <child>
-            <object class="GtkGrid">
+            <object class="GtkBox">
                 <property name="halign">center</property>
                 <property name="valign">center</property>
-                <property name="row-spacing">24</property>
-                <property name="column-spacing">24</property>
+                <property name="spacing">24</property>
 
                 <child>
-                    <object class="GtkGrid">
+                    <object class="GtkBox">
                         <property name="halign">center</property>
-                        <property name="valign">center</property>
-                        <property name="row-spacing">6</property>
-                        <property name="column-spacing">24</property>
+                        <property name="valign">start</property>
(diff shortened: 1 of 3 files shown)


wwmm:
>I added an expander. It seems better than the stackpage because when you change band, you still see the main compress options.

I agree. As I never used this widget before I did not even think about it. It makes sense in this context because we tune the compressor parameters more often than the sidechain.

wwmm:
Totally unrelated comment but I still wish that this lsp plugin did not have the first band that is different from the others. I was staring at the interface and telling to myself "we have to remember to add the control to set the band frequency here". And after some time I remembered that the first band is unique and does not have such control... I still do not get why he had to design this that way...

Digitalone1:
The answer is because it could be used also as a single band compressor, so the first band is always enabled. I prefer that way. It's not really different from before where you had 4 bands with 3 split frequency. Now, since you can choose up to 8 band, if you choose only 1, then don't need the split frequency.

Issue #1031, opened 2021-07-17
LSP Multiband Compressor


Digitalone1:
I started to port the multiband compressor from calf to lsp.

I made a [new branch](https://github.com/Digitalone1/easyeffects/tree/lsp-mbc) on my repository.

Gsettings schema and the bindings should be updated. I didn't compile, but most of the work is already done on the code side. Maybe something could be easily redefined by @wwmm. Can you make a new branch so I can pull a merge request to it and you can work on it?

It seems that lsp mbc does not have output levels for bands, so I don't know where to bind the output connections. I see some curve level meter parameters for each band, maybe its the same as output level for calf.

```
clm-0               : Curve level meter 0
                        Float. Range:      0 - 15,84893 
                        Default:               0
``` 

It could be tested.

On the UI side. I don't even know where to start for GTK4 interface. I was thinking to start modifying the old interface from glade and maybe @wwmm can modify it to make it work on GTK4.

Then I was also thinking to add some methods to mbc preset class to make old calf presets portable to the new lsp one. I have to work on it to test if it's possible.

wwmm:
>I see some curve level meter parameters for each band, maybe its the same as output level for calf.

It is possible. Only testing to know for sure.

>I was thinking to start modifying the old interface from glade and maybe @wwmm can modify it to make it work on GTK4.

It is a bad idea. I tried to reuse glade xml files and I gave up. When people say glade adds tons of unnecessary things to the xml they are not joking. And to make things worse the grid layout works in a different way in gtk4. In the end it will be the same as rewriting the ui file from scratch.

>Then I was also thinking to add some methods to mbc preset class to make old calf presets portable to the new lsp one. I have to work on it to test if it's possible.

I do not think it is worth the effort. We will have more code to maintain and test for little gain. The number of multiband plugins users in EasyEffects is probably very low. And unless the compressors use the same algorithm trying to reuse Calf parameters may not even make sense.

wwmm:
> And to make things worse the grid layout works in a different way in gtk4

Other widgets also have small changes. Like not having some of the old properties glade will add to the file.

Digitalone1:
So I have to try to rework the existing GTK4 interface, But is there at least a way to show the graphical result of the modified xml?

wwmm:
>But is there at least a way to show the graphical result of the modified xml?

[...]

PR #1060, opened 2021-08-01
LSP MBC presets completed

Digitalone1:
Linked all the presets.

I have issues on loading, wrote about it in #1031

The plugin works. It only remains to link frequency_end and reduction plugin ports to their labels inside the UI. I think "reduction" is the gain. "Curve level" is the band level instead? There's also "envelope meter" which I don't have any clue what it is, maybe the band sidechain level? I don't even know if is useful to show it the UI.

Can you suggest me how to link these ports to show them in the UI?

About the tests, nothing wrong except that I keep getting a critical log line during the application shutdown

```
(easyeffects:6103): easyeffects-DEBUG: 11:37:45.364: application_ui: destroyed

(easyeffects:6103): GLib-GIO-CRITICAL **: 11:37:45.389: g_application_release: assertion 'application->priv->use_count > 0' failed
(easyeffects:6103): easyeffects-DEBUG: 11:37:45.390: application:  destroyed
(easyeffects:6103): easyeffects-DEBUG: 11:37:45.390: presets_manager: destroyed
(easyeffects:6103): easyeffects-DEBUG: 11:37:45.391: sie: destroyed
(easyeffects:6103): easyeffects-DEBUG: 11:37:45.391: effects_base: destroyed

```

### DIFF
--- a/data/schemas/com.github.wwmm.easyeffects.multibandcompressor.gschema.xml
+++ b/data/schemas/com.github.wwmm.easyeffects.multibandcompressor.gschema.xml
@@ -30,11 +30,11 @@
     </enum>
     <schema id="com.github.wwmm.easyeffects.multibandcompressor">
         <key name="input-gain" type="d">
-            <range min="-40.0" max="20.0" />
+            <range min="-36.1" max="36.1" />
             <default>0.0</default>
         </key>
         <key name="output-gain" type="d">
-            <range min="-40.0" max="20.0" />
+            <range min="-36.1" max="36.1" />
             <default>0.0</default>
         </key>
 
--- a/data/ui/multiband_compressor_band.ui
+++ b/data/ui/multiband_compressor_band.ui
@@ -305,7 +305,7 @@
                                 <property name="adjustment">
                                     <object class="GtkAdjustment">
                                         <property name="lower">-120</property>
-                                        <property name="lower">-36</property>
+                                        <property name="upper">-36</property>
                                         <property name="value">-120</property>
                                         <property name="step-increment">0.01</property>
                                         <property name="page-increment">0.1</property>
@@ -702,7 +702,7 @@
                                         <property name="update-policy">if-valid</property>
                                         <property name="adjustment">
                                             <object class="GtkAdjustment">
-                                                <property name="upper">-120</property>
+                                                <property name="lower">-120</property>
                                                 <property name="upper">40</property>
                                                 <property name="step-increment">0.1</property>
                                                 <property name="page-increment">1</property>
--- a/include/multiband_compressor_preset.hpp
+++ b/include/multiband_compressor_preset.hpp
@@ -27,6 +27,8 @@ class MultibandCompressorPreset : public PluginPresetBase {
   MultibandCompressorPreset();
 
  private:
+  static constexpr uint n_bands = 8;
+
   void save(nlohmann::json& json, const std::string& section, const Glib::RefPtr<Gio::Settings>& settings) override;
 
   void load(const nlohmann::json& json,
--- a/src/multiband_compressor.cpp
+++ b/src/multiband_compressor.cpp
@@ -45,57 +45,57 @@ MultibandCompressor::MultibandCompressor(const std::string& tag,
   lv2_wrapper->bind_key_enum(settings, "envelope-boost", "envb");
 
   for (uint n = 0; n < n_bands; n++) {
-    if (n > 0) {
-      lv2_wrapper->bind_key_bool(settings, "enable-band" + std::to_string(n), "cbe_" + std::to_string(n));
+    auto nstr = std::to_string(n);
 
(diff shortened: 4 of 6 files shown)


wwmm:
>I think "reduction" is the gain.

It seems so. The plugin ttl file has
```
a lv2:OutputPort, lv2:ControlPort ;
lv2:index 65 ;
lv2:symbol "rlm_0" ;
lv2:name "Reduction level meter 0" ;
units:unit [
                        a units:Unit ;
                        rdfs:label "gain" ;
                        units:symbol "G" ;
                        units:render "%.8f G" ;
 ] ;
lv2:portProperty pp:logarithmic, pp:hasStrictBounds ;
lv2:minimum 0.000000 ;
lv2:maximum 3981.072998 ;
lv2:default 0.000000 ;
```
The label is set to `gain`.

>"Curve level" is the band level instead?

I think this is that dot that moves over the compressor curve graph. For now let's just call it `Curve Level`.

>There's also "envelope meter" which I don't have any clue what it is, maybe the band sidechain level?

>About the tests, nothing wrong except that I keep getting a critical log line during the application shutdown

I do not have this error. I think you may be seeing them because of the gsettings error.

I am not sure about what this is measuring. But based on what is written in this link https://en.wikipedia.org/wiki/Envelope_detector it seems to be the average or peak level of the envelop curve. It is probably better to just call it `Envelope Level`.

>Can you suggest me how to link these ports to show them in the UI?

I think I can do that today. I will take a look at it and update the lsp branch. I will also run clang-format on a few files.

Digitalone1:
Envelope and curve level meter bars should be added to the UI. I put the gain level bar on top of the sidechain expander. But with the addition of the other two bars, I don't know if this is a good idea or it's better to move all of them at the bottom.

After the implementation of the lsp mbc, are you OK on replacing the calf limiter with the lsp one?

wwmm:
>But with the addition of the other two bars, I don't know if this is a good idea or it's better to move all of them at the bottom.

I also think that this may become necessary. Let me handle the connection of these signals first. Now I remember why I did not do that before... But I think I can still do that today.

>After the implementation of the lsp mbc, are you OK on replacing the calf limiter with the lsp one?

I almost did that a long time ago. But at the time it did not seem to have as many advantages as the move to the lsp compressor had. To me it is fine moving to the lsp limiter.

wwmm:
I was able to find a solution for the signals connections that I think it is acceptable. I connected only the frequency range signal and the band gain.I am not sure yet about how we can deal with the insane range that the band gain has. It goes from zero to `3981.072998`. After setting the gtk scale to this max value we can not even see most of the values in the level bar
Screenshot from 2021-08-01 12-41-47
When the compressor is in the downward mode the gain stays between 0 and 1. The other modes can go from 1 to this insane high level. I think that we will have to check in which mode the compressor is whenever a new signal comes and update the max level of the gtk scale accordingly. What may not be that good for performance. But I am not seeing another way...

I have also included the multiband in the latency reporting shown in our window bottom panel. When setting a lookahead value above zero this plugin introduces latency.

wwmm:
>It goes from zero to 3981.072998

Linear scale.

wwmm:
The lsp branch already has these changes in case you want to update yours.

wwmm:
I have added the envelope and the curve level meters
Screenshot from 2021-08-01 13-05-29
Same problem with the scale range. And the plugin interface height is a lot bigger.

The lsp branch has already been updated with these changes

wwmm:
I think there is not need to write `Band Gain` or `Band Envelope`. Just `'Gain` and `Envelope` is enough. Same thing for the curve level meter.

wwmm:
Maybe one solution to the height problem is putting the compressor widgets also inside an expander.

wwmm:
While switching presets I also saw the error 
```
Settings schema 'com.github.wwmm.easyeffects.multibandcompressor' does not contain a key named 'compression-mode'
```
Something is not right in the class `MultibandCompressorPreset`

wwmm:
Problem fixed. Some keys were missing a `+ nstr` in the code.

wwmm:
I think that this branch is good enough to be merged to the master. There is the issue with the level meters scale and the window height. But it is something that can be fixed.

wwmm:
Merge done. I will delete the lsp_mbc branch

Digitalone1:
> I am not sure yet about how we can deal with the insane range that the band gain has. It goes from zero to `3981.072998`. After setting the gtk scale to this max value we can not even see most of the values in the level bar

My suggestion is to just show only the 3 labels on one row and get rid of the level bars.

wwmm:
>My suggestion is to just show only the 3 labels on one row and get rid of the level bars.

This crossed my mind. I am not convinced these level bars are worth the trouble.

Digitalone1:
> This crossed my mind. I am not convinced these level bars are worth the trouble.

As I'm seeing, this levels are also available for the singleband compressor, but we don't show them. We are not forced to show them, but at least for the multiband compressor putting them on one line without levelbars is the best solution.

Issue #1000, opened 2021-07-07
Build error on Ubuntu


ThatcherC:
Just tried building v6.0.0 on Ubuntu and encountered this vague error. Is this a known issue? Happy to try out any proposed fixes and report back. Thanks!

```
$ git checkout v6.0.0 
$ debuild -i -us -uc
 dpkg-buildpackage -us -uc -ui -i
dpkg-buildpackage: info: source package pulseeffects
dpkg-buildpackage: info: source version 4.8.2-1
dpkg-buildpackage: info: source distribution unstable
dpkg-buildpackage: info: source changed by Mikhail Novosyolov <mikhailnov@dumalogiya.ru>
 dpkg-source -i --before-build .
dpkg-buildpackage: info: host architecture amd64
dpkg-source: error: source package has two conflicting values - easyeffects and pulseeffects
dpkg-buildpackage: error: dpkg-source -i --before-build . subprocess returned exit status 25
debuild: fatal error at line 1182:
dpkg-buildpackage -us -uc -ui -i failed
```

**Platform:**
- Ubuntu 20.10
- Linux 5.8.0-59-generic

wwmm:
Our scripts for Debian packages still have to be updated. As I do not use a Debian based distribution I can not fix that.

ThatcherC:
Okay! Thanks for the speedy response. I'll poke around and see if I can get a working build but I'm not an expert on build systems. Out of curiosity, what distributions are you developing on? And are there any older releases that would be better for a recent Debian? It might be nice to have a list of supported platforms/distros in the README. Really hoping I can this up and running!

wwmm:
>what distributions are you developing on?

I run Arch Linux on all my computers.

>And are there any older releases that would be better for a recent Debian?

It depends. If you are using Pulseaudio it would be `4.8.5`. If you are using PipeWire and you can not build `6.0.0` then the next option is `5.0.4`

jinwk00:
> It depends. If you are using Pulseaudio it would be `4.8.5`. If you are using PipeWire and you can not build `6.0.0` then the next option is `5.0.4`

Apparently the script for Debian has not been updated since 4.8.2-1
5.0.4:
` dpkg-buildpackage -us -uc -ui -i
dpkg-buildpackage: info: source package pulseeffects
dpkg-buildpackage: info: source version 4.8.2-1
dpkg-buildpackage: info: source distribution unstable
[...]

PR #1063, opened 2021-08-01
Update description and rename appdata to metainfo

vchernin:
Current AppStream guidelines (not just Flatpak's guidelines) prefer `appname.metainfo.xml` over `appname.appdata.xml`: [Flatpak docs](https://github.com/flathub/flathub/wiki/AppData-Guidelines), [AppStream docs](https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Application.html). Nothing is really broken with `appdata`, it is still permitted for legacy compatibility. However, it seems reasonable to update it as `metainfo` could become a requirement.

I have already tested this succesfully with the [Flatpak build](https://github.com/flathub/flathub/pull/2429#issuecomment-891622857). 

The AppData description was also updated so users see upfront what the difference is between EasyEffects and PulseEffects.

I did notice there are translations and `/util/add-appdata-release.sh` which refer to appdata, so I changed those too as part of the renaming commit.

The Debian package seems like it should also be consistent, but it seems that needs more work than just renaming: https://github.com/wwmm/easyeffects/issues/1000


wwmm:
Ok. Let me know when you think it is ready to be merged.

vchernin:
I wasn't very certain how this would affect the translations in the software center so I tried it out a bit.

I tried the Flatpak build with different languages (English and French), and I didn't spot any description regressions. Also I couldn't find any language regressions in regard to the app itself.

While they weren't regressions, I am unsure how well translated AppStream data is being handled currently. I switched my session over to French (France) and apparently GNOME Software thinks it's not translated? But it looks like it should be as there are French translations for them: https://github.com/wwmm/easyeffects/blob/master/po/fr_FR.po

For all I know these could be mostly due to Flatpak-related issues, as I didn't test much with the Arch package.

[Gnome software showing data tested before this PR](https://user-images.githubusercontent.com/38842733/128120257-9becba96-3d39-4e90-a9fb-59b8938ebb40.png).

(The broken screenshots are an entirely unrelated Flathub issue).

This translation stuff might be best to improve/discuss in a future PR/Issue, but I don't personally feel comfortable merging this until I have some idea of what is going on with AppStream data translations.

Issue #1062, opened 2021-08-01
Link pw filters only once when needed


Digitalone1:
Related to #1018 but here we'll discuss a possible solution.

Since connecting and disconnecting the filters to/from pw causes issues, I think we can link them only once when needed.

The idea is to connect a filter only when the relative plugin is added to the list. Or, when at the start it's already added to the list, link the filter before the plugin gets audio in input the first time. Then do not disconnect it, take it linked to pw graph.

This has different advantages. If I use few plugins, I only get them linked to pw, all the others are disconnected without flooding the pw graph with unneeded nodes. If I try all the plugins adding them to the list and then I delete them all, I got all of them linked, but at the next start I still have only the few needed.

I already tested this adding a bool flag to plugin_base and delaying the function to link the filter to pw. The application was starting without errors, except that the UI was now showed. I thought it was something related to UI classes that needed the filter to be linked, but that can be changed.

Unfortunately I deleted that test branch, but I can make another. This has to be discussed.

wwmm:
>The idea is to connect a filter only when the relative plugin is added to the list

I do not understand. That is already what happens. The plugins are always connected to PipeWire but they are linked to something else only when the user adds them to the list.

wwmm:
Nevermind. I read coonect but in mind mind it was `link`

wwmm:
It is an idea to consider but I think that at the end of the day the difficulties will be similar. And people will still complain that after adding a filter they remain in PipeWire graph after removing. IT will probably be better to try to fix this once and for all. The problem is time.

Digitalone1:
Well, you can still tell them, "close and reopen EE", only the needed nodes are connected.

Tomorrow I will make a test branch. At least I'd like to know why the application UI is not starting even if no errors are shown in the log.

Digitalone1:
I was talking something like [that](https://github.com/Digitalone1/easyeffects/commit/57b26eba7e3c445cec0bada03934c646810b2495). New method [connect_pw_filter](https://github.com/Digitalone1/easyeffects/blob/57b26eba7e3c445cec0bada03934c646810b2495/src/plugin_base.cpp#L175).

No compilation errors, no critical debug lines in the log, but the application UI won't show. Why?

<details>
<summary> Log </summary>

```
(process:13141): easyeffects-DEBUG: 15:56:26.210: main: locale directory: /usr/share/locale
[...]

PR #1069, opened 2021-08-03
Try to connect pw filters once when needed

Digitalone1:
Related to #1062

### DIFF
--- a/include/application.hpp
+++ b/include/application.hpp
@@ -1,5 +1,5 @@
 /*
- *  Copyright © 2017-2020 Wellington Wallace
+ *  Copyright © 2017-2022 Wellington Wallace
  *
  *  This file is part of EasyEffects.
  *
--- a/include/application_ui.hpp
+++ b/include/application_ui.hpp
@@ -1,5 +1,5 @@
 /*
- *  Copyright © 2017-2020 Wellington Wallace
+ *  Copyright © 2017-2022 Wellington Wallace
  *
  *  This file is part of EasyEffects.
  *
--- a/include/autogain.hpp
+++ b/include/autogain.hpp
@@ -1,5 +1,5 @@
 /*
- *  Copyright © 2017-2020 Wellington Wallace
+ *  Copyright © 2017-2022 Wellington Wallace
  *
  *  This file is part of EasyEffects.
  *
--- a/include/autogain_preset.hpp
+++ b/include/autogain_preset.hpp
@@ -1,5 +1,5 @@
 /*
- *  Copyright © 2017-2020 Wellington Wallace
+ *  Copyright © 2017-2022 Wellington Wallace
  *
  *  This file is part of EasyEffects.
  *
--- a/include/autogain_ui.hpp
+++ b/include/autogain_ui.hpp
@@ -1,5 +1,5 @@
 /*
- *  Copyright © 2017-2020 Wellington Wallace
+ *  Copyright © 2017-2022 Wellington Wallace
  *
  *  This file is part of EasyEffects.
  *
--- a/include/bass_enhancer.hpp
+++ b/include/bass_enhancer.hpp
@@ -1,5 +1,5 @@
 /*
- *  Copyright © 2017-2020 Wellington Wallace
+ *  Copyright © 2017-2022 Wellington Wallace
  *
  *  This file is part of EasyEffects.
  *
--- a/include/bass_enhancer_preset.hpp
+++ b/include/bass_enhancer_preset.hpp
@@ -1,5 +1,5 @@
 /*
- *  Copyright © 2017-2020 Wellington Wallace
+ *  Copyright © 2017-2022 Wellington Wallace
  *
  *  This file is part of EasyEffects
  *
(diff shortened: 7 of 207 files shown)


wwmm:
Handling the call to `connect_to_pw` inside `EffectsBaseUi` is not a good idea. That class is instantiated only when the window is opened. If we start in service mode `connect_to_pw` will never be called.

Digitalone1:
> Handling the call to `connect_to_pw` inside `EffectsBaseUi` is not a good idea. That class is instantiated only when the window is opened. If we start in service mode `connect_to_pw` will never be called.

I tried to do it in process method, but on_process function is not executed when the filter is disconnected, so it will never be connected. 

Where do you suggest to do it?

wwmm:
>I tried to do it in process method, but on_process function is not executed when the filter is disconnected, so it will never be connected.

Yes. The on_process methos is called only when the plugin is linked and processing audio buffers. So even if it were connected it would not be enough.

There are two two things to keep in mind when starting in service mode. We have to read the plugins order that is already saved and link the filters accordingly. So when initializing we have to read the plugins array and connect the ones we will need before trying to link them.

The second thing is that the plugins array may change. So just like it is already done somewhere else  a signal connection like `settings->signal_changed("plugins").connect()` will have to be made. Inside the connect callback we will have to check again if the filters we need are connected and if not we have to connect them before trying to do the link.

wwmm:
Or maybe to check for the connection before trying to do the link is enough

 https://github.com/wwmm/easyeffects/blob/13b0ab00657f2e369920a4bdbe6b71ce17aa74ba/src/stream_output_effects.cpp#L156

https://github.com/wwmm/easyeffects/blob/13b0ab00657f2e369920a4bdbe6b71ce17aa74ba/src/stream_input_effects.cpp#L160

Digitalone1:
I'll try to do other attempts when have time.

Digitalone1:
I did as you said and it's working as expected!

The only issue remaining is the application not exiting well, I have to kill it from task manager.

However, after dragging the plugin in the new position, I get the following warning:

```
(easyeffects:2763): Gtk-WARNING **: 10:15:45.837: Allocating size to GtkDragIcon 0x5640cc5e4160 without calling gtk_widget_measure(). How does the code know the size to allocate?

```

Digitalone1:
As I'm seeing from the logs, the PipeManager destructor is not executed. That's tricky to resolve for me.

Digitalone1:
@wwmm can you check why the PipeManager is not destructed? Thanks.

Digitalone1:
Found the reason. Spectrum and Output Level classes are not destroyed. But why?

wwmm:
Today is going to be a busy day so I do not know if I will be able to do a proper investigation. But after pulling your latest branch to my computer I see a potential problem
```
    auto list_size = list.size();
    auto links = pm->link_nodes(pm->pe_sink_node.id, plugins[list[0]]->get_node_id());

    for (const auto& link : links) {
      list_proxies.emplace_back(link);
    }

    for (size_t n = 0; n < list_size; n++) {
      if (!plugins[list[n]]->connected_to_pw) {
        plugins[list[n]]->connect_to_pw();
      }

      if (n > 0) {
        auto links = pm->link_nodes(plugins[list[n - 1]]->get_node_id(), plugins[list[n]]->get_node_id());

        for (const auto& link : links) {
          list_proxies.emplace_back(link);
        }
      }
    }
```
As this code is now it attempts to link `plugins[list[0]]` before checking if it is connected to PipeWire. Something like
```
      if (!plugins[list[0]]->connected_to_pw) {
        plugins[list[0]]->connect_to_pw();
      }
```
has to be done before
```
auto links = pm->link_nodes(pm->pe_sink_node.id, plugins[list[0]]->get_node_id());
```

Digitalone1:
Counter is starting from 0, so the first plugin on 0 index is checked.

wwmm:
>Counter is starting from 0, so the first plugin on 0 index is checked.

It is not going to help because a link to the first filter is done outside the for loop

Digitalone1:
> > Counter is starting from 0, so the first plugin on 0 index is checked.
> 
> It is not going to help because a link to the first filter is done outside the for loop

Right. I will give a check.

wwmm:
In any case the problem related to the app closing is being caused by something else.

wwmm:
Looking atthe spectrum destructor I found a problem
```
  pw_thread_loop_lock(pm->thread_loop);

  if (connected_to_pw) {
    disconnect_from_pw();
  }

  pw_core_sync(pm->core, PW_ID_CORE, 0);

  pw_thread_loop_wait(pm->thread_loop);

  pw_thread_loop_unlock(pm->thread_loop);
```
you are locking the PiprWire main loop `pw_thread_loop_lock(pm->thread_loop);` before calling `disconnect_from_pw`. But the first thing that disconnect_from_pw does is trying to lock the main loop. But it can't do that because the first lock was not released.

wwmm:
I think you can do the same thing you did to the other filters. Keep
```
if (connected_to_pw) {
    disconnect_from_pw();
  }
```
and remove the calls to the pipewire functions that are already inside `disconnect_from_pw`.

Digitalone1:
Yes, that's the reason. Spectrum is the only class I did that, my mistake with copy-paste.

Digitalone1:
Should be OK now. No issues to me. Please, test on your system @wwmm when you have time.

wwmm:
I tested both pipelines and it seems fine :-)

PR #1073, opened 2021-08-06
Implemented LSP Limiter

### LINKED ISSUE #1054: LSP Limiter
Same as #1031 - would be nice to use LSP limiter with ALR and avoid https://github.com/calf-studio-gear/calf/issues/217

The only deficiency in LSP limiter now is that it's not multiband.
Hopefully, this will be alleviated with https://github.com/sadko4u/lsp-plugins/issues/140
And then also https://github.com/sadko4u/lsp-plugins/issues/196

Digitalone1:
Fixes #1054 

@wmm I also added an hidden device dropdown widget to MCB and Limiter if in the future you are adding the external sidechain feature.

Do not release soon, let's test for some days waiting for bugs and translations.


Digitalone1:
@wwmm I'm testing the plugin and the application. Sometimes I see the last added plugin mutes the reproduction. In Helvum I see the latest filter added don't get connected to the output_level node. This occurs even if the filter gets successfully connected and no warnings are shown by the link_nodes method in pipe manager. The issue is resolved changing the plugin order.

Anyway, even if no errors are shown, sometimes the nodes don't get connected. Once happened that testing with lots of plugins the second to last was connected to the last plugin only with 1 channel and the last plugin was not connected to output level node. This seems similar to what is happening on KDE linking only one channel. 

It's weird that nothing wrong gets reported by EE. If a links is not made, a warning should be reported by the application. Seems pw is reporting the link is successfully established even if it's not really made.

For now I tried to write a safer way to connect the filter to pw. Maybe the same should be done with node linking adding a try/catch statement in link_nodes function.

wwmm:
I think that the try/catch around PipeWire functions won't have the effect you want because PipeWire is written in C and as a result does not throw exceptions. It is not going to cause any harm to have the try catch. But it is probably useless.

wwmm:
>It's weird that nothing wrong gets reported by EE

Yes. I will try to do some minor optimizations to how we do links. I think we can lock PipeWire's loop for a little less time. Maybe it helps somehow. I will also try to see if PipeWire offers a better way to check if the link has failed.

wwmm:
>If a links is not made, a warning should be reported by the application

Yes. I think that the link is being destroyed right after being created. It is the only explanation for the lack of warnings when a link is clearly missing.

wwmm:
I have updated the master branch with latency reporting for the new limiter plugin. The lookahead feature introduces latency. Same thing probably happened to the Calf limiter too. But that plugin did not report its latency.

I also removed the try/catch I talked about before. Unfortunately it is not going to help. We are also locking the PipeWire loop for a smaller amount of time now. Let's see if this makes any difference to the link situation. But I am not confident this is the cause of the problem.

Digitalone1:
Okay. I suggest to test for 3/4 days, then you can release. Meanwhile I'm going to update the translation.

PR #1078, opened 2021-08-08
Optimized node linking

Digitalone1:
This should be slightly better. Not only it skips the unconnected filter, but also the node with less than two channels linked.

Then saving and using the prev/next node id is better and more intuitive.

It's working on my side, please test on yours.


wwmm:
It is good. I will add warnings for when `link_size < 2`. As it is right now we are just silently skipping plugins. And if the link to the sound card or microphone device fails the user will still have no sound.

The skipped filters with partial links may still become active. But I think that the real solution to that is not having failed links in the first place.

PR #1079, opened 2021-08-09
Updated Italian translation

Digitalone1:
Also small fixes to ui.

I think it could be ready for a new release. Since the big changes and two plugins switched to LSP, I suggest **6.1.0**.


Digitalone1:
I forgot to mention the changes to Crystalizer in the changelog.

PR #1081, opened 2021-08-10
Update AppData and translations (improved #1063)

vchernin:
As I worked on #1063 I realized it was probably best to redo my changes. I had made some redundant fixes that could have been done better.

This is the same thing but hopefully a lot clearer.

I still want to test this and see if it regresses anything, but I think I understand how translations work now so hopefully it works.

Original description:

Current AppStream guidelines (not just Flatpak's guidelines) prefer `appname.metainfo.xml` over `appname.appdata.xml`: [Flatpak docs](https://github.com/flathub/flathub/wiki/AppData-Guidelines), [AppStream docs](https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Application.html). Nothing is really broken with `appdata`, it is still permitted for legacy compatibility. However, it seems reasonable to update it as `metainfo` could become a requirement.

I have already tested this succesfully with the [Flatpak build](https://github.com/flathub/flathub/pull/2429#issuecomment-891622857). 

The AppData description was also updated so users see upfront what the difference is between EasyEffects and PulseEffects.

I did notice there are translations and `/util/add-appdata-release.sh` which refer to appdata, so I changed those too as part of the renaming commit.

The Debian package seems like it should also be consistent, but it seems that needs more work than just renaming: https://github.com/wwmm/easyeffects/issues/1000


vchernin:
So I'm still not exactly sure why the translations for PulseEffects Flatpak work, but EasyEffects Flatpak doesn't work regardless of this patch. I am specifically referring to the app description that shows up in GNOME Software or KDE Discover. This PR updates the English AppData description as expected for other languages (at least French) it doesn't. I am really not sure what is incorrect here.

Regardless, does it make sense to update all the language translation files to match the latest template (in https://github.com/wwmm/easyeffects/pull/1081/commits/d879e62362ea0cbb18ab08c7480a18b8f8c18ef9)?

wwmm:
Flatpak appdata file is already in our POTFILES https://github.com/wwmm/easyeffects/blob/e06c379516b0268d256d8f0ec8f823ef009320cb/po/POTFILES.in#L1. So the same steps done to update translations for our window should work for the appdata xml too.

Digitalone1:
> I haven't updated the pt_BR translations in ages. But if I am not mistaken the manual steps at the end of the page https://github.com/wwmm/easyeffects/wiki/Translating-EasyEffects should still work. The first step after changing the original text in English is running `ninja easyeffects-pot` to update the translation template. And then `ninja easyeffects-update-po` to update all the po files with the new English strings. After that the translator updates the desired po file with the new translations. Am I missing something @Digitalone1 ?

Those instructions work. I'm still following them.

vchernin:
What might have been the issue is I tried to update the translation template file by hand. Then I use poedit to update each language file.

I'll try the following steps then:
1. Update English text
2. Update template file with ninja (following manual instructions)
3. Update each language (following manual instructions).

I tried to follow these steps earlier but I ran into dependency issues building on Fedora, I'll play around with it a bit.

See #1085

PR #1085, opened 2021-08-10
Update AppData and Translations

vchernin:
Ok, I think this one will work. I am testing it in a bit. Sorry about all these PRs, I really did not expect this to be so difficult for me.

I did the following:
1. Update English text
2. Update template file with ninja (following manual instructions)
3. Update each language (following manual instructions).

I dropped the appdata>metainfo change for now as I really just want to get this working properly. If this works I'll update it in a future PR.

See #1081


vchernin:
So I think I was doing it right the whole time. The problem is apparently just I was expecting an incomplete translation to work.

For me here in French the AppData description is still in English, but in German it is actaully translated. So this change is doing what it is supposed to. 

It updates the English version and correctly updates the translation files to reflect it.

Since I now finally seem to understand that this makes sense, I think this is ready to merge.

I'll try to follow up with changes that improve the French translation, and also rename appdata>metainfo. But I don't think those things need to block that change.

PR #1091, opened 2021-08-11
Fixed APO preset loading, updated translation and changelog

### LINKED ISSUE #1094: Can't load APO presets properly when locale is different than C
Since `stringstream` rely on system locale, double values with digits can't be properly imported on certain locales because the stream expects comma rather than dot. A locale independent method should be used instead.

I'm trying to fix it.

Digitalone1:
As the title says.

Fixes #1094

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
   Calf Studio.
 - LV2 filters now are spawned in PipeWire graph only when loaded the first time. Once loaded, they remain connected
   until EasyEffects shutdown.
+- Equalizer APO presets loading is now working properly on locales different than C.
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


Digitalone1:
Fixes #1094

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
   Calf Studio.
 - LV2 filters now are spawned in PipeWire graph only when loaded the first time. Once loaded, they remain connected
   until EasyEffects shutdown.
+- Equalizer APO presets loading is now working properly on locales different than C.
```
Digitalone1, line 16: sorry @wwmm this line should be in fixed, not added.

wwmm: Ok.

PR #1100, opened 2021-08-14
Rename appdata to metainfo

vchernin:
Built fine for me in the Flatpak, AppData in gnome software seems to be same as before.

Last part of what was originally #1063 and #1081.

Would you prefer if I rebased the last 2 commits into the first as they don't seem worth keeping as seperate commits? (they don't change much since I ended up updating everything in the first commit by hand anyway)

Edit: might have discovered a regression, marking as draft for now just to be sure.
Edit 2: nevermind, what I noticed wasn't related or introduced by this, I think it's safe to merge.


wwmm:
>Would you prefer if I rebased the last 2 commits into the first as they don't seem worth keeping as seperate commits? 

Not needed in this case. And I have to admit I do not have the patience to rebase many of my commits even when I should.

PR #1101, opened 2021-08-15
New default values for split frequencies in LSP MBC

Digitalone1:
I noticed that band frequency end in mbc ui is not always updating on split frequency changes. This is because the emit method of frequency_range signal is inside process fuction, so it's working only when an audio stream is playing.

I would be better to make it work always, not only during playback, since it could be confusing to the user seeing a wrong frequency on frequency split update.

Unfortunately I'm not so good with signals, so I wasn't able to make it update always. Maybe I'm missing something and you @wwmm can suggest me how to do it correctly.

Anyway, if it can't be done with signals, another solution could be calculating manually the frequencies on split changes. I made some tests and it should be done on settings changes because on spinbuttons changes the signal is not emitted when the value changes through preset application.

It's also quite tricky to to it this way since the bands can also be non-contiguous, so for every enabled band we have to scan all other enabled bands to pick the minimum split frequency major than the current split frequency.

It's not impossible, but if it could be done listening always to plugin ports changes it would be better and easier.

Meanwhile I changed default values for bands since the plugin has 0, 2, 4 and 6 enabled by default while we have the first four. To me seems better to have the first four with other values.

### DIFF
--- a/data/schemas/com.github.wwmm.easyeffects.multibandcompressor.gschema.xml
+++ b/data/schemas/com.github.wwmm.easyeffects.multibandcompressor.gschema.xml
@@ -69,31 +69,31 @@
 
         <key name="split-frequency1" type="d">
             <range min="10.0" max="20000.0" />
-            <default>40.0</default>
+            <default>500.0</default>
         </key>
         <key name="split-frequency2" type="d">
             <range min="10.0" max="20000.0" />
-            <default>100.0</default>
+            <default>1000.0</default>
         </key>
         <key name="split-frequency3" type="d">
             <range min="10.0" max="20000.0" />
-            <default>252.0</default>
+            <default>2000.0</default>
         </key>
         <key name="split-frequency4" type="d">
             <range min="10.0" max="20000.0" />
-            <default>632.0</default>
+            <default>4000.0</default>
         </key>
         <key name="split-frequency5" type="d">
             <range min="10.0" max="20000.0" />
-            <default>1587.0</default>
+            <default>8000.0</default>
         </key>
         <key name="split-frequency6" type="d">
             <range min="10.0" max="20000.0" />
-            <default>3984.0</default>
+            <default>12000.0</default>
         </key>
         <key name="split-frequency7" type="d">
             <range min="10.0" max="20000.0" />
-            <default>10000.0</default>
+            <default>16000.0</default>
         </key>
 
         <key name="sidechain-mode0" enum="com.github.wwmm.easyeffects.multibandcompressor.sidechainmode.enum">
--- a/src/multiband_compressor_ui.cpp
+++ b/src/multiband_compressor_ui.cpp
@@ -232,7 +232,7 @@ MultibandCompressorUi::MultibandCompressorUi(BaseObjectType* cobject,
     auto* selected_row = listbox->get_selected_row();
 
     if (selected_row != nullptr) {
-      int row = selected_row->get_index();
+      const auto row = selected_row->get_index();
 
       if (row > -1) {
         stack->set_visible_child("band" + std::to_string(row));


wwmm:
The first thing that comes to my mind is going to this for loop https://github.com/wwmm/easyeffects/blob/847de956d8c6d96d9c8c4a325f73a1af4e8f049c/src/multiband_compressor.cpp#L47 and adding something like
```
settings->signal_changed("split-frequency" + nstr).connect([=, this](auto key) {
   frequency_range_end_array.at(n) = lv2_wrapper->get_control_port_value("fre_" + nstr);

   Glib::signal_idle().connect_once([=, this] {
       frequency_range.emit(frequency_range_end_array);
   }
});
```

Although a little ineficient because the emitted array will force an update on all the frequency labels it won't be the end of the world. Manual updates to these keys do not happen as often as the ones done inside the `process` function. And it seems the easiest thing to do.

>another solution could be calculating manually the frequencies on split changes.

Besides the problems you have already seen this approach has another one. If LSP changes how this is calculated for some reason our code will need update too. It is better to avoid this.

wwmm:
My idea assumes that `lv2_wrapper->get_control_port_value` will have updated values even when the LSP plugin process function was not called. If the plugin is not doing that then the only solution would be to try to do the calculation ourselves in the ui class. But I really hope this is not going to be necessary...

wwmm:
There may be a catch in my approach. It may require that the plugin has at least been initialized. What is only going to happen once its `setup` function is called. But for that to happen the user will have to play something at east once.

The reason is that we have to know the audio buffer size and the sampling rate to initialize the plugin. Almost all plugins requires this. Not only LV2.

wwmm:
What does the native plugin interface do? Does it update the value even when the plugin was not initialized?

Digitalone1:
> What does the native plugin interface do? Does it update the value even when the plugin was not initialized?

Not when I start it from application list, but yes in Carla even if nothing is playing.

wwmm:
>Not when I start it from application list, but yes in Carla even if nothing is playing.

The situation may be easier for Carla... It probably has a fixed sampling rate and audio buffer size. So it can initialize the plugin even when not playing. But this is something we do not have. PipeWire latency changes dynamically and support for dynamic changes to sampling rate is being work on.

wwmm:
Maybe in the multiband compressor code we could at the and of its constructor set `n_samples = 512`, `rate = 48000` and call `setup()` to make sure that the plugin is initialized to something even when nothing has been played. I think that these frequencies calculation are not dependent on the buffer size and sampling rate value. So it should be fine to do that in the multiband compressor without the risk of misleading the user with meaningless values.

But we should do this only in this plugin. Most of the other plugins have additional structures that have to be initialized with the right buffer size and sampling rate. Otherwise the displayed values won't have much meaning. So it is not a good idea to blindly replicate this "default initialization solution" to other plugins.

Digitalone1:
The first approach you mentioned it's not working. The frequency is updated only on settings changes (that could not be done at startup) and when the signal is emitted all values are 1000. I showed them on terminal with a debug string:

```
(easyeffects:14514): easyeffects-DEBUG: 17:55:51.454: setting frequency end 1000.000000
(easyeffects:14514): easyeffects-DEBUG: 17:55:51.455: setting frequency end 1000.000000
(easyeffects:14514): easyeffects-DEBUG: 17:55:51.455: setting frequency end 1000.000000
(easyeffects:14514): easyeffects-DEBUG: 17:55:51.455: setting frequency end 1000.000000
(easyeffects:14514): easyeffects-DEBUG: 17:55:51.455: setting frequency end 1000.000000
(easyeffects:14514): easyeffects-DEBUG: 17:55:51.455: setting frequency end 1000.000000
(easyeffects:14514): easyeffects-DEBUG: 17:55:51.455: setting frequency end 1000.000000
(easyeffects:14514): easyeffects-DEBUG: 17:55:51.455: setting frequency end 1000.000000

```

When a band is updated during playback, this is what I get:

```
(easyeffects:14514): easyeffects-DEBUG: 17:55:51.454: setting frequency end 1000.000000
(easyeffects:14514): easyeffects-DEBUG: 17:55:51.455: setting frequency end 1000.000000
(easyeffects:14514): easyeffects-DEBUG: 17:55:51.455: setting frequency end 1000.000000
(easyeffects:14514): easyeffects-DEBUG: 17:55:51.455: setting frequency end 1000.000000
(easyeffects:14514): easyeffects-DEBUG: 17:55:51.455: setting frequency end 1000.000000
(easyeffects:14514): easyeffects-DEBUG: 17:55:51.455: setting frequency end 1000.000000
(easyeffects:14514): easyeffects-DEBUG: 17:55:51.455: setting frequency end 1000.000000
(easyeffects:14514): easyeffects-DEBUG: 17:55:51.455: setting frequency end 1000.000000

```

So the values are only initialized when the plugin starts to process.

It's worth a try to call setup in the constructor as you said.

Digitalone1:
No luck. Added the following at the end of multiband compressor constructor:

```
  n_samples = 512;

  rate = 48000;

  setup();
```

Frequencies are updating only on playback startup.

wwmm:
>The frequency is updated only on settings changes

`settings->signal_changed` will only be triggered when the user changes the values in the interface. I thought that was all you wanted. In order to update the values after the initialization the solution will have to be more elaborated...

>Frequencies are updating only on playback startup.

It seems that the plugin has to receive at least one buffer to actually calculate these values... In that case things get harder... Even if we pass an instance of the lower level multiband compressor class to the ui class it will be useless if the plugin only gives the right values after it started processing buffers.

wwmm:
>It seems that the plugin has to receive at least one buffer to actually calculate these values

Another possibility is that this isn't actually necessary. It may be working on Carla because the native plugin interface is getting the right value through its communication port to the plugin. Just like we are using ports to set and read values the LV2 standard has ports that are only used for the communication between the plugin and its interface. This is how the interface gets the info for the graphs it shows. Maybe that is how this is working on Carla. But that is a path we can not use...

Digitalone1:
I pushed the commit to let you see the changes. In this state all frequencies are 1000 Hz at startup and they are correctly updated only when the split frequency is changed during playback. No luck otherwise.

The only solution to me seems stopping listening this ports and calculating values manually on spinbuttons change and enable checkbuttons toggle.

Digitalone1:
> `settings->signal_changed` will only be triggered when the user changes the values in the interface. I thought that was all you wanted. In order to update the values after the initialization the solution will have to be more elaborated...

I'm pretty sure in this state someone after the release will open an issue "band frequency end in mbc is wrong at startup". A workaround should be found.

wwmm:
>The only solution to me seems stopping listening these ports and calculating values manually on frequency split spinbuttons change and band checkbuttons toggle.

If the value is always right it can be done this way. The only problem is having one more thing to pay attention to when a new LSP update comes. If its author decides to change how they are calculated we will be showing wrong values if our code is not synchronized to his.

Personally I do not mind having to play something to see the proper values. I only change parameters when playing something so I cal listen to the result. But I am not against trying to make the proper values to be shown when the plugin was not initialized yet. As long as we do not over complicate things there is not problem.

Digitalone1:
I'm afraid the value won't be always right, especially when it's specified a frequency equal to another band. In this case it's difficult to pick which band will get the frequency end relative to the next split frequency band.

Considering that the behaviour could also change in the future, maybe the better thing to do is leaving the plugin as it is now. If someone will come here for the frequency label not updating, we can only suggest him/her to play something or activate a test signal.

So I will make a force push dropping the last commit and this pull request is only for updating new default values for mbc, if it's ok to you @wwmm.

wwmm:
>maybe the better thing to do is leaving the plugin as it is now.

I think it is fine leaving it as it is for now. I also would like to have this working properly in the initialization eventually. But at the same time it is not the end of the world to ask the user to play something first. In a way it is a similar situation to the graphs we already do not have in the equalizer and the compressors. Not using the native plugin interface has limitations.

PR #1102, opened 2021-08-16
New default hcf/lcf values for MBC

Digitalone1:
According to new split frequencies, also low-cut/high-cut frequencies needs to be updated.

### DIFF
--- a/data/schemas/com.github.wwmm.easyeffects.multibandcompressor.gschema.xml
+++ b/data/schemas/com.github.wwmm.easyeffects.multibandcompressor.gschema.xml
@@ -301,60 +301,60 @@
         </key>
         <key name="sidechain-lowcut-frequency1" type="d">
             <range min="10.0" max="20000.0" />
-            <default>40.0</default>
+            <default>500.0</default>
         </key>
         <key name="sidechain-lowcut-frequency2" type="d">
             <range min="10.0" max="20000.0" />
-            <default>100.0</default>
+            <default>1000.0</default>
         </key>
         <key name="sidechain-lowcut-frequency3" type="d">
             <range min="10.0" max="20000.0" />
-            <default>252.0</default>
+            <default>2000.0</default>
         </key>
         <key name="sidechain-lowcut-frequency4" type="d">
             <range min="10.0" max="20000.0" />
-            <default>632.0</default>
+            <default>4000.0</default>
         </key>
         <key name="sidechain-lowcut-frequency5" type="d">
             <range min="10.0" max="20000.0" />
-            <default>1587.0</default>
+            <default>8000.0</default>
         </key>
         <key name="sidechain-lowcut-frequency6" type="d">
             <range min="10.0" max="20000.0" />
-            <default>3984.0</default>
+            <default>12000.0</default>
         </key>
         <key name="sidechain-lowcut-frequency7" type="d">
             <range min="10.0" max="20000.0" />
-            <default>10000.0</default>
+            <default>16000.0</default>
         </key>
 
         <key name="sidechain-highcut-frequency0" type="d">
             <range min="10.0" max="20000.0" />
-            <default>40.0</default>
+            <default>500.0</default>
         </key>
         <key name="sidechain-highcut-frequency1" type="d">
             <range min="10.0" max="20000.0" />
-            <default>100.0</default>
+            <default>1000.0</default>
         </key>
         <key name="sidechain-highcut-frequency2" type="d">
             <range min="10.0" max="20000.0" />
-            <default>252.0</default>
+            <default>2000.0</default>
         </key>
         <key name="sidechain-highcut-frequency3" type="d">
             <range min="10.0" max="20000.0" />
-            <default>632.0</default>
+            <default>4000.0</default>
         </key>
         <key name="sidechain-highcut-frequency4" type="d">
             <range min="10.0" max="20000.0" />
-            <default>1587.0</default>
+            <default>8000.0</default>
         </key>
         <key name="sidechain-highcut-frequency5" type="d">
             <range min="10.0" max="20000.0" />
-            <default>3984.0</default>
+            <default>12000.0</default>
         </key>
         <key name="sidechain-highcut-frequency6" type="d">
             <range min="10.0" max="20000.0" />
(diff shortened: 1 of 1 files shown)


wwmm:
I think I will make a new release. The approach for channel remapping proposed by PipeWire's developer will be more complicated to do than I thought. And I am not sure I like the idea of having the loopback device stream visible to the user. This kind of thing generated confusion in the past when we had visible GStreamer streams. So I do not think I will hold the next release because of this.

Is there anything more you want to do before the release?

Digitalone1:
No, I don't have nothing to do, or at least I think I didn't forget nothing.

For the channel remapping matter, maybe you can introduce an option so in the default disabled state the things are like now and most of the users won't touch it. Then when enabled the loopback stream shows up to make the remapping if the user need it.

Digitalone1:
Forget to mention it, next release **6.1.0**, right?

wwmm:
>No, I don't have nothing to do, or at least I think I didn't forget nothing.

Ok.

>For the channel remapping matter, maybe you can introduce an option so in the default disabled state the things are like now and most of the users won't touch it. Then when enabled the loopback stream shows up to make the remapping if the user need it.

I have been thinking about doing that. But the re is still the fact that the API to actually implement the loopback is not obvious. And asking the users to create themselves a loobpack device and setting it as default does not seem nice.

>Forget to mention it, next release 6.1.0, right?

Yes.

Maybe before doing the new release we should change the name of the option `Run in Background` in our settings menu. The current name is misleading but I did not have a good idea to replace it yet. It does not make us run in background. If we are running as a service and that option is disabled closing the window kills the process. Some people asked for this in the past. I would like to have a short text describing this feature. `Run in Background` is giving people the wrong idea.

Digitalone1:
Maybe something like "shutdown on window close" and the active poverty is inverted.

If you change it, wait me for the translation.

Digitalone1:
> I have been thinking about doing that. But the re is still the fact that the API to actually implement the loopback is not obvious. And asking the users to create themselves a loobpack device and setting it as default does not seem nice.

Well, if this task is not easily provided by Pipewire now, maybe you can wait for it to be mature in the future. If I don't remembered wrong, Pulseaudio was doing it automatically, right? 

Or maybe you can add a method to discover additional ports on the output device and link them properly. It should be done only on the output device.

wwmm:
>Maybe something like "shutdown on window close" and the active poverty is inverted.

That is the expression I was trying to remember but I could not XD. I will change it.

>If you change it, wait me for the translation.

Ok.

>Well, if this task is not easily provided by Pipewire now, maybe you can wait for it to be mature in the future.

It does the same Pulseaudio did to streams. THat is why it worked with GStreamer even on PipeWire. The problem is that filters (what Pulseaudio does not have) are not handled yet by PipeWire's session manager. So we have to do the link ourselves.

>Or maybe you can add a method to discover additional ports on the output device and link them properly. It should be done only on the output device.

If it were just the link it would be easy. The problem is handling the upmixing and downmixing ourselves. From stereo to four channels is one equation. From stereo do 7.1 surround another and so on. The server is in a much better position to do that right.

wwmm:
>If you change it, wait me for the translation.

Master branch updated.

Digitalone1:
> If it were just the link it would be easy. The problem is handling the upmixing and downmixing ourselves. From stereo to four channels is one equation. From stereo to 7.1 surround another equation and so on. The server is in a much better position to do that right.

So wait lv2 filters to be handled by Pipewire manager.

Digitalone1:
> Master branch updated.

I'll do it tomorrow.

wwmm:
>So wait lv2 filters to be handled by Pipewire manager.

No need for LV2. If PipeWire native filters links were handled by the session manager things would be easier.

Digitalone1:
> > So wait lv2 filters to be handled by Pipewire manager.
> 
> No need for LV2. If PipeWire native filters links were handled by the session manager things would be easier.

I don't know if I understood correctly.

You're saying that links from app to `easyeffects sink` (which can be selected in pavucontrol) or from easyeffects sink to a device can be handled by Pipewire session manager, but links from filters like `output_level` can't, right? And session manager can do channel remapping automatically, right? 

If I'm right, a solution could be adding an `easyeffects output sink`, so output_level redirects stream to it and it will remap channels to speakers when needed by Pipewire session manager.

wwmm:
>You're saying that links from app to easyeffects sink (which can be selected in pavucontrol) or from easyeffects sink to a device can be handled by Pipewire session manager, but links from filters like output_level can't, right?

Yes. Although in the case where a null sink is linked to a device it is not clear if the upmixing/downmixing is done. At least when when the session manager links streams to them upmixing/downmixing is done. I think it probably works when a null sink is used too.

>And session manager can do channel remapping automatically, right?

I am not sure exactly about who is doing the upmixing/downmixing inside Pipewire. But if the session manager is the one doing the link it happens somehow.

>If I'm right, a solution could be adding an easyeffects output sink, so output_level redirects stream to it and it will remap channels to speakers when needed by Pipewire session manager.

It would probably work. But I do not like the idea of having one more sink visible to the user that he/she is not supposed to be touching. Since I started this project I have lost count of how many times users asked which sink they had to use. People really get confused with this. One of the good things of our current approach is showing only devices that the user has to interact with.

That being said if I find the way to create a null sink that is invisible in pavucontrol, etc it would be nice.

Digitalone1:
Well, I can say that Wim Taymans suggested you with a non optimal method because I used pw-loopback sinks with Carla and they were showing inside pavucontrol applications list, so it's not really what you're looking for.

You can use a null sink naming it `Null Device Sink`. In this way the user should not be confused because he/she will think it's not something related to easyeffects, so it should not be used.

Talking about something invisible to pavucontrol, I think whichever Pulseaudio sink emulated by Pipewire is showing inside that app, even loopback sink showing in application list. Only things not catched are Pipewire native nodes and jack hosts.

I think that channel remapping is working with Pulseaudio sinks because Pipewire has to emulate Pulseaudio behaviour. Meanwhile it's not implemented yet between Pipewire native filters.

Digitalone1:
@wwmm I made it and it's not showing in pavucontrol! 

Will make the PR.

Issue #1041, opened 2021-07-23
Subwoofer not working after switching to easyeffects


susonthapa:
The sound is only comming from the speaker but not the subwoofer. If I kill `easyeffects` then the sound comes from the subwoofer as well.
I had this issue previously https://github.com/wwmm/easyeffects/issues/953 and  it was fixed after I set the downmixing in `/etc/pulsewire/client.conf`. 

However after the latest update (switching from **Pulseeffects** to **Easyeffects**), the subwoofer is not working. This time I believe it might be `easyeffects` issue as other apps are working fine. Any help would be appreciated. Also I took the screenshot of the `Pipewire` tab and as you can see `easyeffects` is not using `pipewire-pulse` API. I don't know much about that but it's just something I noticed.

![image](https://user-images.githubusercontent.com/33973551/126745968-f622526a-2513-4d45-83e0-ee725d9b6bac.png)

wwmm:
>easyeffects is not using pipewire-pulse API

That is not the problem. EasyEffects uses native PipeWire filters to implement its plugins. There are many advantages but also some difficulties in the new approach. Pipewire's media session does not handle links between filters. So I have to link each channel manually. And as our filters have only 2 channels they are being linked only to the left and right channels of your sound card.

I will open an issue at PipeWire's page about this. I do not think that doing the upmixing on our side is the best solution. THe server is in a much better position to do that right.

wwmm:
Link to the issue I have opened https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/1447

Digitalone1:
Can be closed as a duplicate of #1126.

A possible workaround is reported in https://github.com/wwmm/easyeffects/issues/1126#issuecomment-1141315558

PR #1104, opened 2021-08-17
Virtual Device for output stream down/upmixing

Digitalone1:
Related to #1041 

As an attempt to make down/umixing, my idea was to create an additional sink to connect to the output device so that channel remap could be done by Pipewire session manager.

I first tried with a null sink, but failed because Pipewire sees it like something that does not emit output, indeed trying to connect it to output device leads to making the link from the device to the sink, which is the inverse of what I wanted.

So I created instead a null source which is seen as something outputting and the connection is made in the right direction, from the null source to the output device. In connect_filters method of StreamOutputEffects, the manual linking from output_level to this null source is correctly done. 

Tested, it's working. Since it's a null source, it's not showed as output sink/device in pavucontrol.

Schermata del 2021-08-17 14-33-41

Schermata del 2021-08-17 14-34-04

But as a downside, it's showing as a source in source tab. It remains to test if the upmixing is happening. I can't test it because I have only a stereo laptop

Anyway, even if it's now happening, it's a good test to know how Pipewire handles these things. Read here as a [reference](https://gitlab.freedesktop.org/pipewire/pipewire/-/wikis/Virtual-Devices).

I named it Virtual Output Device because it's connected to the output device and didn't use the EasyEffects word because I thought it was showing in pavucontrol as a selectable for output apps and didn't want to make the user confuse.

If this is the path we have to choose to make the upmixing working, something has to be redefined. Surely I'm missing something. The source remains connected to the device, don't know if it's better to disconnect it and don't know how to do it. And I don't even know what happens when the default output device is changed.

### DIFF
--- a/include/pipe_manager.hpp
+++ b/include/pipe_manager.hpp
@@ -186,7 +186,7 @@ class PipeManager {
 
   std::vector<DeviceInfo> list_devices;
 
-  NodeInfo pe_sink_node, pe_source_node;
+  NodeInfo pe_sink_node, pe_source_node, pe_virtual_output_device;
 
   NodeInfo default_output_device, default_input_device;
 
@@ -272,7 +272,7 @@ class PipeManager {
   bool context_ready = false;
 
   pw_context* context = nullptr;
-  pw_proxy *proxy_stream_output_sink = nullptr, *proxy_stream_input_source = nullptr;
+  pw_proxy *proxy_stream_output_sink = nullptr, *proxy_stream_virtual_output = nullptr, *proxy_stream_input_source = nullptr;
 
   spa_hook core_listener{}, registry_listener{};
 };
--- a/include/stream_output_effects.hpp
+++ b/include/stream_output_effects.hpp
@@ -42,6 +42,8 @@ class StreamOutputEffects : public EffectsBase {
 
   void disconnect_filters();
 
+  void connect_output_device();
+
   /*
     Do not pass nd_info by reference. Sometimes it dies before we use it and a segmentation fault happens
   */
--- a/po/it_IT.po
+++ b/po/it_IT.po
@@ -858,7 +858,7 @@ msgstr "Usa Tema Scuro"
 
 #: data/ui/general_settings.ui:101
 msgid "Shutdown on Window Close"
-msgstr ""
+msgstr "Termina alla Chiusura della Finestra"
 
 #: data/ui/general_settings.ui:138
 msgid "Reset All Settings"
--- a/src/pipe_manager.cpp
+++ b/src/pipe_manager.cpp
@@ -1163,6 +1163,20 @@ PipeManager::PipeManager() {
   proxy_stream_output_sink = static_cast<pw_proxy*>(
       pw_core_create_object(core, "adapter", PW_TYPE_INTERFACE_Node, PW_VERSION_NODE, &props_sink->dict, 0));
 
+  // loading our virtual output device (needed for output channel down/upmixing)
+
+  pw_properties* props_virtual = pw_properties_new(nullptr, nullptr);
+
+  pw_properties_set(props_virtual, PW_KEY_NODE_NAME, "easyeffects_virtual_output_device");
+  pw_properties_set(props_virtual, PW_KEY_NODE_DESCRIPTION, "Virtual Output Device");
+  pw_properties_set(props_virtual, "factory.name", "support.null-audio-sink");
+  pw_properties_set(props_virtual, PW_KEY_MEDIA_CLASS, "Audio/Source/Virtual");
+  pw_properties_set(props_virtual, "audio.position", "FL,FR");
+  pw_properties_set(props_virtual, "monitor.channel-volumes", "true");
+
+  proxy_stream_virtual_output = static_cast<pw_proxy*>(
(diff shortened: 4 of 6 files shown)


wwmm:
I think it is better to have a seperated pull request for the translation update. That work on the upmixing/downmixing may take time. And testing.

>I first tried with a null sink, but failed because Pipewire sees it like something that does not emit output, indeed trying to connect it to output device leads to making the link from the device to the sink, which is the inverse of what I wanted.

It does have ouputs. Its monitors act as source ports. We link to these ports in `EasyEffects Sink` and the session manager was totally capable of linking the GStreamer stream to these monitor ports. Maybe the order of the nodes in your command was the problem.

>Anyway, even if it's now happening, it's a good test to know how Pipewire handles these things. Read here as a reference.

I know about this page. The problem the approach taken there is intentionally designed so that the devices are visible in Pulseaudio clients like Pavucontrol or Gnome control center. What is what we want in some situations. But I would like to avoid in this one.

>The source remains connected to the device, don't know if it's better to disconnect it and don't know how to do it. And I don't even know what happens when the default output device is changed.

The link from the last node to the output device has to follow the default device or the manual selection done by the user. This is something we already do but would need adaptation whatever solution is applied to this problem.

Digitalone1:
> I think it is better to have a separated pull request for the translation update. That work on the upmixing/downmixing may take time. And testing.

Sure, already did that.

> It does have ouputs. Its monitors act as source ports. We link to these ports in `EasyEffects Sink` and the session manager was totally capable of linking the GStreamer stream to these monitor ports. Maybe the order of the nodes in your command was the problem.

That's weird, I tried to change the order, but it was still linking them from output device to null sink. I'll try to make another attempt.
 
> The link from the last node to the output device has to follow the default device or the manual selection done by the user. This is something we already do but that will need adaptation to whatever solution is applied to this problem.

Maybe we have to disconnect the virtual device from the physical device after disconnect_filters() method, because if the default one is changed, the risk is that on playback start the virtual node is connected to two devices. But I don't know how to do it.

Digitalone1:
I confirm that setting:

```
pw_properties_set(props_virtual, PW_KEY_MEDIA_CLASS, "Audio/Sink");
```

and 

```
pw_metadata_set_property(pm->metadata, pm->pe_virtual_output_device.id, "target.node", "Spa:Id",
                           std::to_string(pm->output_device.id).c_str());
```

or

```
pw_metadata_set_property(pm->metadata, pm->output_device.id, "target.node", "Spa:Id",
                           std::to_string(pm->pe_virtual_output_device.id).c_str());
```

In both cases I have the two nodes linked from output device to null sink. 

Schermata del 2021-08-17 17-50-51

The only way to invert the direction is setting the virtual node as null source.

Digitalone1:
With all of the changes made, this branch is useless. But the things done here could be taken as reference in the future for the implementation of upmixing and downmixing.

wwmm:
After reading PipeWire's wiki entry for Virtual Devices I got the feeling that the automatic remapping is a thing only for streams and not for nodes.
```
There are two types of virtual devices:

Single nodes in the graph. They are usually implemented with null-sink and have input/output ports that need to be linked manually (with pw-link) after creation. They can be used, for example, to create inputs/outputs for processing chains.

Coupled streams. This is internally implemented as a capture and playback stream that can implement some processing such as channel mixing or remapping. They automatically connect to default/specified source/sinks and can be moved with tools like pavucontrol.
```
In that case the solution in this pull request would not work. And the solution with streams will create streams visible to the user and the he can move around in a way that can break the routing =/. Maybe I will have to think about some kind of custom output level filter that has more ports and does some kind of rudimentary upmixing/downmixing...

Digitalone1:
This wasn't tested completely. As I showed you, the null sink is not showed for output applications in pavucontrol and that's a good point.

Unfortunately it's present in source streams, but the user that need it won't go to touch it and breaking routing.

I think this solution should be tested because doing a custom upmixing would add complexity which is not needed now from a developer point of view.

wwmm:
>I think this solution should be tested

Testing is another problem. I do not have the proper hardware for that at this moment. My speakers are 2.0. When I have the time I will read some of the session manager code to see what it is doing in the link phase.

Digitalone1:
Me too, I got only a stereo output. 

Anyway this comes after other things that have more priority, like the blocklist which is not 100% completed and the app info UI that needs to be improved with a new system for node info update.

Digitalone1:
It's unbelievable, but this is not working anymore. I suspect a change in the last pipewire version is preventing this solution to route from a virtual source to an output device. It's not working either with a null sink. 

I remember with virtual source it was connecting the virtual source output to speakers input and with a null sink it was connecting the speakers output to the null sink input.

At this state there's no easy solution that does not involve a loopback device to do channel remapping, but it's too intrusive since it's always shown in pavucontrol. Unfortunately Pipewire does not make the work easy for us. :(

wwmm:
For now let's live with the current situation. Once I have more free time in my hands I will take a look at PipeWire's code again to see if there is something else we can do that does not involve showing streams to the users. I would really like to keep this clean as it is right now. Showing only one in virtual input/output to the users is less confusing. And there is less room for bugs caused by things messing with the additional streams.

Digitalone1:
This can be closed since it's not working anymore.

PR #1105, opened 2021-08-17
Add info about EasyEffects Flatpak to README

vchernin:
The Flatpak has gotten a “LGTM” so I think I will ask to merge it later today. I’ll update it to the latest master with the latest changes before the 6.1.0 release, if that sounds good to you. I also wanted to update the readme, I will mark this as ready to review only when the Flatpak is actually live.

- Remove command line instructions for Flatpak, these are given on the very Flathub page we link to, so I feel they are a bit unecessary here.
- Explain possible benefits of Flatpak.
- Uncomment distribution packages that have stable packages.

Questions:
- I am not exactly sure if the "Community packages" header serves a purpose here. What is the difference between EasyEffects in the Arch repo and a future EasyEffects in the Debian repo that makes Debian a community package?
- What do you think of the download from Flathub button? Would you prefer a text link instead so it’s uniform with the distribution packages?

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,20 +82,20 @@ You can help me to keep developing EasyEffects through donations. Any amount wil
 
 ## Installation
 
-### GNU/Linux packages
+### Distribution packages
 
-<!-- - [Arch Linux](https://www.archlinux.org/packages/community/x86_64/easyeffects/) / [AUR Git version](https://aur.archlinux.org/packages/easyeffects-git/)
-- [Void Linux](https://github.com/void-linux/void-packages/blob/master/srcpkgs/easyeffects/template)
+- [Arch Linux](https://www.archlinux.org/packages/community/x86_64/easyeffects/) / [AUR Git version](https://aur.archlinux.org/packages/easyeffects-git/)
+- [openSUSE](https://software.opensuse.org/package/easyeffects)
+- [NixOS](https://search.nixos.org/packages?channel=unstable&show=easyeffects&query=easyeffects)
 - [Gentoo](https://packages.gentoo.org/packages/media-sound/easyeffects)
+
+<!--
+- [Void Linux](https://github.com/void-linux/void-packages/blob/master/srcpkgs/easyeffects/template)
 - [ROSA](https://abf.io/import/easyeffects/)
 - [ALT Linux](https://packages.altlinux.org/Sisyphus/srpms/easyeffects/)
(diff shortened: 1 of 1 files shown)


wwmm:
>I am not exactly sure if the "Community packages" header serves a purpose here. What is the difference between EasyEffects in the Arch repo and a future EasyEffects in the Debian repo that makes Debian a community package?

I am not sure about why this was done. Maybe to separate official packages from third party repositories like a Ubuntu PPA. But I do not think it is really necessary. They can be together in the same section.  

>What do you think of the download from Flathub button? Would you prefer a text link instead so it’s uniform with the distribution packages?

It is fine. I think most users interested in the flatpak package will prefer to have the button.

vchernin:
> I am not sure about why this was done. Maybe to separate official packages from third party repositories like a Ubuntu PPA. But I do not think it is really necessary. They can be together in the same section.

Done. 

I also renamed "GNU/Linux packages" to "Distribution packages". As far as I know this makes sense since Flatpak is a GNU/Linux exclusive package format. Distribution packages is closer IMO to what actually seperates Flatpak from DEB, RPM, Arch pkgs, etc.

Even if some distributions include Flathub out of the box, it's still technically not packaged by the distribution. However, if someone like Fedora added EasyEffects to their Flatpak repo, technically that could fall under "Distribution package", but I don't see why anyone would do that.

vchernin:
Sidenote, are you interested in having Flatpak built as part of EasyEffects' CI (along with the Arch one)? There is a flatpak-builder GitHub action, so I don't think it should be too difficult. There are probably more details that would need to be discussed (in another issue), but I won't start if it's not something you think should be added.

wwmm:
>Sidenote, are you interested in having Flatpak built as part of EasyEffects' CI (along with the Arch one)?

I don't know. Maybe it would help us to see situations like that one related to the use of "too new" compiler features. If it is easy to do go ahead. But if it gives too much trouble it may not be worth the effort.

vchernin:
Ok, build should be on Flathub stable within a few hours.

@wwmm Can you confirm you got an invite to be collaborator? If not I'll ask the admins as I don't think I am able to do so.

wwmm:
>Can you confirm you got an invite to be collaborator? If not I'll ask the admins as I don't think I am able to do so.

Yes.

vchernin:
@wwmm Flatpak is live now, please merge this

PR #1107, opened 2021-08-18
Track pw proxies on global bypass

Digitalone1:
I noticed that pw proxies returned by link_nodes() are saved and then destroyed in disconnect_filters(). This is happening when bypass is disabled, but when it is enabled the proxies are not tracked into the vector, so they are not destroyed when disconnect_filters() is called another time on bypass disabling.

Don't know if this is an intended behaviour, but I thought it's better to track them and destroy them afterwards, like it's usually made.

### DIFF
--- a/include/stream_input_effects.hpp
+++ b/include/stream_input_effects.hpp
@@ -38,7 +38,7 @@ class StreamInputEffects : public EffectsBase {
  private:
   bool bypass = false;
 
-  void connect_filters();
+  void connect_filters(const bool& bypass = false);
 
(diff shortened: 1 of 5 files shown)


wwmm:
>Don't know if this is an intended behaviour

I do not remember having a reason for not doing it. I probably just forgot about them. That proxy tracking is something I did not do from the beginning. Initially I was just destroying the links based on their id. But at some point not destroying the proxies became a problem in some operations. The issue probably did not manifest itself in the bypass so I did not think about them.

PR #1115, opened 2021-08-25
Integrate libportal

vchernin:
- Builds in Flatpak, but doesn't do anything yet

Todo:
- Add background portal test
- make libportal optional

### DIFF
--- a/include/general_settings_ui.hpp
+++ b/include/general_settings_ui.hpp
@@ -27,6 +27,7 @@
 #include <filesystem>
 #include "application.hpp"
 #include "util.hpp"
+#include "libportal/background.h"
 
 class GeneralSettingsUi : public Gtk::Box {
  public:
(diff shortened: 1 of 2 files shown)


vchernin:
Ok, I am really stumped why I am unable to change the content in the Exec line.

If I build with something like `- desktop-file-edit --set-key=Name --set-value=EasyEffects-test-62 "${FLATPAK_DEST}/share/applications/${FLATPAK_ID}.desktop"` in `post-install` that change is reflected correctly. The name in the GNOME app launcher adjusts.

However no matter what I try it seems my edits to the Exec line are just ignored. I can enter in a bogus directory that doesn't exist, and yet somehow the final build will still work. I even tried patching the file upstream and that still didn't reflect changes to the Exec line...

I'll try maybe fully deleting my installs with `--delete-data`.

Digitalone1:
This can be closed.

PR #1127, opened 2021-09-02
Code optimizations and fixes

Digitalone1:
I apologize for the last commit being so huge. The problem is that I kept changing the code and forgot to commit, then I found myself with a big one.

So this is the list of changes:

- in folder monitor signal handler, always check for existing items in the list before appending
- prefer pre-increment rather than post-increment for iterators
- preferer references and const references where is possible
- avoid temporary local variables where possible
- if possible, innest local variables in if statements rather than having them on the outside scope
- variables that do not change are marked as const
- if pointers to const are not possible, we do const pointers
- use more `Glib::ustring` rather than `std::string` since Glib widgets most of the time expect this type (unfortunately lots of implicit conversions can't be avoided, but we try to reduce them)
- if we need `std::string` explicitly, convert with `raw` or `c_str` in assignments or comparisons (but when setting text in labels, always avoid c_str)
- parameters passed to signals were mismatched in some parts of the code doing unnecessary conversions: now we try to reduce conversions, set const references for plugins parameters and copy NodeInfo (because Pipewire can unset the data before the signal handler use it)
- cstring header removed: it was used only for `strcmp`, now we use `g_strcmp0` of Glib that perform the same task
- removed other unneeded headers
- we now use `Glib::ustring::format` for string formatting, getting rid of stringstreams that were longer to write and manage
- level_to_localized_string_showpos and float_to_localized_string mothods removed and substituted in place with the more powerful and short Glib::ustring::format
- `std::stof` is locale independent even if cplusplus.com states the opposite, so we use it rather than stringstreams
- values without unit in plugin interfaces now don't show a useless final whitespace
- validate user input text for new presets and truncate to 100 characters if longer
- the app icon name is unset when node info state is `PW_NODE_STATE_CREATING` (it happens sometimes when the application is disabled and then it's removed from the blocklist), however we can't check if the icon name is empty in that case because Pipewire fill it with some whitespaces, so we hide it from the app info UI 
- use node_state_to_string method (that was unused) to set the stream state in app info UI
- removed blocklist_in and blocklist_out from pipe_manager: they were unused
- now the previous state of blocklisted apps removed from the blocklist is correctly restored saving the state in a map and associating it to the node id (previously it was saved in NodeInfoHolder structure that was not storing the state correctly)
- now the blocklist management is more robust: the app is added/removed to/from the blocklist even on new presets that have non-empty blocklist

I tested it, it's working. I had some issues on strings not showing due to the changes, but then I fixed it. @wmm I only ask you to test the input pipeline, the input blocklist and the external sidechain that I can't test.

For other known issues we talked about, I'm working on them in the next days.

### DIFF
--- a/data/ui/app_info.ui
+++ b/data/ui/app_info.ui
@@ -25,6 +25,7 @@
 
                 <child>
                     <object class="GtkImage" id="app_icon">
+                        <property name="visible">False</property>
                         <property name="halign">start</property>
                         <property name="valign">center</property>
                         <property name="icon_size">large</property>
@@ -222,6 +223,7 @@
 
                 <child>
                     <object class="GtkScale" id="scale_volume">
+                        <property name="visible">False</property>
                         <property name="hexpand">1</property>
                         <property name="valign">center</property>
                         <property name="adjustment">volume</property>
--- a/data/ui/crossfeed.ui
+++ b/data/ui/crossfeed.ui
@@ -36,7 +36,6 @@
                 <child>
                     <object class="GtkButton" id="preset_jmeier">
                         <property name="label" translatable="yes">Jmeier</property>
-                        <property name="visible">True</property>
                         <property name="can-focus">True</property>
                         <property name="receives-default">True</property>
                     </object>
--- a/data/ui/effects_base.ui
+++ b/data/ui/effects_base.ui
@@ -250,7 +250,7 @@
 
                 <child type="end">
                     <object class="GtkToggleButton" id="toggle_listen_mic">
-                        <property name="visible">0</property>
+                        <property name="visible">False</property>
                         <child>
                             <object class="GtkBox">
                                 <property name="halign">center</property>
--- a/data/ui/limiter.ui
+++ b/data/ui/limiter.ui
@@ -459,7 +459,7 @@
 
                 <child>
                     <object class="GtkBox">
-                        <property name="visible">0</property>
+                        <property name="visible">False</property>
                         <property name="halign">center</property>
                         <property name="valign">center</property>
                         <property name="spacing">6</property>
@@ -483,7 +483,6 @@
 
                 <child>
                     <object class="GtkBox">
-                        <property name="visible">1</property>
                         <property name="halign">center</property>
                         <property name="valign">center</property>
                         <property name="spacing">24</property>
--- a/data/ui/multiband_compressor_band.ui
+++ b/data/ui/multiband_compressor_band.ui
@@ -207,7 +207,7 @@
 
                         <child>
                             <object class="GtkLabel">
-                                <property name="visible">0</property>
+                                <property name="visible">False</property>
                                 <property name="halign">end</property>
                                 <property name="valign">center</property>
                                 <property name="label" translatable="yes">Device</property>
@@ -334,7 +334,7 @@
 
                         <child>
                             <object class="GtkDropDown" id="dropdown_input_devices">
-                                <property name="visible">0</property>
+                                <property name="visible">False</property>
                                 <property name="valign">center</property>
                                 <property name="hexpand">1</property>
 
--- a/include/application.hpp
+++ b/include/application.hpp
@@ -58,12 +58,12 @@ class Application : public Gtk::Application {
   void on_activate() override;
 
  private:
-  std::string log_tag = "application: ";
+  const std::string log_tag = "application: ";
 
   bool running_as_service = false;
 
   void create_actions();
-  void update_bypass_state(const std::string& key);
+  void update_bypass_state(const Glib::ustring& key);
 };
 
 #endif
--- a/include/application_ui.hpp
+++ b/include/application_ui.hpp
@@ -43,7 +43,7 @@ class ApplicationUi : public Gtk::ApplicationWindow {
   static auto create(Application* app) -> ApplicationUi*;
 
  private:
-  std::string log_tag = "application_ui: ";
+  const std::string log_tag = "application_ui: ";
 
   Application* app;
 
--- a/include/autogain.hpp
+++ b/include/autogain.hpp
@@ -42,9 +42,13 @@ class AutoGain : public PluginBase {
                std::span<float>& left_out,
                std::span<float>& right_out) override;
 
-  // loudness, gain, momentary, shortterm, integrated, relative, range;
-
-  sigc::signal<void(double, double, double, double, double, double, double)> results;
+  sigc::signal<void(const double&,              // loudness
+                    const double&,              // gain
+                    const double&,              // momentary
+                    const double&,              // shortterm
+                    const double&,              // integrated
+                    const double&,              // relative
+                    const double&)> results;    // range
 
  private:
   bool ebur128_ready = false;
--- a/include/bass_enhancer.hpp
+++ b/include/bass_enhancer.hpp
@@ -42,7 +42,7 @@ class BassEnhancer : public PluginBase {
                std::span<float>& left_out,
                std::span<float>& right_out) override;
 
-  sigc::signal<void(double)> harmonics;
+  sigc::signal<void(const double&)> harmonics;
 
  private:
   std::unique_ptr<lv2::Lv2Wrapper> lv2_wrapper;
--- a/include/bass_enhancer_ui.hpp
+++ b/include/bass_enhancer_ui.hpp
@@ -36,7 +36,7 @@ class BassEnhancerUi : public Gtk::Box, public PluginUiBase {
 
   static auto add_to_stack(Gtk::Stack* stack, const std::string& schema_path) -> BassEnhancerUi*;
 
-  void on_new_harmonics_level(double value);
+  void on_new_harmonics_level(const double& value);
 
   void reset() override;
 
--- a/include/calibration_mic.hpp
+++ b/include/calibration_mic.hpp
@@ -22,7 +22,6 @@
 
 #include <gst/gst.h>
 #include <sigc++/sigc++.h>
-#include <iostream>
 #include <vector>
 
 class CalibrationMic {
@@ -33,8 +32,6 @@ class CalibrationMic {
   CalibrationMic(const CalibrationMic&&) = delete;
   auto operator=(const CalibrationMic&&) -> CalibrationMic& = delete;
   ~CalibrationMic();
-
-  std::string log_tag = "calibration_mic: ";
 
   GstElement *pipeline = nullptr, *source = nullptr, *sink = nullptr, *spectrum = nullptr;
 
@@ -57,6 +54,8 @@ class CalibrationMic {
   void set_input_node_id(const uint& id) const;
 
  private:
+  const std::string log_tag = "calibration_mic: ";
+
   GstBus* bus = nullptr;
 };
 
--- a/include/calibration_mic_ui.hpp
+++ b/include/calibration_mic_ui.hpp
@@ -40,7 +40,7 @@ class CalibrationMicUi : public Gtk::Box {
   std::unique_ptr<CalibrationMic> cm;
 
  private:
-  std::string log_tag = "calibration_mic_ui: ";
+  const std::string log_tag = "calibration_mic_ui: ";
 
   const double default_time_window = 2.0;  // seconds
 
--- a/include/calibration_signals.hpp
+++ b/include/calibration_signals.hpp
@@ -22,7 +22,6 @@
 
 #include <gst/gst.h>
 #include <sigc++/sigc++.h>
-#include <iostream>
 #include <vector>
 
 class CalibrationSignals {
@@ -33,8 +32,6 @@ class CalibrationSignals {
   CalibrationSignals(const CalibrationSignals&&) = delete;
   auto operator=(const CalibrationSignals&&) -> CalibrationSignals& = delete;
   ~CalibrationSignals();
-
-  std::string log_tag = "calibration_signals: ";
 
   GstElement *pipeline = nullptr, *source = nullptr, *sink = nullptr, *spectrum = nullptr;
 
@@ -55,6 +52,8 @@ class CalibrationSignals {
   void set_volume(const double& value) const;
 
  private:
+  const std::string log_tag = "calibration_signals: ";
+
   GstBus* bus = nullptr;
 };
 
--- a/include/calibration_signals_ui.hpp
+++ b/include/calibration_signals_ui.hpp
@@ -40,7 +40,7 @@ class CalibrationSignalsUi : public Gtk::Box {
   std::unique_ptr<CalibrationSignals> cs;
 
  private:
-  std::string log_tag = "calibration_signals_ui: ";
+  const std::string log_tag = "calibration_signals_ui: ";
 
   Gtk::Switch* enable = nullptr;
   Gtk::Scale* volume = nullptr;
--- a/include/calibration_ui.hpp
+++ b/include/calibration_ui.hpp
@@ -44,7 +44,7 @@ class CalibrationUi : public Gtk::Window {
   void set_input_node_id(const uint& id);
 
  private:
-  std::string log_tag = "calibration_ui: ";
+  const std::string log_tag = "calibration_ui: ";
 
   Gtk::Stack* stack = nullptr;
   Gtk::DrawingArea* spectrum = nullptr;
@@ -59,7 +59,7 @@ class CalibrationUi : public Gtk::Window {
   CalibrationSignalsUi* calibration_signals_ui = nullptr;
   CalibrationMicUi* calibration_mic_ui = nullptr;
 
-  void on_new_spectrum(const std::vector<float>& magnitudes);
+  void on_new_spectrum(std::vector<float> magnitudes);
 
   auto on_spectrum_draw(const Cairo::RefPtr<Cairo::Context>& ctx) -> bool;
 
--- a/include/compressor.hpp
+++ b/include/compressor.hpp
@@ -44,7 +44,7 @@ class Compressor : public PluginBase {
                std::span<float>& probe_left,
                std::span<float>& probe_right) override;
 
-  sigc::signal<void(double)> reduction, sidechain, curve, envelope, latency;
+  sigc::signal<void(const float&)> reduction, sidechain, curve, envelope, latency;
 
  private:
   uint latency_n_frames = 0U;
--- a/include/compressor_ui.hpp
+++ b/include/compressor_ui.hpp
@@ -20,7 +20,6 @@
 #ifndef COMPRESSOR_UI_HPP
 #define COMPRESSOR_UI_HPP
 
-#include <cstring>
 #include "info_holders.hpp"
 #include "plugin_ui_base.hpp"
 
@@ -38,13 +37,13 @@ class CompressorUi : public Gtk::Box, public PluginUiBase {
 
   static auto add_to_stack(Gtk::Stack* stack, const std::string& schema_path) -> CompressorUi*;
 
-  void on_new_reduction(double value);
+  void on_new_reduction(const float& value);
 
-  void on_new_envelope(double value);
+  void on_new_envelope(const float& value);
 
-  void on_new_sidechain(double value);
+  void on_new_sidechain(const float& value);
 
-  void on_new_curve(double value);
+  void on_new_curve(const float& value);
 
   void set_pipe_manager_ptr(PipeManager* pipe_manager);
 
--- a/include/convolver.hpp
+++ b/include/convolver.hpp
@@ -47,7 +47,7 @@ class Convolver : public PluginBase {
                std::span<float>& left_out,
                std::span<float>& right_out) override;
 
-  sigc::signal<void(double)> latency;
+  sigc::signal<void(const float&)> latency;
 
  private:
   bool kernel_is_initialized = false;
@@ -90,7 +90,7 @@ class Convolver : public PluginBase {
     std::copy(data_right.begin(), data_right.end(), conv_right_in.begin());
 
     if (zita_ready) {
-      int ret = conv->process(true);  // thread sync mode set to true
+      const int& ret = conv->process(true);  // thread sync mode set to true
 
       if (ret != 0) {
         util::debug(log_tag + "IR: process failed: " + std::to_string(ret));
--- a/include/convolver_ui.hpp
+++ b/include/convolver_ui.hpp
@@ -49,7 +49,9 @@ class ConvolverUi : public Gtk::Box, public PluginUiBase {
   void reset() override;
 
  private:
-  std::string log_tag = "convolver_ui: ";
+  const std::string log_tag = "convolver_ui: ";
+
+  const std::string irs_ext = ".irs";
 
   Gtk::SpinButton* ir_width = nullptr;
 
@@ -93,7 +95,7 @@ class ConvolverUi : public Gtk::Box, public PluginUiBase {
 
   void setup_listview();
 
-  auto get_irs_names() -> std::vector<std::string>;
+  auto get_irs_names() -> std::vector<Glib::ustring>;
 
   void import_irs_file(const std::string& file_path);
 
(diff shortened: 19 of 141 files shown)


wwmm:
In the crystalizer header I think it is fine to use const references to the vector elements in most cases. But in the lines 189 and 190 this may not be a good idea https://github.com/wwmm/easyeffects/blob/ecad81ebd4df5de1f6ac827e58860c98dd395028/include/crystalizer.hpp#L189. The reason is that the idea is saving the original value in the next iteration. But as `L` and `R` are just a reference to band_data_L.at(n)[m] that is changed before the line 189 they do not have the old value anymore.

wwmm:
Unless that is in this situation `const float&` is not passing by reference. But I have my doubts it would be passing by value.

wwmm:
IF what I said above is a problem the same thing may happen at https://github.com/wwmm/easyeffects/blob/ecad81ebd4df5de1f6ac827e58860c98dd395028/include/crystalizer.hpp#L131. I think it may be safer not using const references in that part of the crystalizer code. It is a situation where a copy has to be made.

wwmm:
`float&` will create just a reference variable indeed https://www.tutorialspoint.com/what-is-a-reference-variable-in-cplusplus. IN some places the crystalizer algorithm requires a copy that will be saved for later usage. That part will have to be fixed.

Digitalone1:
> `float&` will create just a reference variable indeed https://www.tutorialspoint.com/what-is-a-reference-variable-in-cplusplus. In some places the crystalizer algorithm requires a copy that will be saved for later usage. That part will have to be fixed.

Sure, can you do it after the merge?

wwmm:
>Sure, can you do it after the merge?

Yes.

wwmm:
It is very subtle but I think I would avoid references in situations like this one too https://github.com/wwmm/easyeffects/blob/ecad81ebd4df5de1f6ac827e58860c98dd395028/src/compressor.cpp#L207. It is possible it won´t cause problems in this particular situation. But I am not 100% confident it won't. The reason why I wrap a few emits with `Glib::signal_idle().connect_once` is that multithreading data races caused lots of headaches in the past. Most of the emits we do from the PipeWire thread are meant to be used by the main ui thread handled by GTK. And gui toolkits in general do not like other threads messing with them. I wonder if the fact that gtk will be receiving a reference to variables whose lifetime is in another thread will cause problems.

What concerns me is that the problems caused by these kind of data races tend to be random. Sometimes we do many tests and nothing unusual happens. And then one day the app mysteriously freezes. I still thing that we can merge. But I think we should be on the safe side and avoiding other threads receiving a points to variables whose lifetime is managed by another thread. It is better after the merge to make those variables a copy again.

Digitalone1:
I though that values from lv2 plugins could cause issues, but I tested for more than a week and never had issues. Anyway if you want, I can convert them to copy.

Digitalone1:
We don't destroy Pipewire filters after initialized the first time, so it could be for that reason it's not crashing even if they are references.

wwmm:
>Anyway if you want, I can convert them to copy.

It is better. When I started this project I wasted many hours, sometimes days, debugging random freezes that had not clear cause. Until I eventually realized in gdb that the reason was having a thread trying to use variables from another thread and whose lifetime had already finished in this other thread. In this particular case I think this risk does not exist. The variable lives long enough. But interaction between threads is too tricky to take the risk. If we can use a copy it is better to use one.

I am still reading the changes here on github. In a few more moments I will merge.

Digitalone1:
Okay, let's go with copy then. Anyway I think it would be better to copy in the function parameters rather than that specific place. What do you think?

wwmm:
>Anyway I think it would be better to copy in the function parameters rather than that specific place. 

I did not understand. What do you mean by  "copy in the function parameters"?

Digitalone1:
> I did not understand. What do you mean by "copy in the function parameters"?

That reduction reference is passed, but on_new_reduction function accept the value, copying it.

wwmm:
>That reduction reference is passed, but on_new_reduction function accept the value, copying it.

I see. It is risky. I think it would not be a problem in this case because that part of the code is probably free of the problem I am talking about anyway. But let's consider for the sake of argument that we were in a critical situation where the problem happens for sure. `Glib::signal_idle().connect_once` schedules a function to be executed by the main thread(gui thread) later. If we let the copy to be made later the variable from the plugin thread that contains the value may not be alive anymore. And then we would have a problem. If I remember way I had similar problems in the past. When Pulseaudio gave me a value I scheduled a function call in the main thread that sometimes happened when the local variable that had been created in the Pulseaudio thread was already out of scope. I think the only reason why you did not see any problem is because you are passing a reference to `get_control_port_value` output. And this output is a a pointer to a member variable that lives while the app is running. It never gets out of scope. 

Another situation where having references in signals can be undesirable is if the value should not be overwritten before it is used by its destination. In these filters signals it does not matter much if the plugin thread overwrites the variable before `Glib::signal_idle().connect_once` schedules the signal emission. They are just a level estimation. But in a situation where the data "can not be lost" the reference may be updated before its original context is actually used by its destination. There may not be a crash  or a freeze. But some algorithms may not work well if some bits of a data stream are missing.

Digitalone1:
I understood. So we would have to do two copies. One in reduction and another in `Glib::signal_idle().connect_once`, right?

Unfortunately all those copies have impact on performance.

wwmm:
>One in reduction and another in Glib::signal_idle().connect_once, right?

I think we do not need to pass by value in the signals slots (like the reduction callback) as long as `Glib::signal_idle().connect_once` captures a local variable by copy. So I tihnk we can make a copy only in lines like `const float& reduction_value = lv2_wrapper->get_control_port_value("rlm");`

>I only ask you to test the input pipeline, the input blocklist and the external sidechain that I can't test.

At this moment I am in the middle of a videoconference in MS Teams. The perfect place to test everything at once. So far so good. No problems.

wwmm:
>So I think we can make a copy only in lines like const float& reduction_value = lv2_wrapper->get_control_port_value("rlm");

In other words `float&` would become `float`. This should be enough.

Digitalone1:
I don't know much of signals. So you say that signals captures the variable till the end of the scope. Therefore we have to do a copy before the connection, but in the connection we can pass references to the lambda function bound to the signal, right? 

If it's so, we can also pass references of NodeInfo structures, as long as we make a copy first?

wwmm:
>I don't know much of signals. So you say that signals captures the variable till the end of the scope. Therefore we have to do a copy before the connection, but in the connection we can pass references to the lambda function bound to the signal, right?

The lambda capture inside `Glib::signal_idle().connect_once` will make a copy of the local variable for itself. So even if the original `reduction_value` variable gets out of scope it won't be a problem. When the scheduled function is executed the copy tha was made during the lambda capture will be used. No relations with the other thread variable anymore. So we should be able to use references in the signal slots.

>If it's so, we can also pass references of NodeInfo structures, as long as we make a copy first?

That is what I thought. But for some weird reason I could not understand I had some crashes  in the beginning of the port that seemed to be linked to the lifetime of these structures. Maybe it was a false positive. I think it may be worth to test this particular case again. When I started the port there was so much to do I could not properly investigate why that particular case seemd to cause problems. Maybe how the code is now it is fine.

wwmm:
>but in the connection we can pass references to the lambda function bound to the signal, right?

I think that while there is still code running inside the lambda the variable copy should still live. Unless that there is something subtle I am not aware of.

wwmm:
The worst-case scenario would be `emit` getting out of the lambda scope before the slot used the variable. If that happens maybe the variable copy does before the slot is called. Maybe that was the reason why I had some crashes when passing NodeInfo by reference. If this kind of thing  really can happen we will have to pass by values in the slot too. The problem is how to verify that in a reliable way.

wwmm:
>The worst-case scenario would be emit getting out of the lambda scope before the slot used the variable.

I think that this kind of thing could happen if the emit operations are done asynchronously. In that case `emit` would finish and get out of the scope after scheduling the emission. And with that the original copy would be destroyed as there is nothing more to be done inside the lambda. By the time the slots run it would have a reference to a dead variable.

The question is if this kind of thing can really happen. Maybe that was what caused the crashes I saw before with the NodeInfo structures. It is hard to say. IS this performance penalty really noticeable?

Digitalone1:
I'm thinking that the most reliable think would be copy those plugin port values in specific plugin class properties, so they are alive until the application is terminated (we do not start all plugin filters at startup, but once started when needed, their not destroyed).  In that case we can pass by reference.

And since these ports are only for showing estimations in the UI, we don't care about values modified/overwritten.

wwmm:
>I'm thinking that the most reliable think would be copy those plugin port values in specific plugin class properties, so they are alive until the application is terminated (we do not start all plugin filters at startup, but once started when needed, their not destroyed). In that case we can pass by reference.

In a way that is already there. `get_control_port_value` gets its values from a vector that is a class member. As long as the plugin class is not destroyed we should be fine. Let's keep the code for the ports as it is in this pull request for a while and see what happens.

But in the slots for other things like the NodeInfo structures we have to be careful. I think that there is a potential for the use of references in signals to surprise us negatively. And in this case we do not want the data to be erased. THis may break some parts of our code depending on which information is lost.

wwmm:
>Sure, can you do it after the merge?

Done.

wwmm:
@Digitalone1 there is a small issue with the new code handling the icons that I only noticed now that I installed a package with the new code
Screenshot from 2021-09-02 17-55-49
When we are running as a service and the window is closed and then opened again the icon is missing. Disabling effects for the app and enabling it again makes the icon visible again.

wwmm:
That line seems suspicious https://github.com/wwmm/easyeffects/blob/e7597e192c1cee2db1f30500c2fa3c5a5ef875fe/src/effects_base_ui.cpp#L710. When the window is reopened the audio app has already doe its initialization so its state probably won't be `PW_NODE_STATE_CREATING` by the time this code runs.

Digitalone1:
I think it's because I made them invisible as default. Will make visible as default and change the property on the specific value.

wwmm:
>I think it's because I made them invisible as default. Will make visible as default and change the property on the specific value.

I see. But is `app_icon->set_from_icon_name` called somewhere else to set the icon? If not the problem will still happen.

Digitalone1:
> That line seems suspicious
> 
> https://github.com/wwmm/easyeffects/blob/e7597e192c1cee2db1f30500c2fa3c5a5ef875fe/src/effects_base_ui.cpp#L710
> . When the window is reopened the audio app has already doe its initialization so its state probably won't be `PW_NODE_STATE_CREATING` by the time this code runs.

I did it because sometimes on that state the icon is broken and the volume scale is not functional, so I didn't like this and wanted to hide them. Don't know why they're missing when the window is started, I set them visible on other states.

Digitalone1:
> I see. But is `app_icon->set_from_icon_name` called somewhere else to set the icon? If not the problem will still happen.

The problem is the visible property, not the set icon name.

Digitalone1:
I'm fixing it tomorrow.

wwmm:
>I'm fixing it tomorrow.

Ok.

wwmm:
>The problem is the visible property, not the set icon name.

It makes sense. For some reason I read `i.state != PW_NODE_STATE_CREATING` as `i.state == PW_NODE_STATE_CREATING`

Digitalone1:
It seems that I should use opacity property as done before.

wwmm:
Besides the icon that text that is displayed with media information below the app name also does not show. But I imagine the same fix will serve to both.

Digitalone1:
I found out why.

In info_updated signal NodeInfo was passed by reference. See [here](https://github.com/wwmm/easyeffects/blob/e58d195d4840c4d32f38b650c956e9dec4fd04aa/src/effects_base_ui.cpp#L740).

Since I wanted to unify all NodeInfo passing them by copy, now it makes a copy (see [here](https://github.com/wwmm/easyeffects/blob/e7597e192c1cee2db1f30500c2fa3c5a5ef875fe/src/effects_base_ui.cpp#L701)), but this copy have the state I mentioned before, without icon, media name and ability to change volume.

Did you make it explicitly as a reference because you already observed this behaviour?

Digitalone1:
I changed it to reference, but I keep getting the same state 

Anyway I don't fully understand a thing. Before my changes, you defined the `info_updated` signal as a void function with NodeInfo parameter by value. See [here](https://github.com/wwmm/easyeffects/blob/e58d195d4840c4d32f38b650c956e9dec4fd04aa/include/info_holders.hpp#L32)

Then where you connect a lambda function to this signal, that lambda is a void function with NodeInfo passed by reference. See [here](https://github.com/wwmm/easyeffects/blob/e58d195d4840c4d32f38b650c956e9dec4fd04aa/src/effects_base_ui.cpp#L740). Isn't this a contradiction?

I thought it was a mistake and the compiler was implicitly converting it as value copy. So that means the signal is copying it, then it's passing the reference to the copy? Or maybe the compiler is implicitly converting the whole passage as reference?

Anyway, even with the reference it's not working. I have to investigate why.

Playing with log messages I see that the info_updated is emitted when the window is show from the lambda in `factory->signal_bind().connect`. But the NodeInfo we got is the one with that state and the only way to get updated info is to trigger an operation like muting or enable/disable.

I wonder why it was working before, maybe I have to test that part of the code before my changes.

Digitalone1:
I did not consider that methods in effects_base_ui where not executed when the window is not showed.

The problem is that on the last release the blocklist wasn't working because there wasn't the check on the app blocklisted, indeed blocklist_in/out in pipe_manager were unused. If pipewire was redirecting the stream to EE, it was processed even if the app was in the blocklist.

I added the check in on_app_added inside effects_base_ui and the blocklist is now working, but only if EE is started with the window open. Without the window, the operation of checking the blocklist and disconnect if the app is blocklisted have to be moved outside effects_base_ui. :(

Digitalone1:
So there are interesting news. 

I resolved the issue related to the app info ui changing `EffectsBaseUi::on_app_added` and making NodeInfo& parameter as a reference.

Lambda connected to info_updated still makes a copy of NodeInfo, but it's showing the icon correctly. I suspect this is because the `NodeInfoHolder::create(node_info)` creates a copy of the reference at a time where the app is not in the creating state, while when on_app_added was making a copy, it copied the first state of the struct that had the creating state.

The problem is that when the EE ui is showed, we show old info about the application, because that was a copy of the create function, even if they seems corrected. The only way to show updated info is to trigger an operation (mute/volume changes...) and calling on_app_changed.

About the blocklist, it's not working because there's something wrong in `StreamOutputEffects::on_app_added` since that code is not able to detect the connected state of the stream, so when it's blocklisted, the disconnection is not triggered.

Digitalone1:
I'm doing tests. I can confirm that checking if the app is enabled in `StreamOutputEffects::on_app_added` is useless because is always detected as disconnected, even if it should be connected and it's going to be connected afterwards.

Digitalone1:
When the window is showed the blocklist works as expected because there's a check in the setup list view where the check on the enabled state is working (this is for sure, I verified it many times) and the blocklist it's activated there.

But when the window is hidden, `StreamOutputEffects::on_app_added` is the only responsible to enable/disable the app, but the check on the enabled state always fail, so the blacklist is not activated.

The stream is connected anyway when the `process-all-outputs` settings is enabled, even if the stream is already connected to the pipeline.

Since the enabled state is not working there, maybe because it's too early, I'm trying to find another approach, maybe enabling/disabling the app only considering blocklist and process-all-outputs.

Digitalone1:
So I changed how the new application is connected to the pipeline in `StreamOutputEffects::on_app_added(NodeInfo node_info)`

```
if (is_blocklisted) {
    pm->disconnect_stream_output(node_info);
  } else if (global_settings->get_boolean("process-all-outputs")) {
    pm->connect_stream_output(node_info);
  }
``` 

Blocklist it's now working both on window and service mode.

But with this changes I now can't get the correct information on `EffectsBaseUi::on_app_added(NodeInfo node_info)`.

I thinks it's because without the check on the enabled state, the code is faster and the following line

`all_players_model->append(NodeInfoHolder::create(pd->nd_info));`

is called with a copy of the nodeinfo containing the `PW_NODE_STATE_CREATING` without useful info like the icon name. So every update triggered manually passing the `holder` is giving old app info.

I managed to resolve this issue in a raw way seeing how you get the node info copy in pipe_manager.

```
void EffectsBaseUi::on_app_added(NodeInfo node_info) {
  for (guint n = 0U; n < all_players_model->get_n_items(); n++) {
    if (all_players_model->get_item(n)->info.id == node_info.id) {
      return;
    }
  }

  struct node_data {
    pw_proxy* proxy = nullptr;
    spa_hook proxy_listener{};
    spa_hook object_listener{};
    PipeManager* pm = nullptr;
    NodeInfo nd_info;
  };

  auto* pd = static_cast<node_data*>(pw_proxy_get_user_data(node_info.proxy));

  all_players_model->append(NodeInfoHolder::create(pd->nd_info));

  // Blocklist check

  if (app_is_blocklisted(node_info.name) && !settings->get_boolean("show-blocklisted-apps")) {
    return;
  }

  players_model->append(NodeInfoHolder::create(pd->nd_info));
}
```

In this way I can get the correct updated info on window opening in service mode. So we can be sure that it's a problem related to the time the copy of NodeInfo struct is done.

Now we have to find a way to get the NodeInfo at an updated state when arriving in `EffectsBaseUi::on_app_added(NodeInfo node_info)`

Passing by reference would solve the issue, but it's not an option because can lead to segmentation fault ifPpipewire releases the NodeInfo structure.

So I suggest two options: 

1. passing the proxy and get the NodeInfo updated when we need it at window opening
2. find a way to make a timeout and wait some time (100-300 ms I think) to get the struct with `PW_NODE_STATE_RUNNING` (this way we may solve the icon issue, but info will still be old in the window is opened in the middle of the stream)

wwmm:
>Anyway I don't fully understand a thing. Before my changes, you defined the info_updated signal as a void function with NodeInfo parameter by value. See here

>Then where you connect a lambda function to this signal, that lambda is a void function with NodeInfo passed by reference. See here. Isn't this a contradiction?

I am not sure about what the compiler does in this case. I may have made it a reference for the habit of doing so. The compiler may have silently turned that into a copy to match the signal declaration. It is a good question. 

Now I see at some point I added a call to `info_updated.emit` inside its own slot https://github.com/wwmm/easyeffects/blob/7d084e299cd2999d3746903c43e59e282a26ddab/src/effects_base_ui.cpp#L845. I probably did that by mistake. IT is not a good idea to emit the signal inside its own callback. GTK probably has protection against this kind of callback loop but it is probably better to fix this.

wwmm:
>I managed to resolve this issue in a raw way seeing how you get the node info copy in pipe_manager.

That sounds like an overkill. pipe_manager already has a list o nodes that should be updated with the latest information whenever these signals come. If the info that comes as argument of the signal slot is old iterating through the list managed by pipe_manager should be enough as the info saved there should be the same as the one returned by `pw_proxy_get_user_data`

wwmm:
When the window is opened this iteration is done by `StreamOutputEffectsUi` constructor https://github.com/wwmm/easyeffects/blob/e7597e192c1cee2db1f30500c2fa3c5a5ef875fe/src/stream_output_effects_ui.cpp#L27. There `on_app_added` is called with the node info. IF that info is outdated somehow then the real issue is the list not being properly updated for some reason. And the fix for that has to be done in the pipe_manager class.

wwmm:
Looking at the code the gets the "changed" signals about nodes I think it is missing that update https://github.com/wwmm/easyeffects/blob/e7597e192c1cee2db1f30500c2fa3c5a5ef875fe/src/pipe_manager.cpp#L175

`nd->nd_info` is being populated with the latest info and a copy of it is sent in the signals we emit. But It seems I forgot to write something like `node = nd->nd_info;` to update the list before emitting the signals. This may be the reason why the info is old. I do not think there is another place updating each  `pm->list_nodes` element. This should be done inside `on_node_info` but I forgot =/

Digitalone1:
There you made the connection, but the NodeInfo passed is the one from here

https://github.com/wwmm/easyeffects/blob/e7597e192c1cee2db1f30500c2fa3c5a5ef875fe/src/pipe_manager.cpp#L835

I don't fully understand, but that `nd_info_copy` is made right after the application is started? If yes, it's reasonable the state is still at `PW_NODE_STATE_CREATING`.

I wanted to ask you a curiosity since you know Pipewire better then me. Is there an operation on the stream we can do that is invisible to the user, but it makes the signal stream change activate so that the update is done from `on_app_changed`?

wwmm:
>There you made the connection, but the NodeInfo passed is the one from here

I had to pass its copy because of those damn lifetime issues caused segmentation faults at some point. Nodes do not leave forever. They may be created and destroyed for a number of reasons. And when that happened and I tried to use its pointer I had a segmentation fault. So we will have to do our best to ensure that this list is updated. Maybe the `node = nd->nd_info; ` I forgot will be enough to keep `list_node` with the info we need.

>Is there an operation on the stream we can do that is invisible to the user, but it makes the signal "stream changes" activate so that the update is done from on_app_changed?

Assuming that the permission was not denied (and it could be if the user configures PipeWire that way) only the same that the user can do on its own. So I would not say they are "invisible".

Digitalone1:
So I suggest the following points:
* pipe_manager has the `list_node` updated
* we pass only the node id to on_app_added/changed/removed since with it we can access to list_node in pipe_manager
* maybe doing the `list_node` as a map so we don't have to iterate to access it, but use only the id as the key.

wwmm:
>we pass only the node id to on_app_added/changed/removed since with it we can access to list_node in pipe_manager

It is a way. I think that in any of the places where we connect to these signals we can iterate over list_node. Unless there is somewhere where we have no access to the pipe_manager instance. But I do not remember having such places in our code.

>maybe doing the list_node as a map so we don't have to iterate to access it, but use only the id as the key.

I think it should be fine. I had to create this list from the very beginning so I could start to interact with PipeWire nodes. I did not even think about using a mat at the time.

wwmm:
But first try to see if `node = nd->nd_info;` fixes the interface update issue. This way we can focus only on improving the infrastructure without this problem in the way.

Digitalone1:
> But first try to see if `node = nd->nd_info;` fixes the interface update issue. This way we can focus only on improving the infrastructure without this problem in the way.

I will try tomorrow.

> I think it should be fine. I had to create this list from the very beginning so I could start to interact with PipeWire nodes. I did not even think about using a map at the time.

Using a map it's better because we access at the node_info as an array not having to write a for cycle everytime. The only issue would be when the item is not found and we should handle an exception while with the for cycle you just exit the loop.

Digitalone1:
Or maybe we can check the existence with [contains](https://en.cppreference.com/w/cpp/container/map/contains), but it's a C++20 feature and I'm afraid it's not working for old compilers.

wwmm:
>Or maybe we can check the existence with contains, but it's a C++20 feature and I'm afraid it's not working for old compilers.

Search for `std::map::contains()` in this page https://en.cppreference.com/w/cpp/compiler_support/20. It was implemented in `gcc 9`. I think we are already forcing gcc 10 so we can use this feature.

vchernin:
> Search for `std::map::contains()` in this page https://en.cppreference.com/w/cpp/compiler_support/20. It was implemented in `gcc 9`. I think we are already forcing gcc 10 so we can use this feature.

Whatever the case is I can easily try building the current Flatpak and seeing what happens.

wwmm:
>Whatever the case is I can easily try building the current Flatpak and seeing what happens.

That would help :-). But I think there is a big chance we can use feature. I doubt people with gcc < 10 can compile EasyEffects.

Digitalone1:
> Search for `std::map::contains()` in this page https://en.cppreference.com/w/cpp/compiler_support/20. It was implemented in `gcc 9`. I think we are already forcing gcc 10 so we can use this feature.

Thank you. Anyway if we will use maps, I think I will pick `find` because it return an iterator avoiding to scan the map a second time to access the needed data.

Digitalone1:
@wwmm I resolved the update issue of app info in service mode. When everything is completed, I will make a new PR.

PR #1132, opened 2021-09-05
Nodes tracked in map rather than vector

### LINKED ISSUE #1124: Issues with global bypass, gain scales and levelbars
I'm working on the code and discovered other issues.

- [x] Global bypass is not followed when a plugin is added to the list or their order is changed. The button stays toggled even if the plugins are operating.
- [x] Input and output gains of Autogain and Maximizer are not working. I'm shocked that I only found out this now.

Some levelbar have issues:  Detected in Deesser and Armonics in Bass Enhancer and Exciter. The level goes out of the range on certain parameters. Maybe also the one in Stereo Tools has to be checked, it stays to 0 all the time. Don't know if It's broken or I'm not able to find a combination of parameters to make it move.  

@wwmm since I made lots of changes, it's better if the fixes are applied on top on my commits. I have not finished yet because I'm trying to improve the blocklist management that still fails in some situations.  So if you have any hint, let me know how I can fix the issues above. Then I'll apply the merge request which surely will be the biggest I ever made.

Digitalone1:
* the vector of nodes has been replaced with a map indexed by node id
* resolved the issue related to app info ui: now the correct info are show when the window is opened in service mode
* on_node_info a new state of the NodeInfo is updated inside the node map 
* now on stream added/changed/removed we only pass node id, not the whole NodeInfo struct; with the id we can scan the node map to get the updated NodeInfo data
* some functions have been partially rewritten the pass only the relevant data and not the whole NodeInfo
* string_to_float function removed: it was needed only for eq APO import; it has been replaced with std:stof    

edit: Tested, it's working to me. Please @wwmm, test the input pipeline.

edit2: Fixes #1124

### DIFF
--- a/data/schemas/com.github.wwmm.easyeffects.autogain.gschema.xml
+++ b/data/schemas/com.github.wwmm.easyeffects.autogain.gschema.xml
@@ -1,9 +1,17 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <schemalist>
     <schema id="com.github.wwmm.easyeffects.autogain">
+        <key name="input-gain" type="d">
+            <range min="-36" max="36" />
+            <default>0</default>
+        </key>
+        <key name="output-gain" type="d">
+            <range min="-36" max="36" />
+            <default>0</default>
+        </key>
         <key name="target" type="d">
-            <range min="-100.0" max="0.0" />
-            <default>-23.0</default>
+            <range min="-100" max="0" />
+            <default>-23</default>
         </key>
         <key name="reset-history" type="b">
             <default>false</default>
--- a/data/schemas/com.github.wwmm.easyeffects.bassenhancer.gschema.xml
+++ b/data/schemas/com.github.wwmm.easyeffects.bassenhancer.gschema.xml
@@ -2,15 +2,15 @@
 <schemalist>
     <schema id="com.github.wwmm.easyeffects.bassenhancer">
         <key name="input-gain" type="d">
-            <range min="-36.1" max="36.1" />
-            <default>0.0</default>
+            <range min="-36" max="36" />
+            <default>0</default>
         </key>
         <key name="output-gain" type="d">
-            <range min="-36.1" max="36.1" />
-            <default>0.0</default>
+            <range min="-36" max="36" />
+            <default>0</default>
         </key>
         <key name="amount" type="d">
-            <range min="-100" max="36.1" />
+            <range min="-100" max="36" />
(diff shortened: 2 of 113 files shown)


Digitalone1:
* now input and output gain scales work also for autogain and maximizer

wwmm:
What editor do you use to write the code? We have some profiles for clang-tidy / clang-formatter already in the repository that I usually use in vscode. As I am seeing ti doing suggestions like replacing `toggle_listen_mic->set_visible((child_name == "page_players") ? false : true);` by just `toggle_listen_mic->set_visible(child_name != "page_players");` I wonder if you are using an editor that is not benefiting from the clang profiles we already have.

wwmm:
I have customized our clang-tidy profile over time because sometimes it suggests things that are not practical. And I still think that there still are some things here and there that I will eventually disable. But putting these minor things aside it is usually worth to take look at them.

Digitalone1:
I use [atom](https://archlinux.org/packages/community/x86_64/atom/)

Don't know if there's something I can do to use these profiles inside it.

wwmm:
I used to use Atom. At least for clang-format there are definitely plugins. But there are probably plugins for clang tidy too. Usually all the plugins have to do is reading these files
https://github.com/wwmm/easyeffects/blob/master/.clang-tidy
https://github.com/wwmm/easyeffects/blob/master/.clang-format

Hopefully vscode and atoms plugins won't require them to be at different paths.

Digitalone1:
I'll take a look at it

wwmm:
Something I use on vscode that is really good but I do not know if there is an alternative on Atom is this https://marketplace.visualstudio.com/items?itemName=llvm-vs-code-extensions.vscode-clangd

wwmm:
A long time ago I also added to repository a vscode profile https://github.com/wwmm/easyeffects/blob/master/easyeffects.code-workspace. clang language server is nice but its default behavior of automatically adding headers has some questionable results from time to time. So that feature I disabled.

For the the gsettings xml I have been using a Red Hat plugins for auto formatting https://marketplace.visualstudio.com/items?itemName=redhat.vscode-xml. I have no idea if they have a release for Atom.

Digitalone1:
For lots of substitutions I use folder search and replace. With regular expressions it's very powerful, I'm pretty good at writing them.

wwmm:
Sometimes I use vscode search and replace. But so far I have been lucky and no regex was needed. They are not something I am good at XD

PR #1133, opened 2021-09-06
Add new plugins before a limiter, if present

Digitalone1:
I added a new security feature which is quite helpful.

Adding a new plugin to the stack we have to be careful because sometimes the output level could be too high and that may damage the audio device.

That's because it's recommended to use a plugin of type limiter at the last position of the filter chain (the limiter itself or the maximizer).

If the user don't use one of these plugins in such place we can't do nothing. But if the user has been already careful to put one of them at the final position, we follow this responsible behaviour adding new plugins at the second to last position, just before the limiter.

### DIFF
--- a/include/effects_base_ui.hpp
+++ b/include/effects_base_ui.hpp
@@ -125,7 +125,7 @@ class EffectsBaseUi {
 
   Glib::RefPtr<Gtk::StringList> blocklist, plugins, selected_plugins;
 
-  std::map<std::string, std::string> plugins_names{{plugin_name::autogain, _("Autogain")},
+  std::map<Glib::ustring, Glib::ustring> plugins_names{{plugin_name::autogain, _("Autogain")},
                                                    {plugin_name::bass_enhancer, _("Bass Enhancer")},
                                                    {plugin_name::bass_loudness, _("Bass Loudness")},
                                                    {plugin_name::compressor, _("Compressor")},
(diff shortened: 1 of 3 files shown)


wwmm:
I think I had a reason to compare `translated_name` with `value.c_str()`. It may not be relevant anymore but there were situations where Glib::ustring behaved in an odd way when gtkmm creates it from a null string passed byt gtk. If i am not mistaken sometimes that led to segmentation faults. I think we should try to test this somehow before merging.

wwmm:
>If i am not mistaken sometimes that led to segmentation faults

Led to segmentaion faults when trying to compare it to a Glib::string that was created with a valid c string.

Digitalone1:
`plugins_names` has now both key and value to Glib::ustring.

`value` is now Glib::ustring and can compare directly with `translated_name`. `value` before was a std::string that can't compare to Glib::ustring.

Anyway I tested it, no issues so far.

Digitalone1:
> I think I had a reason to compare `translated_name` with `value.c_str()`. It may not be relevant anymore but there were situations where Glib::ustring behaved in an odd way when gtkmm created it from a null string passed byt gtk. If i am not mistaken sometimes that led to segmentation faults. I think we should try to test this somehow before merging.

I had issues with GTK labels that always expect a Glib::ustring. Passing them a char pointer array caused labels to show weird characters. Better to have Glib::ustring directly or passing a std::string.

But for comparison, there's no way with Glib::ustring and std::string, one of them has to convert to char pointers or Glib::ustring to std::string with raw().

Issue #1047, opened 2021-07-27
Output app list doesn't display icons of Flatpak applications


vchernin:
In the output section, while audio can be manipulated for Flatpak apps, the icon for the app doesn't seem to show. This issue is specifically about other apps being a Flatpak, which affects both the Arch package and the EasyEffects Flatpak. 

Arch package with Chromium Flatpak installed:
![image](https://user-images.githubusercontent.com/38842733/127086556-63184a79-3bfd-4e1d-8192-ac72b8328be0.png)

Flatpak Package with Chromium Flatpak:
![image](https://user-images.githubusercontent.com/38842733/127086647-ea7a10d5-8173-40d6-8be1-0eb6c7a5ea04.png)

Tested on Arch with KDE, also occurs on GNOME.

wwmm:
This is probably a gtk issue when running in flatpak. After getting the app icon name from PipeWire we just set this name in the image widget. gtk is the one actually finding the icon and loading it. Flatpak sandbox must be breaking whatever gtk is doing to load these icons.

wwmm:
For example under normal circunstances Chromium icon can be found in a path similar to `/usr/share/icons/hicolor/256x256/apps/`. gtk finds icons in standard places on its own. But the flatpak package probably has them somewhere else. If gtk is supposed to be able to find them in the flatpak sandbox I do not know.

vchernin:
The equivalent directory for Flatpaks are here. Note you shouldn't need to look inside the sandbox.

`/var/lib/flatpak/exports/share/icons/hicolor` (system wide installed Flatpaks) and `~/.local/share/flatpak/exports/share/icons/hicolor` (per user installed Flatpaks).

In each directory you will find a bunch of directories something like:
```
128x128  24x24    32x32  512x512  96x96             index.theme
16x16    256x256  48x48  64x64    icon-theme.cache  scalable
```
And similar to the traditional way each icon size has an `apps` directory within it, and that's where the icons are.
So for 256x256 icons:
`/var/lib/flatpak/exports/share/icons/hicolor/256x256/apps` and `~/.local/share/flatpak/exports/share/icons/hicolor/256x256/apps`

Since these are accesible without looking into the sandbox, it seems the issue is gtk just isn't aware where to look?

Further reference:
To install a Flatpak system wide: `flatpak install com.github.wwmm.easyeffects` 
To install a Flatpak per user: `flatpak install --user com.github.wwmm.easyeffects`

[...]

PR #1134, opened 2021-09-07
Optimize app icon management

Digitalone1:
Related to #1047. The explanation for this is [here](https://github.com/wwmm/easyeffects/issues/1047#issuecomment-914156327).


wwmm:
Based on my experience people will complain about the `util::warning` flooding their logs. I will replace it by `util::debug`

Digitalone1:
It's only updated once, not everytime the lambda is called.

wwmm:
Ops. It is true.  While I was reading the code here on github I've got the impression this was somewhere else.

PR #1136, opened 2021-09-08
Icons fixed and changelog updated

### LINKED ISSUE #1047: Output app list doesn't display icons of Flatpak applications

### LINKED ISSUE #1050: Plugins "puzzle piece" icon not visible on KDE
I had some issues with icons and text not displaying while testing the Flatpak on KDE.
The majority of these were solved for the Flatpak by installing `xdg-desktop-portal-gtk` on the host system.

However one issue seems to remain for both the Arch package and the Flatpak:

Arch package:
image

Flatpak:
image

The plugin "puzzle piece" icon isn't visible in the bottom center. This doesn't occur for me on GNOME:
image

Sidenote about theming in general:
Do you prefer the app to follow the Adwaita theme everywhere by default? Currently there doesn't seem to be much support for GTK 4 themes integrated in Flatpak yet, so by default on KDE the Flatpak will look as shown.

However, by workarounding with `sudo flatpak override --filesystem=~/.themes` all Flatpaks are given access to the system themes, so the [Flatpak looks identical to the Arch package](https://user-images.githubusercontent.com/38842733/127287795-141e53d7-85b5-4a51-975b-97d796a89cd4.png). This particular puzzle piece icon issue occurs either way though.

Edit: somehow I took all these screenshots without realizing there's a black bar where actual info should be. I guess that's a bug too. I'll try to figure that out later. Edit: that's just because I haven't started playing audio yet.

Digitalone1:
Fixes #1047 
Fixes #1050 

There are also some optimizations and text changes coming from clang-format.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,13 @@
 
 ### Added
 
-- Removing an application from the blocklist, its previous enabled state is restored.
 - When trying to add an autoloading profile for a device already in the list its target preset will be updated. This
   way we can change the profile preset without having to remove and recreating it.
 - The preset autoloading support implementation was redesigned again. It should work on more hardware now. For more
   information see issue [1051](https://github.com/wwmm/easyeffects/issues/1051).
+- If the Limiter or the Maximizer are set in the last position of the plugin stack, new plugins are added at the
+  second to last position in order to prevent hardware damage on eventually high output level.
+- Removing an application from the blocklist, its previous enabled state is restored.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 - Now the blocklist is correctly set when switching presets.
 - Now the status of the global bypass button is correctly updated when changing plugin stack.
 - Missing icons on the system should not be shown inside the application info UI.
+- Some icons not showing in Plasma DE with Breeze icon theme should appear now.
 
 ## [6.1.0]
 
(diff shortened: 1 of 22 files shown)


wwmm:
It is not related to this pull request but at some point we introduced a regression in the code that switches output devices when the user changes the default. I will take a look. I am doing tests now with my second sound card and EasyEffects is not switching automatically to the new default device after a new selection is done in pavucontrol. Strange...

Digitalone1:
I can try to test with the hdmi port on my laptop. Tell me where the switch is done in the code and I'll take a look tomorrow.

wwmm:
I've found the source of the regression. And it is really subtle. In the callback of the signal `settings->signal_changed("output-device").connect` the following line
```
const auto& name = settings->get_string(key).raw();
```
has to be replaced by
```
const auto name = settings->get_string(key).raw();
```
I think that `auto&` may have created a pointer to a pointer instead of passing by reference. The output of the `raw()` method is already a `const std::string&`.

wwmm:
Master branch updated with the fix.

Digitalone1:
So if I explicitly need a reference to a std::string, I have to construct a new string from raw().

Later I'll give a check to other `raw()` inside the code.

PR #1137, opened 2021-09-09
Fix lv2 port values, icon theme and autoload preset dropdown

Digitalone1:
This should resolve [this](https://github.com/wwmm/easyeffects/pull/1127#issuecomment-911951540).

We make a single copy and, since a static member is used, the lambda will not capture it avoiding another copy. @wwmm correct my if I'm wrong.

They're storage duration is tied to the application, so it should be extremely safe.

### DIFF
--- a/include/application_ui.hpp
+++ b/include/application_ui.hpp
@@ -64,6 +64,8 @@ class ApplicationUi : public Gtk::ApplicationWindow {
   int soe_latency = 0, sie_latency = 0;
 
   static void apply_css_style(const std::string& css_file_name);
+
+  auto setup_icon_theme() -> Glib::RefPtr<Gtk::IconTheme>;
 };
 
 #endif
--- a/include/autogain.hpp
+++ b/include/autogain.hpp
@@ -42,13 +42,14 @@ class AutoGain : public PluginBase {
                std::span<float>& left_out,
                std::span<float>& right_out) override;
 
-  sigc::signal<void(const double&,              // loudness
-                    const double&,              // gain
-                    const double&,              // momentary
-                    const double&,              // shortterm
-                    const double&,              // integrated
-                    const double&,              // relative
-                    const double&)> results;    // range
+  sigc::signal<void(const double&,  // loudness
+                    const double&,  // gain
+                    const double&,  // momentary
+                    const double&,  // shortterm
+                    const double&,  // integrated
+                    const double&,  // relative
+                    const double&)>
+      results;  // range
 
  private:
   bool ebur128_ready = false;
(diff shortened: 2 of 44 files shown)


wwmm:
>We make a single copy and, since a static member is used, the lambda will not capture it avoiding another copy. @wwmm correct my if I'm wrong.

There won't be a copy. But I do not think it is because a static declaration was used. The lambda capturing process works a little different with member variables. `[=, this]` passes the class `this` pointer by value. All the member variables are accessed through the class pointer. So they won't be copied even if the static keyword is not used.

>They're storage duration is tied to the application, so it should be extremely safe.

The static keyword does more than that and I do not think we want the rest in this situation https://en.cppreference.com/w/cpp/language/static
```
Static data members are not associated with any object. They exist even if no objects of the class have been defined. There is only one instance of the static data member in the entire program with static storage duration, unless the keyword thread_local is used, in which case there is one such object per thread with thread storage duration (since C++11). 
```
At this moment we do not support multiple instances of a plugin. So I would not expect bad things to happen. But if we try to do that in the future having these members as static may be a problem if all the class instances use the same static member variable. At least at first sight it seems to me that this is what will happen.

As we don't really need them to be static it is probably better not doing so. Other than that I do not see problems. Performance will probably be the same. But what is happening may be easier to understand as it is coded in the pull request. And like before the variable value may be overwritten before use. What isn't really a problem in the level meter case.

Digitalone1:
So, it's better if they're simple non-static class member. Since plugins classes instance live alongside plugin_ui instances, there's no issue.

Digitalone1:
Forced a push update, now all ports are non-static members.

Digitalone1:
I saw you already constructed an icon them object adding a resource path in application_ui

```
add_resource_path("/com/github/wwmm/easyeffects/icons")
```

So I thought there's no need to make other two same objects in soe_ui and sie_ui, just pass the previous in initialization list.

Digitalone1:
The autoload preset dropdown is sorted with a GTK sorter and initially its position is the same as input/output preset string lists, but when a new preset is added, it is positioned at the last position while the sorter makes another order, so the autoload selection is wrong in some places.

To fix this we do not pick the position id, but the selected label corresponding to the preset filename.

wwmm:
>To fix this we do not pick the position id, but the selected label corresponding to the preset filename.

I see. I did not do mane tests while adding presets new presets in the presets menu.

I did some tests int this pull request branch and it seems fine.

When optimizing the signals take care to not pass by reference things like the spectrum vector. If PipeWire changes latency this vector size changes and the code on the other side may read or write out of bounds. As far as I remember only the spectrum plugin needs vectors in its signals. Some plugins like the multiband compressor have std::array. But in this case the size does not change.

Issue #535, opened 2019-06-24
Optional support for new xdg background portal


AsavarTzeth:
Thanks to a new portal interface ([org.freedesktop.impl.portal.Background](https://github.com/flatpak/xdg-desktop-portal/blob/master/data/org.freedesktop.impl.portal.Background.xml)) it should now be possible to solve this old issue: https://github.com/flathub/com.github.wwmm.pulseeffects/issues/18

In time this should ideally be used everywhere, whenever possible. It does however add a dependency on `xdg-desktop-portal >= 1.4.0`, which means distributions providing older versions is the primary reason it can't always be relied upon at this time.

Would it be possible to check for the availability of this DBus API at runtime and if available use that instead of writing `pulseeffects-service.desktop` directly?

I am no expert but if you need any more information or assistance I will try my best to provide it.

wwmm:
At least at compile time it is possible to check if it is available as they use pkg-config. I have no idea about runtime. And I also have no idea about how to use it. I think I will need to see a code example. The dbus service file we just have to put in the right place. DBUS and GTK take care of the rest on their own.

wwmm:
Wait... The DBUS service file is not related to the autostart. It only starts PE background service when the user clicks on PE lunch icon. The autostart switch in PE menu does not use the DBUS service file. It creates the autostart file directly as shown here https://github.com/wwmm/pulseeffects/blob/8e95621d6e11a0376b04e12beed7402e942c5db9/src/general_settings_ui.cpp#L140

wwmm:
Forget the previsou messages. I was thinking about another service file XD hahaha

wwmm:
These are the places where changes have to be made https://github.com/wwmm/pulseeffects/blob/8e95621d6e11a0376b04e12beed7402e942c5db9/src/general_settings_ui.cpp#L140 and https://github.com/wwmm/pulseeffects/blob/8e95621d6e11a0376b04e12beed7402e942c5db9/src/general_settings_ui.cpp#L124. When we find out how to use this portal it is only a matter of using a  preprocessor `#ifdef` based on the availability of `xdg-desktop-portal >= 1.4.0` at compile time. I imagine that the flatpak package is the only one that needs to use this portal. So we could have a build time option that you could use to make meson compile PE with this portal.

AsavarTzeth:
[...]

PR #1141, opened 2021-09-10
Integrate libportal for autostart

Digitalone1:
Related to #535 

@vchernin please test these changes. It's compiling and the request is completed successfully, but I can't test if it's autostarting on Flatpak since I don't use it.

@wwmm I wonder if this is working only for Flatpak or we can use it also for the default app, but I dont' know how to check the initial state since we only look if the desktop entry is present in the config folder.

### DIFF
--- a/include/autogain.hpp
+++ b/include/autogain.hpp
@@ -42,13 +42,14 @@ class AutoGain : public PluginBase {
                std::span<float>& left_out,
                std::span<float>& right_out) override;
 
-  sigc::signal<void(const double&,              // loudness
-                    const double&,              // gain
-                    const double&,              // momentary
-                    const double&,              // shortterm
-                    const double&,              // integrated
-                    const double&,              // relative
-                    const double&)> results;    // range
+  sigc::signal<void(const double&,  // loudness
+                    const double&,  // gain
+                    const double&,  // momentary
+                    const double&,  // shortterm
+                    const double&,  // integrated
+                    const double&,  // relative
+                    const double&)>
+      results;  // range
 
  private:
   bool ebur128_ready = false;
--- a/include/presets_manager.hpp
+++ b/include/presets_manager.hpp
@@ -75,14 +75,16 @@ class PresetsManager {
 
   void save_preset_file(const PresetType& preset_type, const Glib::ustring& name);
 
-  void write_plugins_preset(const PresetType& preset_type, const std::vector<Glib::ustring>& plugins,
+  void write_plugins_preset(const PresetType& preset_type,
+                            const std::vector<Glib::ustring>& plugins,
                             nlohmann::json& json);
 
   void remove(const PresetType& preset_type, const Glib::ustring& name);
 
   void load_preset_file(const PresetType& preset_type, const Glib::ustring& name);
 
-  void read_plugins_preset(const PresetType& preset_type, const std::vector<Glib::ustring>& plugins,
+  void read_plugins_preset(const PresetType& preset_type,
+                           const std::vector<Glib::ustring>& plugins,
                            const nlohmann::json& json);
 
   void import(const PresetType& preset_type, const std::string& file_path);
--- a/include/scale_helper.hpp
+++ b/include/scale_helper.hpp
@@ -24,8 +24,8 @@
 
 inline auto prepare_scale(Gtk::Scale* scale, const std::string& unit) {
   scale->set_format_value_func([=](const auto& value) {
-    const auto& v_str = Glib::ustring::format(std::setprecision(scale->get_digits()), std::fixed,
-                                              scale->get_adjustment()->get_value());
+    const auto& v_str =
+        Glib::ustring::format(std::setprecision(scale->get_digits()), std::fixed, scale->get_adjustment()->get_value());
 
     return v_str + ((unit.empty()) ? "" : (" " + unit));
   });
--- a/include/spinbutton_helper.hpp
+++ b/include/spinbutton_helper.hpp
@@ -24,8 +24,8 @@
 #include <sstream>
 
 inline auto parse_spinbutton_output(Gtk::SpinButton* button, const Glib::ustring& unit) -> bool {
-  const auto& value = Glib::ustring::format(std::setprecision(button->get_digits()), std::fixed,
-                                            button->get_adjustment()->get_value());
+  const auto& value =
+      Glib::ustring::format(std::setprecision(button->get_digits()), std::fixed, button->get_adjustment()->get_value());
 
   button->set_text(value + ((unit.empty()) ? "" : (" " + unit)));
 
--- a/src/fir_filter_base.cpp
+++ b/src/fir_filter_base.cpp
@@ -114,9 +114,9 @@ auto FirFilterBase::create_lowpass_kernel(const float& cutoff, const float& tran
       Blackman window https://www.dspguide.com/ch16/1.htm
     */
 
-    const float w =
-        0.42F - 0.5F * std::cos(2.0F * std::numbers::pi_v<float> * static_cast<float>(n) / static_cast<float>(M)) +
-        0.08F * std::cos(4.0F * std::numbers::pi_v<float> * static_cast<float>(n) / static_cast<float>(M));
+    const float w = 0.42F -
+                    0.5F * std::cos(2.0F * std::numbers::pi_v<float> * static_cast<float>(n) / static_cast<float>(M)) +
+                    0.08F * std::cos(4.0F * std::numbers::pi_v<float> * static_cast<float>(n) / static_cast<float>(M));
 
     output[n] *= w;
 
--- a/src/pipe_info_ui.cpp
+++ b/src/pipe_info_ui.cpp
@@ -229,10 +229,10 @@ PipeInfoUi::PipeInfoUi(BaseObjectType* cobject,
       }
     }
 
-    const auto& id = dropdown_autoloading_output_presets->get_selected();
+    const auto preset_name =
+        dropdown_autoloading_output_presets->get_selected_item()->get_property<Glib::ustring>("string").raw();
 
-    presets_manager->add_autoload(PresetType::output, output_presets_string_list->get_string(id).raw(),
-                                  holder->info.name, device_profile);
+    presets_manager->add_autoload(PresetType::output, preset_name, holder->info.name, device_profile);
   });
 
   autoloading_add_input_profile->signal_clicked().connect([=, this]() {
@@ -264,10 +264,10 @@ PipeInfoUi::PipeInfoUi(BaseObjectType* cobject,
       }
     }
 
-    const auto& id = dropdown_autoloading_input_presets->get_selected();
+    const auto preset_name =
+        dropdown_autoloading_input_presets->get_selected_item()->get_property<Glib::ustring>("string").raw();
 
-    presets_manager->add_autoload(PresetType::input, input_presets_string_list->get_string(id).raw(), holder->info.name,
-                                  device_profile);
+    presets_manager->add_autoload(PresetType::input, preset_name, holder->info.name, device_profile);
   });
 
   spinbutton_test_signal_frequency->signal_output().connect(
--- a/src/plot.cpp
+++ b/src/plot.cpp
@@ -211,9 +211,9 @@ void Plot::on_draw(const Cairo::RefPtr<Cairo::Context>& ctx, const int& width, c
     }
 
     if (controller_motion->contains_pointer()) {
-      const auto& msg = "x = " + Glib::ustring::format(std::setprecision(n_x_decimals), std::fixed, mouse_x) +
-                        " " + x_unit + "  y = " +
-                        Glib::ustring::format(std::setprecision(n_y_decimals), std::fixed, mouse_y) + " " + y_unit;
+      const auto& msg =
+          "x = " + Glib::ustring::format(std::setprecision(n_x_decimals), std::fixed, mouse_x) + " " + x_unit +
+          "  y = " + Glib::ustring::format(std::setprecision(n_y_decimals), std::fixed, mouse_y) + " " + y_unit;
 
       Pango::FontDescription font;
       font.set_family("Monospace");
--- a/src/presets_manager.cpp
+++ b/src/presets_manager.cpp
@@ -371,7 +371,8 @@ void PresetsManager::save_preset_file(const PresetType& preset_type, const Glib:
   util::debug(log_tag + "saved preset: " + output_file.string());
 }
 
-void PresetsManager::write_plugins_preset(const PresetType& preset_type, const std::vector<Glib::ustring>& plugins,
+void PresetsManager::write_plugins_preset(const PresetType& preset_type,
+                                          const std::vector<Glib::ustring>& plugins,
                                           nlohmann::json& json) {
   for (const auto& name : plugins) {
     if (name == plugin_name::autogain) {
@@ -547,8 +548,9 @@ void PresetsManager::load_preset_file(const PresetType& preset_type, const Glib:
   util::debug(log_tag + "loaded preset: " + input_file.string());
 }
 
-void PresetsManager::read_plugins_preset(const PresetType& preset_type, const std::vector<Glib::ustring>& plugins,
-                         const nlohmann::json& json) {
+void PresetsManager::read_plugins_preset(const PresetType& preset_type,
+                                         const std::vector<Glib::ustring>& plugins,
+                                         const nlohmann::json& json) {
   for (const auto& name : plugins) {
     if (name == plugin_name::autogain) {
       autogain->read(preset_type, json);
--- a/src/spectrum.cpp
+++ b/src/spectrum.cpp
@@ -73,9 +73,8 @@ void Spectrum::process(std::span<float>& left_in,
     if (const uint& k = total_count + n; k < real_input.size()) {
       // https://en.wikipedia.org/wiki/Hann_function
 
-      const float w =
-          0.5F * (1.0F - std::cos(2.0F * std::numbers::pi_v<float> * static_cast<float>(k) /
-          static_cast<float>(real_input.size() - 1U)));
+      const float w = 0.5F * (1.0F - std::cos(2.0F * std::numbers::pi_v<float> * static_cast<float>(k) /
+                                              static_cast<float>(real_input.size() - 1U)));
 
       real_input[k] = 0.5F * (left_in[n] + right_in[n]) * w;
     } else {
--- a/src/spectrum_ui.cpp
+++ b/src/spectrum_ui.cpp
@@ -30,10 +30,11 @@ SpectrumUi::SpectrumUi(BaseObjectType* cobject, const Glib::RefPtr<Gtk::Builder>
   connections.emplace_back(
       settings->signal_changed("color-axis-labels").connect([&](const auto& key) { init_frequency_labels_color(); }));
 
-  connections.emplace_back(
-      settings->signal_changed("height").connect([&](const auto& key) { set_content_height(settings->get_int("height")); }));
+  connections.emplace_back(settings->signal_changed("height").connect(
+      [&](const auto& key) { set_content_height(settings->get_int("height")); }));
 
-  connections.emplace_back(settings->signal_changed("n-points").connect([&](const auto& key) { init_frequency_axis(); }));
+  connections.emplace_back(
+      settings->signal_changed("n-points").connect([&](const auto& key) { init_frequency_axis(); }));
 
   connections.emplace_back(
       settings->signal_changed("minimum-frequency").connect([&](const auto& key) { init_frequency_axis(); }));
@@ -43,8 +44,8 @@ SpectrumUi::SpectrumUi(BaseObjectType* cobject, const Glib::RefPtr<Gtk::Builder>
 
   connections.emplace_back(settings->signal_changed("type").connect([&](const auto& key) { init_type(); }));
 
-  connections.emplace_back(
-      settings->signal_changed("fill").connect([&](const auto& key) { plot->set_fill_bars(settings->get_boolean(key)); }));
+  connections.emplace_back(settings->signal_changed("fill").connect(
+      [&](const auto& key) { plot->set_fill_bars(settings->get_boolean(key)); }));
 
   connections.emplace_back(settings->signal_changed("show-bar-border").connect([&](const auto& key) {
     plot->set_draw_bar_border(settings->get_boolean(key));


Digitalone1:
Failing here because libportal is missing, compiling on my side.

vchernin:
@Digitalone1 thanks for working on this! 

I haven’t tried compiling it just yet (I will in a moment), but I don’t think all the necessary header files are included?

See https://github.com/wwmm/easyeffects/issues/535#issuecomment-905819257

Digitalone1:
@vchernin I forgot it and didn't notice since it's compiling, anyway they are needed only if using xdg_parent which is nullable and I passed it as nullptr.

If it's not working, try to add it.

vchernin:
So in trying to test this I think I ran into other issues...

On this build the enabling switch in the output section is just grayed out :

This picture is on stable: 
image

This is on this build:
image
Enabling the switch reliably causes easyeffects to crash within a second.
The only log from doing so is the joyful:
`Segmentation fault (core dumped)`

I think that issue is likely a regerssion from some other change in master.

Also I found on Fedora 35 I don't think EasyEffects even stable is working at all... I think this is not EasyEffects problem though. 

Later when I have time I'll try to figure out what's going on so I can actually test this PR.

Digitalone1:
If it's grayed out, it means the application sink is not linked to easyeffects sink.

I can't reproduce on my system. Which version of Pipewire is installed on Fedora? Can you test on Arch?

vchernin:
Ok, so autostart does seem to work. It prompts me as expected, and `~/.config/autostart/com.github.wwmm.easyeffects.desktop` is created.

I do want to test it a bit more extensively as I didn't work the first time, but `flatpak remove easyeffects --delete-data` seemed to help (the flag removes any config stuff so effectively easyeffects is entirely reset). But I might have just done something strange.

I had to rebase to Fedora 34 though. The same version (PipeWire 0.3.35) more or less should be on 34 and 35.  I'll open a new issue for Fedora 35.

Digitalone1:
Anyway if you start with Process All Outputs, the application should be processed anyway, so if it's disabled there should be something preventing its linking to easyeffects sink. How is the spectrum? Do you see it moving?

Glad to see the autostart is working. Maybe it's better to add that missing header anyway. Do you know if libportal can work also for non-flatpak applications? If yes, we might think to use it exclusively rather than using the outputstream to write the desktop entry file.

vchernin:
> Glad to see the autostart is working. Maybe it's better to add that missing header anyway. Do you know if libportal can work also for non-flatpak applications? If yes, we might think to use it exclusively rather than using the outputstream to write the desktop entry file.

libportal and portals in general are supposed to work for non Flatpak apps. The official position is this: https://github.com/flatpak/libportal/issues/33#issuecomment-905779376 But it is possible to run into issues when not using a sandbox. I could swear I saw a non sandboxed app issue about the background portal specifically but now I can’t find it.

I’d suggest just trying it and seeing what happens. You should see a operating prompt when you enable the setting in EasyEffects. If things aren’t working watch what’s going on in the auto start directory. It should create an EasyEffects desktop file for you, and remove it when you disable the auto start feature in EasyEffects.

Digitalone1:
> You should see a operating prompt when you enable the setting in EasyEffects. 
I didn't see nothing. Where do you see exactly this prompt?

> If things aren’t working watch what’s going on in the auto start directory. It should create an EasyEffects desktop file for you, and remove it when you disable the auto start feature in EasyEffects.
On my side the file is created by output string stream function. So to see if libportal is really working, I have to disable this function and leave only libportal. The problem is I don't have any clue to know if the autostart is enabled from libportal. At the moment we check if the file is present, but that check could fail in flatpak.

vchernin:
I realized one cause of inconsistency: once I had agreed to allowing the app to run in background flatpak rememebered taht.
But I can reset that: `flatpak permission-reset com.github.wwmm.easyeffects`

But that won't help non-flatpak.

Here is the prompt:
https://user-images.githubusercontent.com/38842733/132912151-24dad830-c8b2-42bf-8949-aecb127b32a3.mp4

It will occur when you disable or enable that autoswitch button in easyeffects so long as permission is not given per that command above.

Digitalone1:
I don't see that prompt. Maybe because the file is already created. 

Anyway I don't understand. Aren't you able to disable the autostart?

On non-flatpak we just remove the desktop entry in config folder.

The problem is the initialization of the switch in Flatpak. We check the existence of the desktop entry, but if we wouldn't do that, we didn't know how to check through libportal if the autostart is already enabled.

vchernin:
> I don't see that prompt. Maybe because the file is already created.
> 
> Anyway I don't understand. Aren't you able to disable the autostart?

For me enabling the EasyEffects autostart switch instantly creates the file in autostart. Disabling the EasyEffects autostart switch instantly deletes the file in autostart. So yes this seems to work as intended in Flatpak.

> On non-flatpak we just remove the desktop entry in config folder.

As above this seems to be what’s happening with the portal in Flatpak too.
> 
> The problem is the initialization of the switch in Flatpak. We check the existence of the desktop entry, but if we wouldn't do that, we didn't know how to check through libportal if the autostart is already enabled.

This sounds ok to me… do you think this aspect might cause problems? 


I will try to recap the overall situation in Flatpak for clarity:

1. Enabling or disabling the autostart switch triggers the background access portal prompt which is what you see in the video. Only once this permission has been granted by the user can EasyEffects Flatpak do anything to affect autostart.
2. Once EasyEffects has the permission from the user it will keep that permission. The user can enable or disable the EasyEffects autostart switch, which creates or deletes the autostart directory’s desktop file. There will be no additional prompts like shown in the video.
3. If the user decides to revoke the permission given to EasyEffects, EasyEffects should no longer be able to change anything the autostart directory. If the user flips the autostart switch in EasyEffects they will go back to step 1) and see the background access portal prompt again.

Digitalone1:
> This sounds ok to me… do you think this aspect might cause problems?

Yes, if we want to use only libportal and avoid to write a useless file under var flatpak folder. We need to check if autostart is enabled at startup to initialize the switch position. Without checking the file existence, we don't know how to do it through libportal.

We may add a new gsetting key to track the startup option. It's the only solution that came to my mind.

wwmm:
@Digitalone1 is there a reason to define `XdpPortal* portal = nullptr;` outside of the class `GeneralSettingsUi`?  It doesn't look like we will have uses for libportal outside of this particular desktop file configuration.

wwmm:
>We may add a new gsetting key to track the startup option. It's the only solution that came to my mind.

There has to be something better than that. The example libportal maintainer provided does not have anything in this direction?

Digitalone1:
@wwmm because of this recommendation:

https://github.com/flatpak/libportal/blob/40b012ba328c99a0532f4da1f06b173f8ffc26e2/libportal/portal.c#L34

Since general settings ui is constructed and destroyed upon visibility, the only way to let it live alongside the application is to store it in application class.

Digitalone1:
> There has to be something better than that. The example libportal maintainer provided does not have anything in this direction?

I have to check. But tomorrow I'm quite busy, maybe the next week I'll make other tests.

wwmm:
>because of this recommendation: https://github.com/flatpak/libportal/blob/40b012ba328c99a0532f4da1f06b173f8ffc26e2/libportal/portal.c#L34

This kind of recommendation is usually done to avoid people creating multiple objects in the same app. We are not going to use libportal for anything else. And that object is useless in service mode anyway. I do not think it deserves changes in 2 classes constructors. Unless its initialization is a one time operation per app lifetime. What does not seem the case.

wwmm:
>I have to check. But tomorrow I'm quite busy, maybe the next week I'll make other tests.

No problems. This week has been particularly busy for me too. I am not sure I will have much time in the weekend to take a look at the code.

Digitalone1:
Why useless in service mode? 

I can make a new one and delete it in destructor. Or maybe make it static and initialize only if its nullptr.

wwmm:
>Why useless in service mode?

This is one of those functions where the majority of the users will open the window to set if they want the autostart or not. In service mode you will need to open dconf or use a third party tool to try to set this. Very unlikely someone will try to do any of this when you can just open the window and click on the switch.

wwmm:
If there is a strong cause to set this in service mode we will have to init libportal in the main class. But I am not convinced someone will ever need this. Even on headless installations. It is way easier to create the file manually through ssh in this case. And in a headless install you want to configure this once.

wwmm:
>I can make a new one and delete it in destructor. Or maybe make it static and initialize only if its nullptr.

I think it is fine to have the libportal object as a `GeneralSettingsUi` member variable. It could be static if you want. The `GeneralSettingsUi` is not going to have more than one instance running at the same time.

Digitalone1:
Static is better.

For the switch initialization, if libportal is correctly signaling the fail in the callback, I think it's better to track the state through a key in gsettings.

This will be done in less lines of code than now rather than relying on something in libportal which we are not sure it stays the same forever since it's not 100% stable and may change in the future.

Digitalone1:
But this could be a problem if flatpak can revoke permissions when the app is closed. Then we have the gsettings key with an inconsistent value.

I'll check it out.

vchernin:
> if flatpak can revoke permissions when the app is closed. 

Indeed Flatpak seems to be able to revoke permissions at any time from my testing, I don’t know if Flatpak actually explicitly tries to communicate to the app when it does so.

Digitalone1:
I don't see any function to retrieve the state from libportal. Even Gnome ToDo is saving the option in a gsettings key.

So if the permission is revoked when the general settings popover is not shown, the state TRUE of the gsettings key will remake a new request the next time the popover is shown. The use will see the prompt another time. To not show it, the user has to disable the option.

In the next days I'll try to make other tests removing stringstream functions and working only with libportal.

wwmm:
>don't see any function to retrieve the state from libportal. Even Gnome ToDo is saving the option in a gsettings key.

Really?!? That is horrible :-(. I will be honest... I do not like the direction we are being forced to go. Seriously... At this moment we have a simple solution that works in any place that does not put the app in a sandbox. No dependency required for it and if the autostart file is deleted the setting can be easily updated on our side to keep things consistent. But to make this work in flatpak we are forcing a new dependency on everybody and adding more code that we have to maintain. And in the end we have an inferior solution?!? It is ridiculous!

wwmm:
They have to add a way to query this state. Keeping track in gsettings and "hoping for the best" is not good...

wwmm:
I understand that this file won't randomly disappear. If we were in mplementing support for autostart now it would be reasonable starting this way. But replacing a simple solution that requires no dependency by something that feels incomplete is disappointing...

Digitalone1:
I can try to ask on libportal Github. Anyway we are not forced to add it. Or we can use a mixed solution writing the file and making the request through libportal. This would lead to write a useless file in Flatpak, but that's the state of things.

wwmm:
>Or we can use a mixed solution writing the file and making the request through libportal

If we do that we will have even more code to maintain. Either we move this functionality to lib portal or we don't.

Digitalone1:
> If we do that we will have even more code to maintain. Either we move this functionality to lib portal or we don't.

So I'm afraid we can't move to it.

The real problem is that we should know in the code if we are dealing with Flatpak or not. That would make the things easier.

If we had a bool to know it, we would hide the app icon rather than making complicated things with glib icon theme, which is really a bad class.

We could also hide the enable autostart in Flatpak since it's not working. At the user that comes here complaining to add such functionality inside the application we say: look, you are using Flatpak? That option don't work, so you can't do it from the GUI, do it yourself, ask Flatpak to make a utility for starting apps at system login or make a desktop entry in the config folder.

wwmm:
>The real problem is that we should know in the code if we are dealing with Flatpak or not.

I think @vchernin suggested a way to know that. Checking if a specific file exists in the path `/`. Only the Flatpak sandbox would have that file If I remember well.

>We could also hide the enable autostart in Flatpak since it's not working.

I have been thinking about that. I do not like the way libportal is handling this situation. We could hide this switch when running in the Flatpak sandbox and adding to our wiki the contents of an autostart file that will work for Flatpak. And pointing our users to that.

vchernin:
> The real problem is that we should know in the code if we are dealing with Flatpak or not. That would make the things easier.

You can check if you’re in a Flatpak sandbox by checking if `/.flatpak-info` file exists. If that file exists the app can safely assume it is in Flatpak, but otherwise that file shouldn’t be there.

Perhaps libportal could also be an optional build time dependency that only the Flatpak opts in to?

It’s unfortunate this doesn’t quite work as easily as one would hope, without a way to check for the autostart file presence it seems :(

And yeah, ideally portals could replace the traditional way but for now the background portal doesn’t work properly on the host. But that is considered a bug and hopefully will be fixed someday.

I would perhaps ask on libportal github as that seems like a reasonable feature request. Perhaps the #flatpak:matrix.org channel might also have a quick answer, I could ask on your behalf if you’d like.

For what it’s worth the current build seemed to work well for me enough despite the whole issue discussed above. I might try and test that build later to see what happens if the file just disappears.

wwmm:
>You can check if you’re in a Flatpak sandbox by checking if the /.flatpak-info file exists.

Thanks.

>Perhaps libportal could also be an optional dependency that only the Flatpak opts in to?

I considered that. But besides the need for a few `if defined` on our Meson configuration the c++ code will also need some preprocessor code to disable any code using libportal. It can be done. I just do not think it is fair to introduce so much changes on the application side for something that it is ridiculously simple to do outside of Flatpak.

>And yeah, ideally portals could replace the traditional way but for now the background portal doesn’t work properly on the host. But that is considered a bug and hopefully will be fixed someday.

That is my other concern. By introducing one more dependency we are opening the door to more potential bugs. So I would like the solution to at least feel complete to take this kind of risk. Tracking the setting state and hoping that the file is still there feels like a hacking to me and not a proper solution.

Digitalone1:
> I considered that. But besides the need for a few `if defined` on our Meson configuration the c++ code will also need some preprocessor code to disable any code using libportal. It can be done. I just do not think it is fair to introduce so much changes on the application side for something that it is ridiculously simple to do outside of Flatpak.

It would work only for hiding the autostart switch and app icons in flatpak (if we can't solve the issues related to the icon theme object).

Digitalone1:
> I would perhaps ask on libportal github as that seems like a reasonable feature request. Perhaps the #flatpak:matrix.org channel might also have a quick answer, I could ask on your behalf if you’d like.

I already asked and hope that I'm wrong, but if Gnome ToDo is relying on the gsettings key, I'm afraid it's not possible.

Anyway you can ask too, if you want.

vchernin:
So I've made some discoveries that I'm not sure were obvious:

All this assumes the user has given EasyEffects permission to manage autostart via the portal prompt.

If I delete the autostart file manually it is gone. But upon the next launch of EasyEffects with libportal it returns. This assumes the switch in EasyEffects for autostart is still on.

So in the case where the user decides to delete the file without disabling the actual switch EasyEffects will override the user's deletion. Upon next launch EasyEffects will recreate the file. In my mind this makes sense since the user should be using the switch and not something else.

I was curious what would happen for users who had previously manually added an autostart file, but then updated to a build with the autostart feature possible through the libportal.

If the user didn't enable the autostart switch in easyeffects but added the autostart file manually. In my case this was done with the background permission already given.

1. Install stable Flatpak build (ensuring all cache was removed before with `--delete-data`)
2. Manually add autostart file
3. See that autostart works by logging out and in. Now install the libportal build without cleaning configurations.
4. Step 3 is effectively the same process as a EasyEffects Flatpak update would be.
5. Upon launch of EasyEffects the autostart file was removed for me. This is only because the switch was off but the portal permission was given. If the portal permission was not given the file would have been untouched.

If the user *did* enable the autostart switch in easyeffects but added the autostart file manualy, you'll see the user given file will be overwritten by easyeffects. The file content suggested by [the workaround](https://github.com/flathub/com.github.wwmm.easyeffects/issues/3#issuecomment-908632842) and what's done by libportal is slightly different.

So the takeaway here seems to be, with the portal permission the portal it seems will automatically handle cases where the user removes, adds, or modifies the autostart file.

What do you two think about this? Does this eliminate the need for a gsettings key? At least in my mind as a Flatpak user this behaviour seems sensible enough.

When the portal permission is not given, then EasyEffects shouldn't be able to delete or modify anything, but there's no real way of solving that.

Digitalone1:
This is because I added the request everytime the user interface is shown. That is not a bad news. 

If we follow the gsettings key and make the request everytime the user interface is loaded, the state is always updated.

But this forces us to have the portal always living inside the application, not only when the general settings popover is shown. That's not a bug issue. @wwmm what do you think?

wwmm:
>But this forces us to have the portal always living inside the application

Why?

Digitalone1:
> Why?

For different reasons. If we should track the state through a gsettings key, we should request the autostarting everytime the application is shown and maybe also when the application is started, both in window and service mode.

This is to prevent the case when the desktop entry is deleted and we have an inconsistency beetween the gsettings key and the autoload state. Thanks to @vchernin tests we know that if the entry is deleted, it's recreated when a new request is done (I suppose that if the file is created and we make the request on false key by our side, it's deleted).

So rather than asking to libportal the enabled state (which is not possible because this feature is not available), we update the enabled state making a request based on the value of the gsettings bool key.

I think it's better to make these requests at the application startup and everytime the window is shown. Maybe also at the application shutdown. To follow this purpose and considering that libportal [recommends](https://github.com/flatpak/libportal/blob/40b012ba328c99a0532f4da1f06b173f8ffc26e2/libportal/portal.c#L34) to have a single instance throughout application lifetime, we should have the portal object living till the end without deleting it.

wwmm:
>I think it's better to make these requests at the application startup and everytime the window is shown. Maybe also at the application shutdown.

I see what you are thinking but it is only going to be useful when the window is opened. If the autostart file is deleted there is no point in calling libportal in the app initialization because the app won't autostart without the file. In the end the user will start the app manually and the window will be shown anyway. Similar thing on the app shutdown.

As I am not convinced libportal is not going to bite us in the ass in the future I do not want it spread through our code. It will make it harder to remove it in the future if needed. And it will also make it harder to make it an optional dependency. In order for our code to compile without libportal we will have to wrap its usage by C++ preprocessor directives so that these sections of the code can be ignored by the compiler. And doing that in many different places  will force us to spread preprocessor code as well.

I think we can go ahead with the libportal usage. But as much as I dislike having more code to maintain I think it is better to contain the libportal usage for Flatpak. In the Meson files we can create a compilation option that can be used to define a variable that will be used by the C++ preprocessor. When building the Flatpak package this Meson option will be used to enable libportal.

Digitalone1:
You decide and you have the final word.

But I invite you to consider that a good implementation of libportal can make the code smaller than the actual one. If you don't want to make the request at application startup, we can do it only when the window is showed. And since the general settings ui is initialized by application ui, we can put everything in general settings ui. 

The portal object can be static, it's initialized the first time only when it's null, the request is made only in the constructor and on switch changes.

I think implementing this and removing the string stream and the filestystem calls to check and write the file will make the code shorter.

Make your choice.

wwmm:
>I think implementing this and removing the string stream and the filestystem calls to check and write the file will make the code shorter.

That is a good point and the reason why my original intention was replace the current code by the one with libportal. I just lost the confidence that we can rely on this library on the long term.

>You decide and you have the final word.

Let's do a little of both. We can replace the current code by libportal. But let's keep its usage contained in the `GeneralSettingsUi` and see how well things go.

Digitalone1:
I have not good news.

Tested on my system. Libportal is adding the desktop file, but this file has empty name, it's saved as `.desktop`, so the previous saved file is not deleted. This is not consistent.

Then, enabling and disabling the option, I see that the service is only adding the `.desktop` file, but it's not deleting it on gsettings false key. So basically it works only on enable, can't be disabled.

I tried also to add the gtk4 header, but it's giving me an error on gpointer conversion. I think it's not happening in C, so maybe it's not meant to be used with C++.

With that said, I definitely think this is not ready to be adopted.

Better option is to use the `/.flatpak-info` hack to hide the autostart option to Flatpak users so they don't get a non-working option and a file saved in a useless place. If they want, they can still add a custom desktop entry in autostart folder to start easyeffects at system startup

vchernin:
You're testing the Arch build right? If I understand then this PR is blocked on the background portal needing to work on the host? I still can't find an actual issue for that even if I've heard it doesn't work, I can open one then.

Digitalone1:
> If I understand then this PR is blocked on the background portal needing to work on non-sandboxed builds?

Yes, it not fully working as I stated above.

When you disable the switch, the desktop entry is deleted? 

And what is the name of the desktop entry saved in autostart config folder?

vchernin:
> When you disable the switch, the desktop entry is deleted?

Yes and instantly. You can flip the switch many times and it will create or delete the file exactly corresponding to EasyEffects' switch. Of course that assumes EasyEffects has the portal permission to add or delete the autostart file, if it didn't have permission it would ask.

> And what is the name of the desktop entry saved in autostart config folder?

The autostart desktop file is always named `com.github.wwmm.easyeffects.desktop`. It's always in the usual directory `~/.config/autostart`.

vchernin:
If it's interesting the autostart desktop file contents provide by the portal are:

```
[Desktop Entry]
Type=Application
Name=com.github.wwmm.easyeffects
Exec=flatpak run --command=easyeffects com.github.wwmm.easyeffects --gapplication-service
X-Flatpak=com.github.wwmm.easyeffects
```

Digitalone1:
So this means the service only works for sandboxed environments because its not deleting the file on switch disabled and the file is saved as `.desktop`

Tomorrow I'll make a force push to show the last changes I made.

Digitalone1:
> If it's interesting the autostart desktop file contents provide by the portal are:
> 
> ```
> [Desktop Entry]
> Type=Application
> Name=com.github.wwmm.easyeffects
> Exec=flatpak run --command=easyeffects com.github.wwmm.easyeffects --gapplication-service
> X-Flatpak=com.github.wwmm.easyeffects
> ```
To me it was the plain command `easyeffects --gapplication-service`

And the last key X-Flatpak had an empty value.

wwmm:
It is working only in Flatpak we can consider wrapping the libportal calls in preprocessor directives so that this part of the code is only compiled if a Meson build option is given. This way it could still be used in the Flatpak build. But outside of flatpak the current code would be used. Preprocessor directives make the code ugly... But it would be only way to use libportal only where it is actually needed...

Digitalone1:
The problem is we need a gsettings key to track the state through different sessions and we can't use directives for schema files. Otherwise we need to add a new gsettings for plain build even if it's useless since we check the file existence.

wwmm:
>The problem is we need a gsettings key to track the state through different sessions and we can't use directives for schema files. Otherwise we need to add a new gsettings key for plain build even if it's useless since we check the file existence.

I see... I forgot about that...

Digitalone1:
I forgot to commit and force pushed the changes, so the PR has been automatically closed. 

Anyway the new commit is here:

https://github.com/Digitalone1/easyeffects/commit/cea50ac40361333ce40ad72216e7d804806ada88

@vchernin Test it on your system if you want.

I could remake the PR, but I think it's useless, this won't be accepted at this state.

wwmm:
It crossed my mind to make Meson to create a new schema file based on whether or or not the compilation flag for libportal was given. We do something like this in other files but I am not 100% sure the same can be done in the schema xml because it is a more complex file. And developing the libportal integration without a working schema file in the source code will be painful. That Meson solution to generate the file happens in the compilation phase.

Digitalone1:
If you really want to do something with compilation flags, then don't touch anything in gsettings schema and save the autostart state in a custom file under the easyeffects config folder. Integrating with my changes, it can be done.

If it's a very hard work, we can still know if the app is running in sandbox and hide the autostart switch, Flatpak users can autostart it making the file manually.

wwmm:
Honestly I think it is easier to hide the switch for flatpak users.

vchernin:
> The problem is we need a gsettings key to track the state through different sessions and we can't use directives for schema files. Otherwise we need to add a new gsettings key for plain build even if it's useless since we check the file existence.

I may have missed something, but the initial version of the patch from 3 days ago did not have a gsettings key correct? At least from a user's perspective I could not find anything clearly wrong with it's behaviour.

I did find this which seems to explain it:

>> This sounds ok to me… do you think this aspect might cause problems?
> 
> Yes, if we want to use only libportal and avoid to write a useless file under var flatpak folder. We need to check if autostart is enabled at startup to initialize the switch position. Without checking the file existence, we don't know how to do it through libportal.
> 
> We may add a new gsetting key to track the startup option. It's the only solution that came to my mind.

So in another words the older code wrote a useless autostart file, and used it to keep track of the autostart state instead of using a gsettings key?

If the older libportal code worked reasonably well in Flatpak without a gsettings key, couldn't that then be wrapped in preprocessor directives? But obviously writing a useless file is far from ideal... but I don't have any better ideas.

wwmm:
>If the older libportal code worked reasonably well in Flatpak without a gsettings key

The problem is that the position of the corresponding switch in our settings menu won't be synced to the existence or not of the autostart file. So it may be in the off position when the file actually exists. As libportal does not offer a way to query if the file is there the only way to make the switch show something that makes sense is saving the state in a gsettings key. That is why `Gnome ToDo` does that.

>couldn't that then be wrapped in preprocessor directives?

These directives can only be done in the C++ code. But the gsettings schema are define in xml files that are not used by gcc.

vchernin:
> The problem is that the position of the corresponding switch in our settings menu won't be synchronized to the existence or not of the autostart file. So it may be in the off position when the file actually exists. 

I [noted above](https://github.com/wwmm/easyeffects/pull/1141#issuecomment-917451313) that since EasyEffects talks to the portal every time it starts, this shouldn't really be a big issue. If the user for some reason creates or deletes the file unexpectedly it will be taken care of correctly when EasyEffects is started. In my book what EasyEffects does as described above is sensible.

> These directives can only be done in the C++ code. But the gsettings schema are define in xml files that are not used by gcc.

My idea to just do something similar to the original implementation and just not have a gsettings key, even if that's not ideal since a useless autostart file will be created.

Digitalone1:
> So in another words the older code wrote a useless autostart file, and used it to keep track of the autostart state instead of using a gsettings key?

Yes, it was tracking the state on the `easyeffects.desktop` written under `var` folder.
 
> If the older libportal code worked reasonably well in Flatpak without a gsettings key, couldn't that then be wrapped in preprocessor directives? But obviously writing a useless file is far from ideal... but I don't have any better ideas.

It could be done, but it's harder to maintain. If you maintain the build for Flatpak, you could apply the patch from my last commit so you have the libportal feature with a gettings key.

vchernin:
> It could be done, but it's harder to maintain. If you maintain the build for Flatpak, you could apply the patch from my last commit so you have the libportal feature with a gettings key.

I thought of maintaing the libportal patch in the Flatpak repo, but that's probably a last resort, only if either version of the libportal code can't reasonably be maintained upstream. 

I think either patch is suitable for Flatpak, I haven't extensively tested the gsettings version yet but it probably works fine like the non-gsettings version.

Digitalone1:
Please test the one from my repository. It's better and if you could make a patch to apply it to Flatpak builds it's better for us and for Flatpak users.

vchernin:
You mean this one right? https://github.com/Digitalone1/easyeffects/commit/cea50ac40361333ce40ad72216e7d804806ada88

I already built that one here: https://github.com/flathub/com.github.wwmm.easyeffects/pull/5#issuecomment-918398908

If it works correctly I'll clean up that Flathub PR and push that to stable, if it's also alright with @wwmm.

wwmm:
>if it's also alright with @wwmm.

I do not see any problems. It will require more work on the Flatpak side because if we change something in this part of the code the patch applied there will probably have to be updated. Same thing if changes are made to the gsettings schema file. So some level of maintenance may be required on the flatpak package side. But I do not have a better idea. It is either this or hiding the switch.

vchernin:
I tested the latest patch a bit, I think it is suitable to release on Flathub. Thank you both for the working on/discussing this, it is a nice thing to have for Flatpak users. Hopefully one day libportal will be usable on the host.

Now the patch does require fairly recent master to apply, so I can't easily backport it to 6.1.0 on Flathub. But since a new release could hopefully be done soon it can just wait until then I think.

Oh and unrelated, but apparently Christian from Red Hat's desktop team thinks EasyEffects is "a great example of a cool new application built with PipeWire" :)
https://blogs.gnome.org/uraeus/2021/09/16/cool-happenings-in-fedora-workstation-land/

Digitalone1:
You could use the latest master, it's quite stable. Could name it 6.1.0.1, waiting for the next release.

PR #1154, opened 2021-09-16
Fix app icons lookup

### LINKED ISSUE #1143: OBS icon is not shown in the app list
@Digitalone1  I am trying to use OBS today and I noticed we have no icon for it while the app provides an icon in its package. So far I see two problems

- The icon is installed. But for some reason the PipeWire property I usually use for this does not have the icon name. But GNOME can load it with no problems. I am not sure how yet.
- Hiding the icon just because the icon theme does not have it. In the logs I see `effects_base_ui: obs icon name not installed in the Adwaita icon theme in use. The application icon has been hidden.`

After what I am seeing with OBS assuming that icon themes must have the icon for us to show it is not a good idea. We should be able to show OBS icon outside of Flatpak even if the icon theme does not have it because the OBS package is installing the icon. Unless something out of the ordinary has to be done to load this icon. Then it may not be worth the effort. But I still think that we should be able to show icons inside `/usr/share/icons/hicolor/256x256/apps/` or similar paths. It is strange they do not seem to be working.

Digitalone1:
Fixes #1143

More robust way to lookup icons. Pixmaps directories are checked if the icon is not present in the icon theme - tested with Adwaita and Brave (the original one, not brave-bin) which install the icon in pixmaps.

Cases where Pipewire sets an incorrect icon name are now handled with a map.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,8 @@
 - When editing presets files in an external editor, duplicated entries won't be shown in our presets menu.
 - Now the blocklist is correctly set when switching presets.
 - Now the status of the global bypass button is correctly updated when changing plugin stack.
-- Missing icons on the system should not be shown inside the application info UI.
+- Missing icons on the system should not be shown inside the application info UI (if an application icon could not be
+  shown even if you're sure it's correctly installed, please open an issue).
 - Some icons not showing in Plasma DE with Breeze icon theme should appear now.
 
 ## [6.1.0]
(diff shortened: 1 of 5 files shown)


Digitalone1:
@wwmm I think it's quite ready for a new release if nothing shows up in the next week.

Before to do it, please notice so I can make a translation update.

wwmm:
Ok. We can do one now. So you can prepare the translations. I think I have an idea to handle to spectrum and level meters update frequency but I am not sure when I will have time to try it and see if it will really work. So it is not worth delaying the next release because of it.

Issue #1142, opened 2021-09-10
Only FL and FR speakers works with easyeffects.


MarcinW87:
Hi. I have kind of strange issue. My default setup is 5.1. When easyeffects is disabled all speakers works fine, but situation changes dramatically when it is enabled. Only front left and front right is working. 

<a href=https://pastebin.com/e3LMhUhg>pw-dot</a> with easyeffects
<a href=https://pastebin.com/7Nqaaqyc>pw-dot</a> without easyeffects

I'm using pipewire 0.3.34

Digitalone1:
This is a known issue, Easyeffects is a stereo application and at the moment the output pipeline is linking only the front ports, so only FR and FL are working. 

There was an attempt to resolve this problem making an upmix, but it's not working anymore for unknown reasons and Pipewire is not giving us an easy solution to resolve this issue.

wwmm:
It is like Digitalone1 said. This used to work in the past when we used GStreamer because like Pulseaudio PipeWire does upmixing/downmixing automatically when streams are linked to devices. But for many reasons it is better to do what we do with filters instead of streams. But the downside is that we will have to handle upmixing/downmixing ourselves.

THere is something Pipewire's developer suggested to do that you may try yourself in PipeWire's configuration https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/1447#note_1006488. If you create a virtual device with 2 inputs and the outpus matching your surround configuration you can make EasyEffects use it as output. Until I find a way to handle the upmixing/downmixing it is probably the only way.

Digitalone1:
Can be closed as a duplicate of #1126 

A possible workaround is reported in https://github.com/wwmm/easyeffects/issues/1126#issuecomment-1141315558

Issue #1126, opened 2021-09-02
mutichannel (routing)?


capoei:
Is there any way to add multi-channel mangement/routing?
just as an example my use case (which would also require the adition of the possibility of a other instance of the convolver):
stereo signal > room correction convolution > crossovers via convolution resulting in 4 channels > routing of channels to my audio interface outputs

wwmm:
>Is there any way to add multi-channel mangement/routing?

Unfortunately not. There are other open issues asking for similar multichannel features too. But integrating them in an application that evolved around stereo audio processing is really hard. From the user interface to our presets system and effects pipeline the amount of changes required to handle multichannel features properly is huge. And to make things harder I do not have a surround system. My speakers are a stereo 2.0 from Edifier. They are good but not suitable to actually develop multichannel features.

As what you ask seems to require that only the output node actually need to have more than 2 outputs it may be a little more doable than what other people are asking. Let's keep this issue open for now.

>which would also require the adition of the possibility of a other instance of the convolver

Like most of the plugins we use our convolver only has 2 input/output ports. A new plugin with capable of output more than 2 channels while having 2 input channels would be required. And it would probably have to be the last plugin in the pipeline.

capoei:
so there is no wa to include a "split" plugin that splits the plugin chain into 2 or more (stereo ones)?
also when I talked about another instance, it is atm not possible even in the stereo chain. I can't put one convolver after another.
and yes, the crossover would be the last in the chain anyways. so mayby it is posssible to make a "output plugin" where you could put other plugins in paralel?

capoei:
the "output plugin" could have a split feature, where on every split you get 2 new outputs with the stereo inputs. that way the 5.1 guys could make 4 chains and control front/back/center/LFE independently on the output

wwmm:
>so there is no wa to include a "split" plugin that splits the plugin chain into 2 or more (stereo ones)?

Not inside EasyEffects. But I think it can be done if you create a PipeWire loopback device and set it as your default output. OR if you select it as EasyEffects output in our device selection menu. Loopback devices should be able to do something close to what you want. Take a look at https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/1447#note_1006488

capoei:
is it maybe possible to extend the output section in the pipewire tab so that we can select a running aplication as output device?

wwmm:
>is it maybe possible to extend the output section in the pipewire tab so that we can select a running aplication as output device?

[...]

PR #1171, opened 2021-09-27
Loopback test for output stream down/upmixing

Digitalone1:
Related to #1142, #1126 and #1041.

This is a test and I made it only for curiosity following directives [here](https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/1447#note_1006488). 

It implements a loopback output stream for downmixing and upmixing, hopefully performed by Pipewire on the output device. I can't test it because I have only stereo devices, but at least It's playing audio unlike the previous attempt with a virtual source that it's not working with the latest Pipewire versions.

The loopback mode is enabled with a gsettings key. At the default the operating mode is the traditional with manual linking on the output device. I added a check on links to disable the loopback mode if a bad configuration is made by the user.

I also changed other small things, plus the addition of a workaround to the issue related to Pavucontrol not blocklisted in localized sessions.

Anyway I made the pull request only for reference, I don't feel it's worth to merge it and introduce a loopback stream to make the channel remapping on non-stereo devices, this should be handled automatically by Pipewire without the need to spawn another stream and make it visible to users that can mess with it.

But I hope someone can test this because I'd like to know just for the curiosity if the upmixing is really done on output stream. Later or in the next days I'll backport the changes without the loopback mode in a different PR.

### DIFF
--- a/data/schemas/com.github.wwmm.easyeffects.gschema.xml
+++ b/data/schemas/com.github.wwmm.easyeffects.gschema.xml
@@ -7,6 +7,9 @@
         <key name="process-all-inputs" type="b">
             <default>false</default>
         </key>
+        <key name="loopback-mode" type="b">
+            <default>false</default>
+        </key>
         <key name="use-dark-theme" type="b">
             <default>false</default>
         </key>
--- a/include/application.hpp
+++ b/include/application.hpp
@@ -39,6 +39,8 @@ class Application : public Gtk::Application {
   auto operator=(const Application&&) -> Application& = delete;
   ~Application() override;
 
+  inline static bool loopback_mode = false;
+
   static auto create() -> Glib::RefPtr<Application>;
   Glib::RefPtr<Gio::Settings> settings;
   Glib::RefPtr<Gio::Settings> soe_settings;
--- a/include/effects_base_ui.hpp
+++ b/include/effects_base_ui.hpp
@@ -102,7 +102,7 @@ class EffectsBaseUi {
 
   std::vector<sigc::connection> connections;
 
-  void on_app_added(const uint id, const std::string name, const std::string media_class);
+  void on_app_added(const uint id, const std::string name);
   void on_app_changed(const uint id);
   void on_app_removed(const uint id);
 
--- a/include/pipe_manager.hpp
+++ b/include/pipe_manager.hpp
@@ -195,12 +195,12 @@ class PipeManager {
 
   std::vector<DeviceInfo> list_devices;
 
-  NodeInfo ee_sink_node, ee_source_node, ee_loopback_sink, ee_loopback_output;
+  NodeInfo ee_sink_node, ee_source_node, ee_loopback_sink, ee_loopback_playback;
 
   const std::string ee_source_name = "easyeffects_source";
   const std::string ee_sink_name = "easyeffects_sink";
 
-  const std::string loopback_output_name = "easyeffects_loopback_playback";
+  const std::string loopback_playback_name = "easyeffects_loopback_playback";
   const std::string loopback_sink_name = "easyeffects_loopback_sink";
 
   NodeInfo default_output_device, default_input_device;
@@ -236,6 +236,8 @@ class PipeManager {
   void disconnect_stream_output(const uint& id) const;
 
   void disconnect_stream_input(const uint& id) const;
+
+  void set_loopback_target_node(const uint& id);
 
   static void set_node_volume(pw_proxy* proxy, const int& n_vol_ch, const float& value);
 
@@ -301,6 +303,9 @@ class PipeManager {
   spa_hook core_listener{}, registry_listener{};
 
   void set_metadata_target_node(const uint& origin_id, const uint& target_id) const;
+
+  auto get_loopback_capture_property() -> std::string;
+  auto get_loopback_playback_property(const char* target_id = nullptr) -> std::string;
 };
 
 #endif
--- a/include/pipe_manager.hpp
+++ b/include/pipe_manager.hpp
@@ -22,6 +22,7 @@
 
 #include <glibmm.h>
 #include <pipewire/extensions/metadata.h>
+#include <pipewire/impl.h>
 #include <pipewire/pipewire.h>
 #include <sigc++/sigc++.h>
 #include <spa/param/audio/format-utils.h>
@@ -162,7 +163,7 @@ struct DeviceInfo {
 
 class PipeManager {
  public:
-  PipeManager();
+  PipeManager(const bool& lb_mode);
   PipeManager(const PipeManager&) = delete;
   auto operator=(const PipeManager&) -> PipeManager& = delete;
   PipeManager(const PipeManager&&) = delete;
@@ -171,10 +172,14 @@ class PipeManager {
 
   inline static const std::string log_tag = "pipe_manager: ";
 
+  const bool loopback_mode;
+  bool use_output_loopback = false;
+
   pw_thread_loop* thread_loop = nullptr;
   pw_core* core = nullptr;
   pw_registry* registry = nullptr;
   pw_metadata* metadata = nullptr;
+  pw_impl_module* loopback_module = nullptr;
 
   spa_hook metadata_listener{};
 
@@ -190,7 +195,13 @@ class PipeManager {
 
   std::vector<DeviceInfo> list_devices;
 
-  NodeInfo pe_sink_node, pe_source_node;
+  NodeInfo ee_sink_node, ee_source_node, ee_loopback_sink, ee_loopback_output;
+
+  const std::string ee_source_name = "easyeffects_source";
+  const std::string ee_sink_name = "easyeffects_sink";
+
+  const std::string loopback_output_name = "easyeffects_loopback_output";
+  const std::string loopback_sink_name = "easyeffects_loopback_sink";
 
   NodeInfo default_output_device, default_input_device;
 
@@ -218,13 +229,13 @@ class PipeManager {
 
   auto stream_is_connected(const uint& id, const std::string& media_class) -> bool;
 
-  void connect_stream_output(const uint& id, const std::string& media_class) const;
+  void connect_stream_output(const uint& id) const;
 
-  void connect_stream_input(const uint& id, const std::string& media_class) const;
+  void connect_stream_input(const uint& id) const;
 
-  void disconnect_stream_output(const uint& id, const std::string& media_class) const;
+  void disconnect_stream_output(const uint& id) const;
 
-  void disconnect_stream_input(const uint& id, const std::string& media_class) const;
+  void disconnect_stream_input(const uint& id) const;
 
   static void set_node_volume(pw_proxy* proxy, const int& n_vol_ch, const float& value);
 
@@ -255,8 +266,8 @@ class PipeManager {
 
   static auto json_object_find(const char* obj, const char* key, char* value, const size_t& len) -> int;
 
-  sigc::signal<void(const uint, const std::string, const std::string)> stream_output_added;
-  sigc::signal<void(const uint, const std::string, const std::string)> stream_input_added;
+  sigc::signal<void(const uint, const std::string)> stream_output_added;
+  sigc::signal<void(const uint, const std::string)> stream_input_added;
   sigc::signal<void(const uint)> stream_output_changed;
   sigc::signal<void(const uint)> stream_input_changed;
   sigc::signal<void(const uint)> stream_output_removed;
@@ -279,6 +290,8 @@ class PipeManager {
 
   sigc::signal<void(LinkInfo)> link_changed;
 
+  sigc::signal<void()> loopback_mode_disabled;
+
  private:
   bool context_ready = false;
 
@@ -286,6 +299,8 @@ class PipeManager {
   pw_proxy *proxy_stream_output_sink = nullptr, *proxy_stream_input_source = nullptr;
 
   spa_hook core_listener{}, registry_listener{};
+
+  void set_metadata_target_node(const uint& origin_id, const uint& target_id) const;
 };
 
 #endif
--- a/include/pipe_manager.hpp
+++ b/include/pipe_manager.hpp
@@ -200,7 +200,7 @@ class PipeManager {
   const std::string ee_source_name = "easyeffects_source";
   const std::string ee_sink_name = "easyeffects_sink";
 
-  const std::string loopback_output_name = "easyeffects_loopback_output";
+  const std::string loopback_output_name = "easyeffects_loopback_playback";
   const std::string loopback_sink_name = "easyeffects_loopback_sink";
 
   NodeInfo default_output_device, default_input_device;
--- a/include/stream_input_effects.hpp
+++ b/include/stream_input_effects.hpp
@@ -42,7 +42,7 @@ class StreamInputEffects : public EffectsBase {
 
   void disconnect_filters();
 
-  void on_app_added(const uint id, const std::string name, const std::string media_class);
+  void on_app_added(const uint id, const std::string name);
 
   void on_link_changed(LinkInfo link_info);
 };
--- a/include/stream_output_effects.hpp
+++ b/include/stream_output_effects.hpp
@@ -40,9 +40,11 @@ class StreamOutputEffects : public EffectsBase {
 
   void disconnect_filters();
 
-  void on_app_added(const uint id, const std::string name, const std::string media_class);
+  void on_app_added(const uint id, const std::string name);
 
   void on_link_changed(LinkInfo link_info);
+
+  void on_loopback_mode_disabled();
 };
 
 #endif
--- a/src/application.cpp
+++ b/src/application.cpp
@@ -114,7 +114,9 @@ void Application::on_startup() {
 
   create_actions();
 
-  pm = std::make_unique<PipeManager>();
+  loopback_mode = settings->get_boolean("loopback-mode");
+
+  pm = std::make_unique<PipeManager>(loopback_mode);
   soe = std::make_unique<StreamOutputEffects>(pm.get());
   sie = std::make_unique<StreamInputEffects>(pm.get());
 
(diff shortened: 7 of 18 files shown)


wwmm:
Even without a surround amplifier I still can force my onboard card to go to the 5.1 mode. I can not plug anything to test if there is actually audio being played because the card will alternate to stereo analogical mode immediately. But the links are made
Screenshot from 2021-09-27 11-35-49

So I imagine that there will be audio. But I am still uncomfortable with the side effects of using a loopback device. Pavucontrol shows 2 sinks but only one of them is the output device that should be used. But muting both of them causes the audio to be muted. I can see the users that were confused by the GStreamer streams coming back with questions...

There are other problems too. Disabling the loopback mode on they fly will require a certain amount of code complexity. And as far as I could see after enabling both of my sound cards the manual device selection is currently broken when the loopback mode is enabled. No device change is made.

Most of these problems can be fixed in one way or another. But based on what I have seen so far the amount of code and changes involved does not seem to offer an advantage over handling the remixing on our side. At least in this case most of the changes would be contained in the output level filter. I have to find some time to explore this other alternative before going down the loopback device path...

Digitalone1:
> There are other problems too. Disabling the loopback mode on they fly will require a certain amount of code complexity. 

I didn't disable it really because I can't do it. The only thing I did when I detected the wrong connection from the loopback output to easyeffects sink or loopback sink is to reconnect the filter the old way and this is done until the app is shutdown.

> And as far as I could see after enabling both of my sound cards the manual device selection is currently broken when the loopback mode is enabled. No device change is made.

I did not make the device change because I don't know how to do it and this is supposed to be done by Pipewire, so if we really want this to be done, we have to declare a specific device as default so the loopback output is automatically connected to it, and at least Pipewire should do it.

Or one can change it from Pavucontrol, but it's not ideal.

> Most of these problems can be fixed in one way or another. But based on what I have seen so far the amount of code and changes involved does not seem to offer an advantage over handling the remixing on our side. At least in this case most of the changes would be contained in the output level filter. I have to find some time to explore this other alternative before going down the loopback device path...

I don't agree on the remixing by our side because I think that would be more complex, but on the other side I think also that the loopback node is not the ideal solution. This should be handled from Pipewire with a specific property, but unfortunately it does not give us that option and that's a regression compared to Pulseaudio.

wwmm:
>I don't agree on the remixing by our side because I think that would be more complex

I agree that handling the remixing is not cool. But the thing is that the loopback device solution is also requiring changes all over the place. So in a way it is also a complex solution. The difference is just where the complexity is going. For example, at least in theory if we extend the output level filter so it handles more than 2 output channels the current problem about the manual selection device not working would not happen because the whole logic involved for this would be the same we have now. The only difference would be the number of channels being linked. With the loopback device that code has to be changed in a way that changes in selection result in calls that change the device the loopback is linked to. Both paths require work. But the loopback device solution is forcing us to change sections of the code that are working just fine.  

>I did not make the device change because I don't know how to do it and this is supposed to be done by Pipewire

In this case we are the ones that have to do it. The loopback device is creating a stream. It probably can be moved around using the same metadata functions we use to move apps to our devices. But then we go down the path I wrote above. The loopback solution is requiring more and more code. And the chances of introducing bugs in places that were fine gets higher. When the PipeWire developer suggested me this solution I hoped it would be something simple requiring few changes. And the the only downside would be the additional entries in pavucontrol. But in the end the amount o code being written is getting larger than the one a custom remixing on our side would need.

wwmm:
>Upmixing algorithm are very complex. I tried to read what Pipewire is doing and is very hard to understand, They use a matrix and link every port with a specific value to mix the channels.

It depends on the kind of upmixing that is being tried. A different value is used for each port because each of them needs a different fraction of the input signal. It is definitely more involved than converting mono to stereo where you just take half of each channel but the more annoying part is probably writing a different setting of equations for each mapping. And finding these equations. Honestly I prefer that than what the loopback is requiring. At least most of this new code would be in the filter.

>Already tried this. It's not working. That function seems to work only on simple output streams made by applications. Used on other things like filters or loopback streams causes no result.

Humm... That it does not work on filters is known. But internally the loopback module creates streams. Maybe the way to make the call has to be different.

wwmm:
Something else I should find time to evaluate is the other suggestion Taymans gave https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/1447#note_1003024

```
This should normally be done by doing a PortConfig on the sink to configure it in stereo mode. It'll then have 2 input ports and will upmix (when enabled) to the device profile format (5.1 presumably).
```

Maybe this is something more reasonable to do.

wwmm:
I decided to make more questions in the issue I opened about this in the PipeWire page https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/1447. Including about how to move the loopback device stream.

wwmm:
Another possibility to consider is creating a custom output level filter that also has 2 inputs and 2 outputs but that is based on streams instead of a pw_filter. That is essentially how PipeWire's filter chain plugins and possibly the loopback device module are implemented. With lucky this should reduce the amount of code that is being needed on our side. And it would not require loading modules or adding settings to switch modes. But the code handling links from it to the hardware would still need changes.

capoei:
on this test https://github.com/wwmm/easyeffects/issues/1126#issuecomment-927857650
I had no signal other than FL and FR

capoei:
and I think you just need to create e multi-channel sink to test this stuff in your stereo setup. you can use LSP spectrum analyser x12 to see what is on the output

wwmm:
>I had no signal other than FL and FR

I did not understand. I though it had worked as all the ports are linked. Even in this case there is audio only in the front left and right channels?

Digitalone1:
Well, if it's not working I wonder what is the purpose of the loopback node.

Maybe the culprit is the property `stream.dont-remix = true`, but it was suggested [here](https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/1447#note_1006488).

Another difference is that I deleted the playback prop because I thought it's not really needed since the loopback node should handle the remixing. The playback prop was spawning a source node with I was not using because we only have to capture.

It seems the linking is correctly done with the proper ports, but the stream is not upmixed, so only front ports are transmitting while the others are mute. It seems the property don't remix is preventing the upmix stage, but why @wtaymans recommended it?

Digitalone1:
> and I think you just need to create e multi-channel sink to test this stuff in your stereo setup. you can use LSP spectrum analyser x12 to see what is on the output

Thanks. I'll try the spectrum analyser in carla-rack.

wwmm:
>It seems the property dont-remix is preventing the upmix stage, but why @wtaymans recommended it?

Who knows... I also think this may be the reason why only the first 2 channels have audio.

wwmm:
>I also think this may be the reason why only the first 2 channels have audio.

After reading PipeWire's wiki I see I was wrong
```
The stream.dont-remix property makes sure that only the FL port of the capture device is linked (and not all ports mixed into FL).
```

wwmm:
I mean... It still feels like it is related. But the part about `and not all ports mixed into FL` sounds weird. I do not understand why this would even be allowed in the first place.

capoei:
> > I had no signal other than FL and FR
> 
> I did not understand. I thought it had worked as all the ports are linked. Even in this case there is audio only in the front left and front right channels?

yes, no upmxing happening

I just tired 
`channelmix.upmix = true`
in 
client.conf

but still no upmix

wwmm:
After reading a few examples I think I understand. That particular case where the explanation for dont-remix is given has only the FL channel defined. So in that situation remixing would make many channels to be mixed in the FL. The explanation there is not a general statement for all the possible situations. In our case we need the remix.

Digitalone1:
@capoei please reinstall it and test. It's faster if you can try rather than making me configure a multichannel sink on my system.

capoei:
> @capoei please reinstall it and test. It's faster if you can try rather than making me configure a multichannel sink on my system.

I recompiled, restarted....but all the extra channels are still dead

Digitalone1:
Okay, the options are two. The loopback is useless or we explicitly need the playback prop also, spawning another node in the graph (this is overkill).

wwmm:
>Okay, the options are two. The loopback is useless or we explicitly need the playback prop also, spawning another node in the graph (this is overkill).

Maybe it is the missing playback props indeed. I am having another busy day at work. What exactly was removed? But it is still strange considering that the links are all there. I wonder why PipeWire would link a bunch of dead channels. Could it be they are just muted?

Digitalone1:
I'll make the last attempt adding the playback prop, but it's only for testing, there should be an easier way...

wwmm:
>spawning another node in the graph (this is overkill).

Definitely. But if we understand how to make this work it will be worth to invest time in my idea about writing a hybrid output level filter that keeps the input based on pw_filter and uses the streams only to send audio to the hardware device. The is a good chance that the overkill could be removed in this alternative approach. And as streams would be used there would be no need to do custom upmixing either. Depending on how well it goes we could just use that hybrid device in all cases removing the need to even have a setting to enable this. But if the loopback does not work even now it would be pointless to try to write a hybrid filter.

Digitalone1:
> What exactly was removed? 

The `dont-remix` and `passive` properties from capture.

wwmm:
>The dont-remix and passive properties from capture.

Try to set dont-remix to false. Keeping the links passive is a good idea.

Digitalone1:
I noticed a mistake I made. I set `audio.position = "FL,FR"` on the loopback output while Wim set it on capture and playback. Maybe the issue is there. Anyway I'll restore the playback and set the audio position only on the capture and the playback.

Digitalone1:
So I made some interesting discoveries.

The loopback has a configuration made by three parts:
* the main one related to the whole loopback module
* the capture node
* the playback node

If I specify the capture and the playback configuration, the module spawns two nodes, the playback (which is supposed to make the remap) and the capture (which should be a sink to be connected to output streams).

As default, if I don't specify a media class for the capture, it will be an `Audio/Input/Stream`, impossible to be connected to filters. So I have to specify `Audio/Sink` as media class.

If I omit the playback configuration, the module spawns a capture and a playback the same way as above. The configuration in the main part is forwarded to the playback properties, so specifying the `audio.position` in the main part is the same as omitting in the main and writing in the playback one.

With the changes I made it's basically the same as before. I don't think it will work, but I hope I'm wrong.

Digitalone1:
@capoei Please, recompile when you have time. This is the final chance.

capoei:
> @capoei Please, recompile when you have time. This is the final chance.

sorry, no change with r9

I hope my card is not the problem because it is showing 16 outputs. the first 8 are named correctly for a 7.1 system though

capoei:
> > I hope my card is not the problem because it is showing 16 outputs. the first 8 are named correctly for a 7.1 system though
> 
> What happens when you play a stereo video on Youtube when EasyEffects is not running? Does that image with the nodes also show the links but there is audio in all of the channels?

It looks the same, just that all the channels are coming from the browser. ad no, upmixing doesn't work like this either.
I just created a 5.1 sink with 
pw-cli create-node adapter '{ factory.name=support.null-audio-sink node.name=my-sink media.class=Audio/Sink object.linger=true audio.position=[FL FR FC LFE RL RR] }'
chose it as the sink for the browser in pavucontrol. 6 channels form the browers conect to 6 channels from the sink. and there is no upmixing either. I guess it is just not working.
btw: after creating that sink, one can run the spectrum analyser and manualy conect the outputs easily with something like Helvum.

capoei:
ok wait. got it working lol
upmixing has to be enabled in pipewire-pulse.conf

capoei:
So I just found this
`pw-loopback -m '[ FL FR ]' --capture-props='media.class=Audio/Sink node.name=my_sink' --playback-props='node.target="my-default-sink"'`
in https://gitlab.freedesktop.org/pipewire/pipewire/-/wikis/Virtual-Devices

if I use this device as an output for a stereo stream it will do the upmixing

not sure this helps

Digitalone1:
Thanks, maybe it's just a matter of setting the `node.target` on the playback property to make the linking, this would help also on the output device change.

I will test on my system when have time.

Digitalone1:
Today I should have the time to test this playing with the node target and seeing if the output device can be redirected.

capoei:
I am here if needed

Digitalone1:
No luck. Setting the node.target is the same as before. No upmixing (tested on a mulrichannel sink and spectrum analyzer) and no redirection updating the target.node property on the loopback module (so no output device change, but it's modified when switching inside Pavucontrol).

Maybe we should try to update the node.target on the playback node directly rather than updating the loopback, but I can't find a way to locate the pointer of the property object of the node. There's no way to get it with the id and NodeInfo structure does not help us in this case.

Now I wonder how we can make the upmix working if handling with the properties don't change nothing.

On this tasks I don't understand how Pipewire is built. It's very complex. Sadly I have to say PulseAudio was way better.

Digitalone1:
The only thing made differently by pw-loopback command is that it is setting the target.node on the initialization, but we can't do this because we don't know the output device in PipeManager constructor. And anyway we should be able to update it after the initialization when the output device is changing.

Then pw-loopback is setting the node name rather than the id, but I saw in Pipewire code that the node.target property can also accept node id.

wwmm:
>The only thing made differently by pw-loopback command is that it is setting the target.node on the initialization, but we can't do this because we don't know the output device in PipeManager constructor. And anyway we should be able to update it after the initialization when the output device is changing.

Well... That is unfortunate... In any case I think I will still try the hybrid filter approach. Maybe it will be easier to handle the target device this way. The question is when I will have time for that with tons of students exams to grade...

Digitalone1:
> In any case I think I will still try the hybrid filter approach.

For sure, loopback has to be avoided because it has very bad implementation.

I noticed that if I set the node.target in the argument before the initialization, I see it in `pw-cli info`, so it is set only at the beginning. Setting it afterwards or updating, we got no results. I understand now because it's working setting it with pm-loopback.

That said, if we would implement the loopback as it is, we should initialize it only after knowing the output device. And still we don't know how to redirect it. Wim says metadata works, but it doesn't.

wwmm:
>I noticed that if I set the node.target in the argument before the initialization, I see it in pw-cli info, so it is set only at the beginning

That I was already expecting because it was the same with GStreamer.  Some tags are only used when initializing.

>That said, if we would implement the loopback as it is, we should initialize it only after knowing the output device. And still we don't know how to redirect it. Wim says metadata works, but it doesn't.

Yes, that is a little concerning. What I am thinking about is creating an output stream in a similar way as it is done in the loopback module https://gitlab.freedesktop.org/pipewire/pipewire/-/blob/master/src/modules/module-loopback.c#L275. In the worse case scenario we can try to disconnect and reconnect the stream with the proper id instead of `PW_ID_ANY`. But I still hope it will work through the metadata. IT worked with GStreamer and what GStreamer does internally is calling `pw_stream_connect` just like the loopback module is doing.

Digitalone1:
If you give me some directive, I can try to implement your idea.

For the redirect I was also thinking of downloading pavucontrol and looking what it does to connect streams to devices. Maybe if Pipewire doesn't give us an easy method, we could try with the old Pulseaudio API.

Digitalone1:
Pipewire should have something like this:

https://gitlab.freedesktop.org/pulseaudio/pavucontrol/-/blob/master/src/sinkinputwidget.cc#L125

I wonder if I create a simple null sink and move it to output device through that function, will it work?

wwmm:
>we could try with the old Pulseaudio API.
>I wonder if I create a simple null sink and move it to output device through that function, will it work?

Nope. That is going to require a huge amount of work and is not what we want to do in the long term. Besides the fact that Pulseaudio libraries will become one of our dependencies to use them we have to initialize a Pulseaudio mainloop and context. in other words ressurecting the long gone class `PulseManager`. We definetily do not want to handle one mainloop for PipeWire and another for Pulseaudio at the same time. Is asking for guaranteed trouble.

Digitalone1:
Never mind. Write here your idea and give me some directive. Maybe I can do it. Or at least I try.

wwmm:
>Never mind. Write here your idea and give me some directives. Maybe I can do it. Or at least I try.

I think I will be able to least start it this weekend. I am still considering if inheriting from `plugin_base` will be viable or if an independent class will be needed considering the differences to the other filters that use pw_filter ports in the output. So not all of the steps are totally clear in my mind yet. If possible try to focus on the other issue related to the node holders. And maybe on making the node_info callback in the pipe_manager file emitting less signals. I doubt we really have to emit one every time PipeWire tells us something has changed in one of the nodes. A better approach would probably be to emit only when one of the fields in our structures has changed. All the info is there. We just have to think about a clever way to check field by field if at least one of them is different.

wwmm:
>A better approach would probably be to emit only when one of the fields in our structures has changed.

The reason is that nodes have many properties. We do not have even half of them in our NodeInfo structure. So there is a high chance we can just ignore most of PipeWire's notifications about changes in them if we focus only on what is related to what we actually use.

wwmm:
I had some time to start the hybrid filter today. As I do not know if it will really be useful and the approach is quite different from the one in PluginBase I decided to create a new class that does no inherit from anybody https://github.com/wwmm/easyeffects/blob/ww_playground/src/output_stream.cpp. Most of the boilerplate is done but it isn't working yet. Besides doing the link to other filters I still do not fully understand how to prepare the buffers that will feed the playback stream. I was hoping I could just passthrough the buffers from the pw_filter input ports but the streams seem to use something in a different format =/

I am also a little concerned about the dynamic latency and sampling rate management in streams. It is not clear they allow them to change on the fly or not. I hope they do.

wwmm:
As a reference for myself it seems this is what I will have to read to prepare the buffers for the playback stream https://gitlab.freedesktop.org/pipewire/pipewire/-/blob/master/doc/tutorial4.dox. I will take a look the next time I have time.

Digitalone1:
I close this PR since it's not the solution to the multichannel issue and a temporary workaround could be implemented as stated in https://github.com/wwmm/easyeffects/issues/1126#issuecomment-1141315558

PR #1185, opened 2021-10-03
Track nodes by timestamp

Digitalone1:
Maybe this is not needed anymore because it seems all resolved, but who knows, maybe race conditions will come up anyway in the future.

I developed this having in mind a race condition in `effects_base_ui` caused by bad applications spawning and destroying many streams in a very short time, leading issues in `listview_players` forced to append and remove the same holder item from its list.

So these are the points of the present MR:
* nodes are not tracked by their ID, but with UNIX timestamps in nanoseconds converted as string -  this way an app messing with us is not inserting and deleting the same ID from the map, but different indexes - `all_players_model` won't insert and delete the same app id, but different holders - nothing changes by user interface side, but under the hood the things are different
* getter helper are provided in `PipeManager` to pick the node by ID - there's also a getter returning a pointer, so we can check easily for nullptr rather than wrapping every selection in a "try catch" statement
* Regarding signals, `stream_output_added` and `stream_input_added` continue to emit a NodeInfo copy for convenience - this way we can get the only needed data and save them in the `NodeInfoHolder` rather than having a long list of arguments
* `stream_output_changed`, `stream_input_changed`, `stream_output_removed`, and `stream_input_removed` sends only the timestamp because it's the only element needed for operations on `listview_players` 
* `info_updated` callback gets the whole NodeInfo as copy in argument, so we reduce access to NodeInfoHolders pointer
* `scheduled update` flag is removed because `PipeManager` callbacks track the changes related to information showed in `app_info_ui`, therefore only needed updates are done
* operations on `listview_players` are mutually exclusive, so a `GMutex` has been introduced for appending holders, deleting holders and resetting holders for `blocklist_checkbutton` and `show_blocklisted_apps` switch changes
* algorithm for blocklist changes has been optimized
* `on_app_changed` has been optimized to work only on the `players_model` which is the only useful in that context (I made a mistake, `all_players_model` is not used, so emitting signals on its holders is doing nothing)
* deleting a node from the map, the thread waits 200 ms for safety to execute the eventually scheduled updates remained on that node (maybe useful on slow systems when an app is spawning and deleting soon the stream) - proxy pointer is set to nullptr in the meantime and accessing it in `effects_base_ui` we always check for nullptr.

@wwmm don't know if you like this solution, I feel it more safe. Maybe you can release the fix for `connection_info` issue and then merge this to test for the other next release.


wwmm:
>don't know if you like this solution, I feel it more safe.

I do not agree with those mutexes. Besides the fact that all the code on the ui side is running on the same thread they won't protect us from changes gtk does in its own internal threads. If gtk decides to destroy something I doubt locking a mutex on our side will protect us from anything. And there are places where the mutex is locked but the unlock call is inside a if called in a for loop. Although the condition will probably be true in all cases at the end of the day there is the chance that it may fail. And the lock won't be unlocked. And even if those mutex were the true solution to the problem it would be better to use a c++ lock guard or the more advanced version of it whose name I do not remember now. Calling lock and unlock manually even with the c++ version of the mutex can cause unexpected locks if exceptions happens a long the way.

wwmm:
When clicking on the blocklist checkbox to blocklist its entry it instantly removed from the list. I have to go to the blocklist menu and remove it from there for the entry to be visible again. Is this intended?

Digitalone1:
> When clicking on the blocklist checkbox to blocklist the app entry it instantly removed from the list. I have to go to the blocklist menu and remove it from there for the entry to be visible again. Is this intended?

Yes, if show blocklisted apps is unchecked. 

So what do you suggest? Removing the mutexes or replacing with an advanced C++ mutex?

wwmm:
>Yes, if show blocklisted apps is unchecked.

Ok


>So what do you suggest? Removing the mutexes or replacing with an advanced C++ mutex?

Removing them. It is very unlikely they will protected us from gtk object destruction. But as it is now they can lock the whole app if bad conditions come together.

Digitalone1:
Okay, when I have the code at hand, I'll do it.

wwmm:
I think it is fine to save the creation timestamp and do something with it after if it helps. Although it is probably better to save it using the std::chrono type for time instead of converting it to a string. But I am not convinced yet of the advantages of using it over the id to search the node. It is not like the problem before was caused because PipeWire recreated the node with the same id. Audacity was creating and destroying streams one after the other to probe the formats and in most cases like this PipeWire creates the new stream with a different id. The holder object returned by the `get_item` function was captured by the lambda used as callback. But by the time this callback was executed gtk had already destroyed the list entry. Using timestamps instead of id does not change any of that. The real solution is not using the holder object in callbacks that may be called after its destruction as we have no control over the list entries lifetime. 

Although methods like `get_nodeptr` won't cause any harm I am not sure we really need them. `node_map.at(ts)` already returns a reference according to its docs https://en.cppreference.com/w/cpp/container/map/at. And node_map already has to be public anyway. Were they created just to wrap the usage under a try/catch?

wwmm:
>// for safety reasons, we wait for scheduled updates to be executed before deleting the node
>std::this_thread::sleep_for(std::chrono::milliseconds(200));

I understand what you are trying to achieve. But this will change a lot from one hardware to another. 200 milliseconds may be enough on some computers for the operation to be finished on the gtk side. But not on others. It will also depend on how busy the cpu is. If we want gtk to use the node info information without concerns about whether the map entry was deleted or not I see only two ways. Pass the node info by copy in the signals and use it instead of searching the node map or trying to synchronize the main thread and the PipeWire thread through mutexes.  But this kind of synchronization is really hard to do right.

wwmm:
In case you are concerned about the performance loss involved in passing things by copy one good thing to do is using `sudo perf top -p pid_number` where pid_number is easyeffects pid. This will show which functions called by easyeffects have the biggest impact on performance. The functions used internally by the plugins have the biggest impact on performance by very far. I think we can have the luxury of using copies in the other places if they make things simpler to manage.

wwmm:
One example of `perf top` output
Screenshot from 2021-10-03 13-37-03
The plugins are the bottleneck by far.

wwmm:
>don't know if you like this solution, I feel it more safe.

In shot I think that there are good things in this MR. We just have to fine tune it a little. And maybe make more usage of `perf top` to device where it is really worth to optimize. As much as I dislike passing things by copy it is clear that gtk can be trick to deal with when using too much references.

I think I will make a release today based on the current master because too many people are coming here reporting a crash that already seems to be fixed.

wwmm:
>But I am not convinced yet of the advantages of using it over the id to search the node.

It is not that I think it has to be necessarily moved back to id. The first feeling is that we may be just over complicating things.

Digitalone1:
I thought about timestamps because I was seeing Pipewire assigning the same ID to celluloid which I use the most for tests. Timestamps are always different because are tied to creation time. It's okay to return to ID if you want.

In this MR info_updated callback does not use the holder pointer, so this should be enough.

I made the get_nodeptr only to not wrap the node selection in the try catch statement. Checking for nullptr seems better sometimes rather than catching an exception.

Anyway where in the code there's the need to pass by copy now? We do it for stream adding. We don't do it for stream changes and remove, but on_node_changed passes a node from the map and it's copied by value by info_updated callback.

It seems the better things to do are taking these changes, then removing the mutex, the thread wait and restore the map indexed by id. Let me know.

wwmm:
>I thought about timestamps because I was seeing Pipewire assigning the same ID to celluloid which I use the most for tests. 

Really? I see. Usually the id changes. I thing we can keep the timestamp. Although it is not entirely clear if the id repetition is a problem it isn't a big dealt to use timestamps. The mutexes that were dangerous.

>I made the get_nodeptr only to not wrap the node selection in the try catch statement. Checking for nullptr seems better sometimes rather than catching an exception.

Ok then.

>Anyway where in the code there's the need to pass by copy now? 

I wasn't thinking about a place in specific in the current MR but about the changes from past MR. In general we are conditioned to avoid copies for performance reasons but after the latest problems with gtk objects lifetime I think we should be more careful in general. The plugins are so heavy on cpu that any performance we could possible gain from passing by reference in some cases is negligible. It is not about passing by copy everywhere. But passing by copy if we see that the implementation will be simpler. Where it is simple and safe passing by reference there should be no need for changes. 

>It seems the better things to do are taking these changes, then removing the mutex, the thread wait and restore the map indexed by id. Let me know.

I agree. It is probably enough.

wwmm:
>I agree. It is probably enough.

By that I meant removing the mutexes and the thread wait.

Digitalone1:
For curiosity, can you check if a player like celluloid or mpv is spawning with the same ID after the closing?

wwmm:
>For curiosity, can you check if a player like celluloid or mpv is spawning with the same ID after the closing?

I did some tests now. It seems PipeWire behavior has changed indeed. If I restart mpv or vlc without starting other client in the middle the id is reused. It did not use to be that way. It was more random in the past.

I also noticed another thing. If I start mpv, close it and start vlc the id used by vlc is the same mpv was using before. As things are right now the id only changes if another client is started before restarting the one that was closed.

wwmm:
PipeWire seems to be intentionally reusing the id. I wonder if at some point the server can be in a state where there is no new number to give.

wwmm:
>the id only changes if another client is started before restarting the one that was closed.

I imagine the same effects can happen if other types of nodes are started. Like the ones associated to devices.

Digitalone1:
@wwmm I made changes as specified. No issues on my side.

wwmm:
Ok. I did a quick test now and for some reason the status bar is not being updated with our virtual device sampling rate
Screenshot from 2021-10-04 10-58-54

wwmm:
`soe->pm->sink_changed.connect` is being called in `StreamOutputEffectsUi`. So something must be wrong in the value that is coming.

Digitalone1:
I didn't notice it. It's happening to me also. I will investigate.

wwmm:
The `nd_info_copy.rate` on the PipeManager side has the right value. Weird.

Digitalone1:
I track if the rate changes to update the app_info_ui. Maybe I missed a check or I should add something to exclude rate 0. I will investigate, don't have the code at hand node. It will be resolved.

wwmm:
>I track if the rate changes to update the app_info_ui.

I think I've found the problem
Screenshot from 2021-10-04 11-31-49

`app_info_ui_changed` is set to true in the same places as `notify`. So what is probably happening is that we always stop in the first `if` and never update our virtual sink node with the new rate. I think that app_info_ui_changed is redundant in this function because we already have the notify variable. We only need it in the node_info callback.

wwmm:
I did some tests and that `if` is the problem. I will fix it after merging.

Digitalone1:
Okay.

wwmm:
I will also do some small changes to make clang-tidy happy
Screenshot from 2021-10-04 11-49-21
What it is trying to say is that the structured binding in [ts, node] is binding to a reference. And for some reasons I do not know clang-tidy suggests it is not a good idea to capture the `node` variable in a lambda function these cases. IT is also complaining about a for loop
Screenshot from 2021-10-04 11-54-02

I doubt that any of these warnings are actually something we should be concerned about. But it won't harm to make clang-tidy happy and making a copy of the node in the first case and rewriting the for loop in the second case in a more standard way.

wwmm:
And if you ever see clang tidy giving warnings like the one below you can ignore. IT does not understand much about c++20 yet
Screenshot from 2021-10-04 12-02-06

wwmm:
>And for some reasons I do not know clang-tidy suggests it is not a good idea to capture the node variable in a lambda function in these cases.

One is probably the fact that the pointer may not be valid when the lambda is called. But there may be other reasons I am not aware of.

Digitalone1:
I don't see clang-tidy suggestions. It's only working clang. Maybe I should try another plugin.

Anyway I tried to make copies in a loop and after compiling I got a warning that it's better to use reference.

wwmm:
>Anyway I tried to make copies in a loop and after compiling I got a warning that it's better to use reference.

Yes. I saw it doing that suggestion too. But right after you make a reference it complain that it is being captured by the lambda. So the way to remove both warnings was doing a copy before the lambda capture like it is done in other places.

wwmm:
If I am not mistaken the first warning about the copy in the for loop structured binding was related to the timestamp variable. But in a structured binding either you copy everybody or you pass by reference everybody.

Digitalone1:
There's no way to make clang-tidy work on atom unfortunately.

For that warning, I think there's nothing to do either.

Or we can try to iterate through iterators. If clang-tidy don't complain even on them.

wwmm:
I have already fixed both warnings in the master branch
Screenshot from 2021-10-04 14-37-25
Creating a copy of the node was enough.

wwmm:
>There's no way to make clang-tidy work on atom unfortunately.

So Atom is still lacking in this department. I used it to write code in the past but at some point it became too slow and the plugins were not doing everything I wanted. So I moved to vscode.

Digitalone1:
Maybe I'll use it in the future. Now I'm comfortable with atom.

PR #1191, opened 2021-10-06
Optimize safety

Digitalone1:
These are changes in order to improve safety.

* I read [this](https://stackoverflow.com/a/36919571) and [this](https://abseil.io/tips/112), so switched all `emplace_back` to `push_back`.
* Generally it's better to save the container size before loops, but this may cause issues if the vector has a variable size, so I reverted a previous change of mine restoring the call at `size()` member in loop checks rather than saving the size value.
* I installed `code` and saw suggestions by clang-tidy. Most of them are useless, but one seemed significant. It's better to avoid non-const static data member, so I made the icon theme a variable inside the application ui constructor to pass to effects base ui classes.

### DIFF
--- a/include/application_ui.hpp
+++ b/include/application_ui.hpp
@@ -44,8 +44,6 @@ class ApplicationUi : public Gtk::ApplicationWindow {
 
   static auto setup_icon_theme() -> Glib::RefPtr<Gtk::IconTheme>;
 
-  inline static Glib::RefPtr<Gtk::IconTheme> icon_theme = nullptr;
-
  private:
   inline static const std::string log_tag = "application_ui: ";
 
--- a/include/convolver.hpp
+++ b/include/convolver.hpp
@@ -22,9 +22,7 @@
 
 #include <zita-convolver.h>
 #include <algorithm>
-#include <ranges>
 #include <sndfile.hh>
-#include <vector>
 #include "plugin_base.hpp"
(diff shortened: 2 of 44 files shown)


wwmm:
>I read this and this, so switched all emplace_back to push_back.

Ok. In a far future it may be interesting to try to run some kind of profiler and see if on the plugins code that is processing the signals there is any performance advantage in using emplace_back. But on the ui side there is probably no advantage. I do not remember if I used them for the habit that comes from scientific code or because clang-tidy suggested.

>Generally it's better to save the container size before loops, but this may cause issues if the vector has a variable size, so I reverted a previous change of mine restoring the call at size() member in loop checks rather than saving the size value.

Ok. I imagine that the compiler is capable of optimizing the calls to `size` in this case. Or is it or the standard library implementation so dumb that saving the value before actually makes a difference? In any case it would only worth to investigate that in a place where we have a bottleneck.

>I installed code and saw suggestions by clang-tidy. Most of them are useless,

Yes, and if you take a look at https://github.com/wwmm/easyeffects/blob/master/.clang-tidy you will see that  I have disabled many of the checks because they were not helpful. As time wen by I did most of the more interesting suggestions. So nowadays most of them are related to the c++20 that clang-tidy does not understand yet.

wwmm:
>you will see that I have disabled many of the checks

The ones disabled start with a `-`

PR #1193, opened 2021-10-09
APO importer: Parse preamp

ishitatsuyuki:
The APO importer (used for e.g. AutoEQ) can now avoid clipping
out-of-the-box.

### DIFF
--- a/src/equalizer_ui.cpp
+++ b/src/equalizer_ui.cpp
@@ -601,7 +601,7 @@ bool EqualizerUi::parse_apo_preamp(const std::string& line, double& preamp) {
 
   static const auto i = std::regex::icase;
 
-  std::regex_search(line, matches, std::regex(R"(preamp:\s*+([\+-]?+\d++\.?+\d*+)\s*+db)", i));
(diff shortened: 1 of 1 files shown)


Digitalone1:
Hello @ishitatsuyuki, thanks for the PR.

Maybe we can do this directly in `parse_apo_filter` rather than adding a new method, so we won't touch the following line.

https://github.com/wwmm/easyeffects/pull/1193/commits/e061de51ffdadbe6456553e19686e2c432e99a4a#diff-1e4b90474d2a2c327940e6aaaf71cdffdc411228348005a89da5dd58966bbd99L697

ishitatsuyuki:
Not sure if I agree with that idea. Under the organization you have proposed the method no longer `returns false if we cannot parse given line successfully`. It also bloats the function's parameter which IMO makes the code harder to read.

ishitatsuyuki:
I updated the patch.

Digitalone1:
Okay. Please if you can do it, otherwise I'll do after @wwmm will merge, can you please replace all `[\+-]` in regex with `[+-]` since there's no need to escape in those cases.

Thanks.

PR #1214, opened 2021-10-16
Translations updated

Digitalone1:
Added also mbc external sidechain to preset.


Digitalone1:
No, I don't think so. The only things remaining are the upmixing issue and the documentation update, but they are for the other releases, if I manage to do them.

Issue #1108, opened 2021-08-18
What to do with documentation? 


Digitalone1:
With the new multiband compressor and limiter, there's a need to update the documentation. Besides, Markus guide for laptop should be adapted to the new plugins and there are other things to fix for other plugins. 

Anyway new updates should be also translated. Honestly, I'd like to update the main English one, but if I think I have also to translate the Italian one, I'm not motivated to do it.

Unfortunately the documentation is important, but also boring to deal with and the most boring part is the translation. Then, let's be honest, most of the available translations are partially completed and will never be completed. Therefore I suggest to remove translations and leave only the English one, kept 100% updated.

It's not only because I'm not motivated to translate it, but primarily because documentation is too important for this application and we should put all the effort in keeping it updated rather than working on translations which will never be 100% updated and partially translated ones are a big downside because they are more confusing to the user. Obviously this argument does not apply to user interface translation. 

Considering that plugin official documentations are only in English and in our documentation we refer to them, I recommend to keep the documentation really updated removing all the translations, so we have a unique place, same for all locales, easier for us to maintain and without misleading information.

What do you think?

wwmm:
>Anyway new updates should be also translated. Honestly, I'd like to update the main English one, but if I think I have also to translate the Italian one, I'm not motivated to do it.

I have the same feeling about the pt_BR translations. I stopped doing them a long time ago. Even for the window labels...

>Unfortunately the documentation is important, but also boring to deal with and the most boring part is the translation.

I agree. It is really boring. Any free time I have available for this project always goes to the code.

>What do you think?

It is a hard decision to make. But I think we could remove the translations for the documentation and see how people react. Considering how rare is to receive pull requests updating the documentation translations(or the documentation) I think that most people won't even notice they were removed.

servimo:
I could try to translate to "pt_BR" , but I will use a lot of Google Translate.

servimo:
Some words translated to portuguese-BR looks strange when is about some things like: Gate,  Low-Pass, High-Shelf, Slope, Sidechain, Threshold, Knee, Default

Digitalone1:
@wwmm do you plan to make a new release soon? I'm updating the documentation only in English and I think it will be ready in the next week.

It could be implemented in the next release.

wwmm:
>do you plan to make a new release soon?

It does not have to be soon. So far we have only new features and changes for potential bugs that could happen in very specific situations. So the next release can wait for the documentation update.

[...]

Issue #1231, opened 2021-10-31
Typing values into spinboxes doesn't stick


limejet:
### EasyEffects Version

6.1.3

### What package are you using?

Arch Official (easyeffects)

### Distribution

Manjaro

### Describe the bug

When a plugin has a value that can be changed using a numerical input with +/- buttons, typing a value in manually doesn't "stick". Pressing enter or moving focus away form the input causes the value inside to reset to what it was before taking focus. This seems to happen with all the default plugins.

What's interesting is that changing the value does actually work; typing something does change the sound. However, when the value resets, so does the effect it controls.

### Expected Behavior

When pressing enter in a spinbox, the typed value should remain.

### Debug Log

There doesn't seem to be anything relevant in the log, but i've included one anyway. The actions taken here were

1. launch `easyeffects`
2. go to "Input" tab
3. select "Plugins"
4. select a predefined "Stereo tools" plugin
5. click inside the "Balance" spinbox
6. type new value in
7. press enter
8. close application

<details>
<summary>Log</summary>

```
(process:32041): easyeffects-DEBUG: 10:33:04.987: main: locale directory: /usr/share/locale
(easyeffects:32041): easyeffects-DEBUG: 10:33:05.027: presets_manager: system input presets directory: "/etc/easyeffects/input"; 
(easyeffects:32041): easyeffects-DEBUG: 10:33:05.027: presets_manager: system input presets directory: "/etc/xdg/easyeffects/input"; 
(easyeffects:32041): easyeffects-DEBUG: 10:33:05.027: presets_manager: system output presets directory: "/etc/easyeffects/output"; 
(easyeffects:32041): easyeffects-DEBUG: 10:33:05.027: presets_manager: system output presets directory: "/etc/xdg/easyeffects/output"; 
(easyeffects:32041): easyeffects-DEBUG: 10:33:05.027: presets_manager: user presets directory already exists: /home/jet/.config/easyeffects/
(easyeffects:32041): easyeffects-DEBUG: 10:33:05.027: presets_manager: user presets directory already exists: /home/jet/.config/easyeffects/input
(easyeffects:32041): easyeffects-DEBUG: 10:33:05.027: presets_manager: user presets directory already exists: /home/jet/.config/easyeffects/output
(easyeffects:32041): easyeffects-DEBUG: 10:33:05.027: presets_manager: user presets directory already exists: /home/jet/.config/easyeffects/autoload/input
[...]

PR #1235, opened 2021-11-03
English only Documentation updated

Digitalone1:
Related to #1108

Documentation updated. Translations deleted.

I will do some other corrections in the next days, but most of the work is done.


Digitalone1:
`prepare_spinbutton` makes a check on lower/upper end.

Related to #1231

PR #1238, opened 2021-11-05
minor equalizer.page English grammar fixes

AlexFolland:
"The user can choose between 1 and 32 bands" would actually mean that the user can choose only 1 or 32 bands, and nothing between them, despite using "between".  "The user can choose from 1 to 32 bands" is inclusive, meaning the user can choose any number of bands from 1 through 32.

When the noun is countable, "fewer" should be used instead of "less".


wwmm:
Thanks!

PR #1239, opened 2021-11-05
Remember the window maximized state

Digitalone1:
I had issues with EE window because sometimes it wasn't saving the window size across sessions.

Then I discovered that GTK does not get the width/height size when the window is maximized, so the previous unmaximized size values are saved in gsettings on close. Therefore when I close EE maximized, at the next window opening I find it at smaller sizes. 

Don't know why this is happening, anyway I tried to find a workaround. We can save the maximized state on close, so we apply it on window showing. Tested and It's working to me.

I wonder if it works also outside Gnome environment. I hope so, it should be tested.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,12 +4,17 @@
 
 ### Added
 
-- It is now possible to combine impulse responses in the convolver interface. A new impulse file is generated and it
+- It is now possible to combine impulse responses in the Convolver interface. A new impulse file is generated and it
   should be visible in the impulse list.
 - Improved `x axis` drawings in our plots. Now the number of labels is adjusted dynamically depending on our window
   width.
 - The documentation has been updated reflecting the new EasyEffects features. Old references about PulseEffects
   have been removed.
+
+### Fixed
+- When a spinbutton is filled with an out of range value, now it is updated with the lowest/highest value rather than
+  resetting to the previous one.
+- The application window now remembers the maximized state and restores it on the next opening event.
 
 ### Note to packagers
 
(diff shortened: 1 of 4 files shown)


wwmm:
Strange. Looking at the code it seemed fine but somehow it does not work for me. Window still has the non maximized size.

wwmm:
And now it is working... Weird...

wwmm:
It seems that `is_maximized` can have some surprises
Screenshot from 2021-11-05 15-43-21
But I am not sure if there is a better alternative.

wwmm:
I think that because I closed the window as soon as I maximized it the wrong state was saved. And that is why it did not work in my first attempts. I would not bother with this. Under usual circumstances the user will stay for  some time with the window maximized.

As I never use EasyEffects with its window maximized I did not even notice there was a problem with the window size restoration in this case.

Digitalone1:
Maybe it does not save the maximized window size on Wayland, but it works on X11?

I didn't try it. Can you do a test? Usually GTK apps remember the maximized state (I use browsers like Firefox and Brave always maximized) and as far as I remember I didn't have these issues when using Xorg in the past.

There should be something preventing to get the size of the window when it's maximized on Wayland, maybe because the window manager is positioning the window and not the user itself, I don't know. Anyway saving the maximized state is a good workaround.

Digitalone1:
Oh, now I remember, Gnome Terminal does not save the maximized state. But I use another terminal application, Tilix, and it does not save the maximized state too.

wwmm:
>Maybe it does not save the maximized window size on Wayland, but it works on X11?

I did not think about that. I will try to remember to do tests on Xorg. But I agree that the current workaround is good enough and it won't do any harm to keep it there.

PR #1241, opened 2021-11-07
User Interface improvements

Digitalone1:
Various UI refinements.

Phase Correlation level meter removed from Stereo Tools since it's not working.

Documentation updated.


Digitalone1:
Added help documentation button.

Implemented filename sanity check for IR combination as it's done for preset creation.

PR #1247, opened 2021-11-12
Multiband Gate: Make clear Reduction is Max Reduction

jwatt:
It took me a frustratingly long time to figure out how the Multiband Gate plugin actually works even after finding the Calf documentation. A key part of my confusion was the labeling of the "Reduction" control. If this change is acceptable, hopefully it will help future new users from having a similarly frustrating experience.

### DIFF
--- a/data/ui/multiband_gate.ui
+++ b/data/ui/multiband_gate.ui
@@ -426,7 +426,7 @@
 
                                         <child>
                                             <object class="GtkLabel">
-                                                <property name="label" translatable="yes">Reduction</property>
+                                                <property name="label" translatable="yes">Max Reduction</property>
                                                 <layout>
                                                     <property name="column">6</property>
                                                     <property name="row">0</property>
@@ -453,7 +453,7 @@
                                                     <property name="row">1</property>
                                                 </layout>
                                                 <accessibility>
-                                                    <property name="label" translatable="yes">Band 1 Reduction</property>
+                                                    <property name="label" translatable="yes">Band 1 Max Reduction</property>
                                                 </accessibility>
                                             </object>
                                         </child>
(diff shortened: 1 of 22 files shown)


wwmm:
I think it can be done.

wwmm:
I forgot to merge :smile:

jwatt:
I was just wondering. 😄 Thank you!

wwmm:
This is a plugin I almost never use. Let me know if it would be interesting to have the multiband gate from Linux Studio Plugins. We use the multiband compressor from them because Calf has problems in its crossover implementation. I imagine the same problems are present in their multiband gate. But as there is always something else at the top of my priority list I did not try to switch to the one from LSP yet.

jwatt:
I'm only just starting to explore this space so I currently have no idea if that would be preferable.

I'm not sure what you're referring to when you say "problems in its crossover implementation", but one issue I've noticed with the Multiband Gate is that if I set all bands to completely block (attack=10, release=10, threshold=0, ratio=20, knee=0, makeup=0, reduction=-95), whenever I bypass one of the bands there seems to be significant leakage of audio through/onto the other "blocked" bands. At least that's what the frequency response bars at the top of the GUI show. For example, if the Mid Band is set to span 3,000 Hz to 5,000 Hz and only it is bypassed, the bars at the top show output from about 1,200 Hz to 7,000 Hz. It is shaped a bit like a bell chart, so I wonder if this is intentional to avoid sharp cutoff at the set frequencies. However, for my purposes I'm trying to use the Multiband Gate to figure out which frequencies I produce "sss" sibilance on, so it's not what I personally want. :-)

wwmm:
>I'm not sure what you're referring to when you say "problems in its crossover implementation"

Its crossover module is used to split the signal in different frequency bands. It seems there are flaw in how that splitting is being done.

PR #1298, opened 2021-12-18
Add Flatpak manifest

vchernin:
So you can clone and start building with gnome builder without dependency hunting.
Note `org.freedesktop.LinuxAudio.Plugins.LSP` and `org.freedesktop.LinuxAudio.Plugins.ZamPlugins` are not automatically installed, so plugins like the convolver do not work after building in gnome builder.
- Maybe there are ways of workarounding, see https://gitlab.gnome.org/GNOME/gnome-builder/-/issues/1586

Manually copied 4 modules from flathub shared-modules/linux-audio to avoid a git submodule.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,3 +52,18 @@ debian/*
 
 ppastats
 .autobuild
+
+# exclude most of flatpak directory
+util/flatpak/*
+!util/flatpak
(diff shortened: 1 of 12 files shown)


vchernin:
The previous PR the gitignore was slightly incorrect, but it should be correct now

PR #1301, opened 2021-12-22
Danish Translation

Grooty12:
I haven't done this before so not sure I'm doing it right, but I translated EasyEffects to Danish.


wwmm:
Thanks!

PR #1307, opened 2022-01-02
Use NEWS file to track release information

vchernin:
This mostly works as intended. The idea is to use the news file to track release info, and generate both metainfo data and a formatted markdown changelog from news.

Reasons for draft:
- [x] Need to make script more robust (second checkbox needs testing)
- [x] https://github.com/wwmm/easyeffects/commit/04bd3e968f7e0ba79b9672882590affe445dc330 exists to find the current date and update the metainfo with the date. appstreamcli already takes care of updating the metainfo using the release date in the news file. Is there still a need for this script? It could be repurposed to find the current date, and put it in the NEWS file instead.
- [x] Ensure appstreamcli's changes don't break various distro's appstream-glib validation (there are two somewhat incompatible appstream implementations)
- [x] Right now I've only kept releases since 6.0.0 in the news file. Is there a need to bring over older releases as well? If not, should the old changelog be kept somewhere?


wwmm:
>Is there still a need for this script?

Honestly I do not know. As I do not use Flatpak I have been running this script before doing a release assuming we still need it. But I did not test.

wwmm:
>Right now I've only kept releases since 6.0.0 in the news file. Is there a need to bring over older releases as well? If not, should the old changelog be kept somewhere?

I doubt that the public target by Flatpak needs to see the chagelog of older EasyEffects or even PulseEffects releases. I also doubt we have to keep that info somewhere.

vchernin:
> Honestly I do not know. As I do not use Flatpak I have been running this script before doing a release assuming we still need it. But I did not test.

Keep in mind this metainfo file and script are not specific to Flatpak at all. Flatpak happens to want a metainfo file, but metainfo is used by most other package formats as well. So running this script has had benefits to non Flatpak users too.

I think the script has been doing it's job. Basically, you running the script when doing releases meant the correct date, along with the corresponding version number are put in the metainfo file.

Since the only useful functionality from that older script now is the date grabbing, it makes to ask if we still need that.

Would you like to manually add the date to the news file (i.e. just follow the format and manually put in the current date whenever it's release time). Or get a script to automatically insert the correct date in news when it's released?

> I doubt that the public target by Flatpak needs to see the chagelog of older EasyEffects or even PulseEffects releases. I also doubt we have to keep that info somewhere.

Yeah, especially with the rename.

vchernin:
So the overall release process would be:

1. Update meson version number.
2. Add release date and version number to news.
3. Run new script.
4. Commit, tag, push.

(I think that's basically everything?)

wwmm:
>I think that's basically everything?

I think it is.

vchernin:
> I doubt that the public target by Flatpak needs to see the chagelog of older EasyEffects or even PulseEffects releases. I also doubt we have to keep that info somewhere.

To be more specific on this, all 3 files (news, changelog, metainfo) currently only have releases from 6.0.0 onwards. This isn't just about the public target of Flatpak, this change as-is affects anyone who looks at the version history.

I could easily put `PulseEffects-Changelog.md` somewhere with the older changelog in util, so people can see the older history. That file would have all release info < 6.0.0. But I guess the old PulseEffects branch will always exist.

wwmm:
>But I guess the old PulseEffects branch will always exist so maybe it's not needed.

Yes. I do not have a reason to delete it. Looking at things this way there is not need to make a copy of the old changelog.

vchernin:
So this should be ready now. To confirm it works yourself you can just run the script without changing anything in this branch. It will produce a release for 6.2.0 like this test commit:
https://github.com/vchernin/easyeffects/commit/84618ed75d2c5bf6b0600d789b901571efe51bda

Note this script uses both `appstream-util` and `appstreamcli` for metainfo linting. `apptreamcli` gives warnings but they're unimportant (and is actually appstreamcli's fault).

As with the older script, for all this to make sense the script needs to be run every release. When updating the NEWS file keep the formatting as close as possible to the template given, otherwise appstreamcli might complain.

wwmm:
As far as I can see it works fine on my computer. @Digitalone1 the changelog from the repository and the one for Flatpak are now generated from the `util/NEWS` file.

Digitalone1:
Nice. Tomorrow I should have the time to translate and the new release can be done.

PR #1313, opened 2022-01-04
Translations for 6.2.0

Digitalone1:
Regeneratng the templates the first time, I got an error because some files were deleted.

So I updated the `POTFILES.in`. Give a look at it if something is missing. I removed `data/com.github.wwmm.easyeffects.metainfo.xml.in` from the list because I think it's not really needed to translate it. Removed also `translatable` property from some labels that do not need the translation.


vchernin:
I think it would be ideal to also translate at least the title and description in `data/com.github.wwmm.easyeffects.metainfo.xml.in`. I have seen them work in the past. That text is what shows up in gnome software, kde discover, etc.

Now that there are release notes in that file too, technically those can also be translated. But that will be a lot of translating to do (and somehow the translation needs to be done after the release script has been run...)

In any case I think those issues can be revisited later.

Digitalone1:
I excluded it because I saw no point translating the changelogs recently added.

Isn't there any other method to show translations in app management system like gnome software and discover?

@wwmm please, try to reinclude the file and add `translatable="no"` property to xml tags to see if they can be excluded from translation templates.

vchernin:
> Isn't there any other method to show translations in app management system like gnome software and discover?

Well, one way or another you'd need to provide translations for the metainfo file. It is the standard way of presenting data about apps to app management systems.

I found:
https://www.freedesktop.org/software/appstream/docs/sect-Quickstart-Translation.html


At least for release notes (not necessarily the other parts of metainfo) there is a discussion about separating the release notes from the metainfo file. Release notes could be updated separately which would make translations easier, since right now we'd need a ton of translations just before every release.

So I think short term translations for just the metainfo title and description are fine, translating release notes might be challenging without being able to update release notes separately.

wwmm:
>please, try to reinclude the file and add translatable="no" property to xml tags to see if they can be excluded from translation templates.

I reincluded the file but I decided to let the `translatable="no"` test for another moment. The `6.2.0` release would not have translation for that anyway. Translating the hole changelog is probably not worth the effort. We have to see how to keep only what is not going to add to the translation burden.

PR #1318, opened 2022-01-06
Improve news script

vchernin:
Fixes the news script a bit.

I also took this opportunity to take new screenshots since the current ones are out of date now. The readme now uses them, and shows dark ones when in github dark mode (which can sync with your browser).

- Much better AppStream error handling
- Fails fast on errors
- Doesn't make changes if failure (exit early and use temp files before finished)
- Use temp files properly (at least in cases where appstream-util doesn't complain...)
- Sane way of regenerating older releases (asked at script initial run)


vchernin:
This should be ready now.

PR #1347, opened 2022-01-16
Acquire realtime priority on PipeWire

ishitatsuyuki:
Due to a poorly documented behavior [1] in PipeWire, EasyEffects was
running under normal scheduling policy which lead to an excessive amount
of audio dropouts.

Fix this by requesting the realtime client config profile.

[1] https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/2024

### DIFF
--- a/src/pipe_manager.cpp
+++ b/src/pipe_manager.cpp
@@ -1343,6 +1343,7 @@ PipeManager::PipeManager() {
 
   pw_properties* props_context = pw_properties_new(nullptr, nullptr);
 
+  pw_properties_set(props_context, PW_KEY_CONFIG_NAME, "client-rt.conf");
   pw_properties_set(props_context, PW_KEY_MEDIA_TYPE, "Audio");
   pw_properties_set(props_context, PW_KEY_MEDIA_CATEGORY, "Manager");
   pw_properties_set(props_context, PW_KEY_MEDIA_ROLE, "Music");


wwmm:
I was not expecting something like that to happen... Could it be that the filters were actually getting realtime priorities and that what `client-rt.con` is actually doing is applying it to the pipewire main loop on the client side?

In any case I doubt there would be any harm in accpeting the PR. Let me just do some tests on my computer.

ishitatsuyuki:
> Could it be that the filters were actually getting realtime priorities and that what client-rt.conf is actually doing is applying it to the pipewire main loop on the client side?

Filters are executed inside EasyEffects process, not the PipeWire process. The RT client config needs to be applied in order to elevate the EasyEffects filter thread to realtime.

Forward any complaints to the linked PipeWire issue :stuck_out_tongue: I also don't like how this PipeWire behavior is a thing.

After dogfeeding this patch for a few hours I can confidently say that the xruns are gone. I hope it works well on your system too.

wwmm:
I looked at the output of `chrt -ap` and there are no thread scheduled as realtime indeed. There are some scheduled as `SCHED_BATCH` but they could be gtk threads. Well...

And to make things even more bizarre a `pw_filter` can't eve work without `PW_FILTER_FLAG_RT_PROCESS` https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/1016. So in my mind that was a confirmation that the filters were running in a realtime thread and there was "no need for further investigation" =/

Issue #1089, opened 2021-08-11
PipeWire version and backwards compatibility


vchernin:
I am wondering how the PipeWire version should/will be updated for EasyEffects. This is (mostly) a Flatpak issue, but it does somewhat affect distro packages.

**Context:**

EasyEffects currently depends on PipeWire 0.3.31 at build-time. This is fine for Arch, Fedora and other rapidly updating distros. However, when distros like Ubuntu or Debian likely switch to PipeWire by default it is less likely newer versions of PipeWire will be present. In fact, Ubuntu 21.04 ships with 0.3.24, and Debian 11 with 0.3.19. Both far too old to build or likely run EasyEffects.

With Flatpak we can arbitrarily bundle any PipeWire version so long as it builds. However, I have noticed that bundling and building with 0.3.33 (from Flatpak) and running 0.3.32 (From Fedora host) caused EasyEffects to just not start. A user running dietpi (based off Debian 10 with PipeWire 0.2.x) also ran into this. So Flatpak doesn't really solve the outdated distro PipeWire problem, even if it does offer an alternative to distro EasyEffects packaging.

**Question:**

I am wondering how often EasyEffects plans to update the minimum build-time PipeWire version. Since distros like Ubuntu or Debian might have an older PipeWire version, the less the build-time PipeWire dependency is updated the more distros will hopefully be able to use EasyEffects.

I am also wondering if there is a way to give a very straightforward error if the host PipeWire version is too old. This is somewhat more of a Flatpak problem, since traditional package managers (shouldn't) let users install say the Arch EasyEffects package with PipeWire 0.3.4.

I don't (yet) know of a reliable way to stop the Flatpak from installing on distros with too old PipeWire. Perhaps what would help is if EasyEffects had a fallback instead of just not launching when at runtime PipeWire is out of date. Since most users won't read command line logs, perhaps a pop-up saying "Your PipeWire version is out-of-date, EasyEffects requires PipeWire x.y.z. or newer."

P.S. I think the Flatpak is otherwise mostly ready, I'd appreciate if you could look over [the current issue list](https://github.com/flathub/flathub/pull/2429#issue-696395012) and mention if any should block EasyEffects' release onto Flathub stable.

vchernin:
Ok, so I've discovered something that might be useful.

With the Flatpak, you can enter the sandbox where EasyEffects is with:
`flatpak run --command=sh --devel com.github.wwmm.easyeffects`

From there you can run commands like `pipewire --version`

[...]

PR #1421, opened 2022-03-07
Set target.object metadata in addition to target.node

pv:
Pipewire session managers will prioritize target.object metadata, which
uses object.serial values instead of ids.

--------

There's a plan to soft-deprecate target.node, to address the issue that id values are not good for long-term references, since they are reused. Setting target.node will still continue working, but for forward compatibility it's better to also set the other value.

### DIFF
--- a/src/pipe_manager.cpp
+++ b/src/pipe_manager.cpp
@@ -1044,6 +1044,10 @@ void on_registry_global(void* data,
         nd->nd_info->media_class = media_class;
         nd->nd_info->name = name;
 
+        if (const auto* serial = spa_dict_lookup(props, PW_KEY_OBJECT_SERIAL)) {
+          nd->nd_info->serial = std::stoull(serial);
+        }
+
         if (const auto* node_description = spa_dict_lookup(props, PW_KEY_NODE_DESCRIPTION)) {
           nd->nd_info->description = node_description;
         }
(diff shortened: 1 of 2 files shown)


wwmm:
Thanks @pv. I was not aware of the serial key. How is it going to interact with device suspending? I had to put some workarounds in place because the id does not persist when the computer comes back from suspending.

wwmm:
>How is it going to interact with device suspending?

I imagine that the serial will survive that. And what will happen when headphones are plugged in the onboard card jack? The serial will change when the profile changes?

pv:
The serial behaves vs. suspend the same way as id, but with the guarantee that once the object it refers to is gone, the same value will not be reused for a different object. So you don't need to explicitly go around invalidating object references, when objects disappear.

If nodes disappear and reappear, metadata using ids may direct streams to essentially random targets since ids are reused (in order of appearance, not deterministic). With serials, the old metadata always points to non-existent target, and the session directs such streams to the default sink.

For longer-term references, one can put the node.name value into target.object, and then it is not tied to a specific pipewire object instance.

src/pipe_manager.cpp
```diff
@@ -1044,6 +1044,10 @@ void on_registry_global(void* data,
         nd->nd_info->media_class = media_class;
         nd->nd_info->name = name;
 
+        if (const auto* serial = spa_dict_lookup(props, PW_KEY_OBJECT_SERIAL)) {
```
vchernin, line 1047: Not sure if it's really a bug, but when compiling the flatpak with PW 0.3.31 I got:
```
../src/pipe_manager.cpp: In function ‘void {anonymous}::on_registry_global(void*, uint32_t, uint32_t, const char*, uint32_t, const spa_dict*)’:
../src/pipe_manager.cpp:1047:57: error: ‘PW_KEY_OBJECT_SERIAL’ was not declared in this scope; did you mean ‘PW_KEY_DEVICE_SERIAL’?
```

Bumping the bundled PW lib to 0.3.47 fixed it. 0.3.31 was bundled in the past for better compatibility, but things have [improved](https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/1515) upstream. Perhaps I will reinvestigate a better solution for https://github.com/wwmm/easyeffects/issues/1089...

PR #1439, opened 2022-03-19
Set last-used-x-preset when changed from CLI

markonius:
Hi,

if preset was loaded via CLI, the GUI didn't show the change and checking the selected preset via dconf yielded the wrong result.
Please let me know if there are some contributing procedures I should follow, I couldn't find any.


wwmm:
Thanks! As I do not use command line options I did not notice this problem.

PR #1456, opened 2022-03-31
flatpak: fix flatpak build

1player:
There's a patch mentioned in the `util/flatpak/com.github.wwmm.easyeffects.json` manifest that's not included in the source tree, and without it the flatpak can't be built. 

I've copied it from the https://github.com/flathub/com.github.wwmm.easyeffects repository.

EDIT: in fact the manifest is outdated, this PR is to fix the flatpak build in its entirety.


vchernin:
the patch is there, but it has a different name and is out of date. Since copying it over here is manual (and since gnome builder ignored flatpak source patches last I checked), I don't frequently update it.

Although, in the not-too-distant future, it (should) be possible for Flathub apps to avoid the https://github.com/flathub/com.github.wwmm.easyeffects/ repo entirely, and just push it from their own repo. So, all the flatpak stuff would only live in one place.

vchernin (approved):
Please also make sure to delete the old version of the patch `003-libportal-autostart.patch`

1player:
I also see `001-fix-about-page-icon.patch` is unused, shall I remove that one as well?

vchernin:
> I also see `001-fix-about-page-icon.patch` is unused, shall I remove that one as well?

Ah yeah, you can remove it, thanks!

1player:
PTAL, I've imported the manifest from Flathub (sadly that one is in YAML format, so I completely nuked the old JSON one), fixed the patches and now the build works.

Issue #1451, opened 2022-03-28
auto-start no audio until switch source 


jkoking:
### EasyEffects Version

6.2.4

### What package are you using?

Arch (easyeffects)

### Distribution

Arch Linux

### Describe the bug

When I boot up my computer with auto start enabled audio input and output doesn't work
I fix this temporarily by switching to my audio devices and then back to my easyeffects source's 

### Debug Log

<details><summary>Debug Log</summary><br/>

```
(process:3636): easyeffects-DEBUG: 13:12:23.214: easyeffects version: 6.2.4
(process:3636): easyeffects-DEBUG: 13:12:23.216: main: locale directory: /usr/share/locale
(easyeffects:3636): easyeffects-DEBUG: 13:12:23.218: presets_manager: system input presets directory: "/etc/easyeffects/input"; 
(easyeffects:3636): easyeffects-DEBUG: 13:12:23.218: presets_manager: system input presets directory: "/etc/xdg/easyeffects/input"; 
(easyeffects:3636): easyeffects-DEBUG: 13:12:23.218: presets_manager: system input presets directory: "/home/jacob/.config/kdedefaults/easyeffects/input"; 
(easyeffects:3636): easyeffects-DEBUG: 13:12:23.218: presets_manager: system output presets directory: "/etc/easyeffects/output"; 
(easyeffects:3636): easyeffects-DEBUG: 13:12:23.218: presets_manager: system output presets directory: "/etc/xdg/easyeffects/output"; 
(easyeffects:3636): easyeffects-DEBUG: 13:12:23.218: presets_manager: system output presets directory: "/home/jacob/.config/kdedefaults/easyeffects/output"; 
(easyeffects:3636): easyeffects-DEBUG: 13:12:23.219: presets_manager: user presets directory already exists: /home/jacob/.config/easyeffects/
(easyeffects:3636): easyeffects-DEBUG: 13:12:23.219: presets_manager: user presets directory already exists: /home/jacob/.config/easyeffects/input
(easyeffects:3636): easyeffects-DEBUG: 13:12:23.219: presets_manager: user presets directory already exists: /home/jacob/.config/easyeffects/output
(easyeffects:3636): easyeffects-DEBUG: 13:12:23.219: presets_manager: user presets directory already exists: /home/jacob/.config/easyeffects/autoload/input
[...]

PR #1458, opened 2022-03-31
Added audio sources warning

jkoking:
EasyEffects was my first experience with real time audio processing when I installed it I assumed that the virtual  devices needed to be set as default

#1451

I thought that this change might prevent future issue reports


wwmm:
Thanks! As some corner cases do not have a choice other than set our devices as default this was never said explicitly in our readme. But it is the better thing to do.

Issue #1316, opened 2022-01-05
Plans for EasyEffects 7.0


wwmm:
List of features planned for EE `7.0`:

- Support for multiple plugins instances (it is going to be a tough one)
- Finding a way to make the spectrum and level meters update frequency independent of the server latency value
- Adding dry/wet controls to some plugins

wwmm:
In order to support multiple instances we will need to identify these instances in some way so we can bing them to the correct path in the GSettings database. One thing we could something like this
```
"plugins_order": [
            "bass_enhancer#0",
            "compressor#0",
            "exciter#0",
            "compressor#1",
            "autogain#0",
        ]
```
And each plugin section would be named in the presets json file like 
```
"limiter#2": {
...
}
```

In any case I think that thework on multiple instances has to start on these two files 
```
https://github.com/wwmm/easyeffects/blob/master/src/effects_base.cpp
https://github.com/wwmm/easyeffects/blob/master/src/plugins_box.cpp
```
The reason is that there won't be the need to think about the presets format while working on them. It should be possible to make the plugins map to be populated on demand as things are right now. Once that works we can start to think about multiple instances.

Digitalone1:
Another issue is how the multiple instances are saved in gsettings.

At this point I wonder if gsettings is really needed. I know it's advertised as faster for config management, but if we have to handle json files for loading/saving presets, may we use it the same to store the last config on the system and read/write directly from them?

wwmm:
>At this point I wonder if gsettings is really needed. I know it's advertised as faster for config management, but if we have to handle json files for loading/saving presets, may we use it the same to store the last config on the system and read/write directly from them?

The json files for the presets are just one thing among many. Json does not allow us to bind settings directly to the widgets in the interface. I doubt we would gain simplicity by not using gsettings. We would probably be reinventing the wheel.

And we would have a lot more couping between different sections of our code. All those presets classes would have to either be passed to other places or receive instances from other places.

And without GSettings third party scripts or programs like dconf-editor won't be able to change our settings anymore. We will lose more than we will gain(if we gain something) by not using gsettings.

wwmm:
>Another issue is how the multiple instances are saved in gsettings.

With relocated schemas already in place the part related to gsettings is not going to be so hard. Consider this for example https://github.com/wwmm/easyeffects/blob/cd8967b53ec35203384d183fd4816d36c15ea24d/src/autogain_preset.cpp#L26
[...]

PR #1461, opened 2022-04-01
UI improvements

Digitalone1:
Some fixes and improvements. Some of them trying to address what exposed [here](https://github.com/wwmm/easyeffects/issues/1316#issuecomment-1065931764).

- App Info: improved middle dot spacing and spacing between checkbuttons; text centered in volume spinbutton.
- App Info: app name and description are now selectable to be copied in the clipboard (thought it could be useful to the user and the app name could be pasted in the excluded list).
- App Info: `xalign` set to `0` is fixing the blur effect GTK bug on wrapped label and also improving the alignment on wrapped lines. 
- Plugin word replaced by Effect (more intuitive and follows the application name).
- Preset Autoload: labels of parameters have ellipsize property, so the delete button should not be hidden on low width window.
- Blocklist: The UI now exposes it like "Excluded Applications", no Blocklist anymore; we keep calling it as blocklist in the code.
- Excluded Apps Menu: the app name row now has ellipsize and max-width-char properties keeping the popover menu at reasonable width in case of extreme labels.
- Excluded Apps Menu: Show Exclude Applications changed as a checkbutton; The switch seems to big there. 
- Presets Menu: changed Microphone/Speakers in Input/Output, following Input/Output buttons in headerbar.

Edit: Tomorrow I should add the new translations.

### DIFF
--- a/data/ui/app_info.ui
+++ b/data/ui/app_info.ui
@@ -87,11 +87,12 @@
                                                     </object>
                                                 </child>
 
-                                                <!-- middle dot -->
+                                                <!-- Middle dot -->
                                                 <child>
                                                     <object class="GtkLabel">
                                                         <property name="halign">start</property>
-                                                        <property name="margin-end">5</property>
+                                                        <property name="margin-start">4</property>
+                                                        <property name="margin-end">4</property>
                                                         <property name="label">·</property>
                                                         <style>
                                                             <class name="dim-label" />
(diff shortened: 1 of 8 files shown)


Digitalone1:
@wwmm Weblate is not updated yet with the new strings.

wwmm:
I think this time it is because it does not have 24 hours since the last pending operation.

wwmm:
Let's see what happens in the next hours. Weblate worked as expected in the last 2 or 3 days.

Digitalone1:
Still no update. I think a new po regeneration should be submitted to Weblate.

wwmm:
>Still no update. 

There was one 7 hours ago. It is taking longer than expected but somehow it did the update by itself.

Digitalone1:
I'm talking about the update to show the new stings in Weblate to be translated. The ones from the last commits `Effects`, `Excluded Apps`, etc...

wwmm:
>I'm talking about the update to show the new stings in Weblate to be translated. The ones from the last commits Effects, Excluded Apps, etc...

Ok. I did not check the contents. At this moment there is no pending operation in weblate. I will update the template and do a commit. Let's see what will happen.

Issue #1479, opened 2022-04-17
Equalizer's "Load APO Preset" feature is locale dependent


vhborges:
### EasyEffects Version

6.2.4

### What package are you using?

Arch (easyeffects)

### Distribution

ArcoLinux (Arch Linux based)

### Describe the bug

The "Load APO Preset" feature in the Equalizer plugin is dependent on the system's Numeric Locale.

My OS is configured to use the pt-BR Numeric Locale, if I try to read an eq text file on EasyEffects it ignores all the decimal digits and just reads the integer part. When I change my locale to en-US this feature works as expected.

Steps to reproduce:

1. Change your system's numeric locale to pt-BR (restart if needed);
2. Launch EasyEffects, go to plugins > Add Plugin > Equalizer > Load APO Preset;
3. Save the text below to a file and read it on EasyEffects:
4. See the result values of Quality, Gain and Input, all of them were rounded down, the decimal parts were ignored;
5. Repeat the steps above in the en-US numeric locale, all the values are now correct.

> Preamp: -5.5 dB
Filter 1: ON PK Fc 54 Hz Gain -2.3 dB Q 1.0
Filter 2: ON LS Fc 104 Hz Gain 5.5 dB Q 0.71
Filter 3: ON PK Fc 140 Hz Gain -1.4 dB Q 0.7
Filter 4: ON PK Fc 1350 Hz Gain -3.3 dB Q 1.5
Filter 5: ON PK Fc 1830 Hz Gain 2.4 dB Q 1.3
Filter 6: ON PK Fc 3230 Hz Gain -1.4 dB Q 4.0
Filter 7: ON PK Fc 4420 Hz Gain -3.1 dB Q 4.5
Filter 8: ON PK Fc 5600 Hz Gain -4.1 dB Q 2.5
Filter 9: ON PK Fc 7000 Hz Gain 2.5 dB Q 1.2
Filter 10: ON HS Fc 10000 Hz Gain -1.0 dB Q 0.71

### Expected Behavior

The application shouldn't be locale dependent. The Load APO Preset feature should interpret the decimal values in the text file.

### Debug Log

<details><summary>Debug Log</summary><br/>

```
(process:3803): easyeffects-DEBUG: 18:43:14.322: easyeffects version: 6.2.4
(process:3803): easyeffects-DEBUG: 18:43:14.324: main: locale directory: /usr/share/locale
[...]

PR #1480, opened 2022-04-18
Locale independent string/number conversions

### LINKED ISSUE #1484: Issue with locale and spinbuttons
When locale is set different than C, no input over `999` can be inserted in spinbuttons.

I tested this, but did not for numbers over `999`. I suspect `1.000` is interpreted as `1`. I will investigate.

Digitalone1:
Related to #1479 

It's working on my system and APO parsing is fixed.

Anyway this has to be tested for some time on different systems.

I also noticed an issue, but I'll open a new ticket for it.

Edit: Fixes #1484

### DIFF
--- a/include/convolver.hpp
+++ b/include/convolver.hpp
@@ -94,7 +94,7 @@ class Convolver : public PluginBase {
       const int& ret = conv->process(true);  // thread sync mode set to true
 
       if (ret != 0) {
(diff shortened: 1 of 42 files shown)


Digitalone1:
@wwmm Can you please test if there's memory leak with these new features?

PR #1498, opened 2022-04-25
Spectrum: add margin to labels

Digitalone1:
As I said in a previous post, since GTK widgets have a native padding, it would be better to not stick the spectrum labels to the window borders.

There's a `set_margin` method in the spectrum class, but I thought it would add the margin to the chart also while I want it only on the labels.

I simply ended up adding whitespaces.


Digitalone1:
Also, we do not use `data/icons/scalable/emblems/ee-plugins-symbolic.svg` anymore. Should we delete it?

PR #1503, opened 2022-04-30
Default app icon and Effects adw status page

Digitalone1:
- `applications-multimedia-symbolic` set as default visible app icon if the provided icon is not available in the system icon theme.
- `AdwStatusPage` added into plugin box ui to show the empty status when no plugins are placed in the list.

With these two I think a new release could be prepared, but first the translations have to be updated and the default app icon has to be tested on Flatpak because my icon theme does not show it.

@wwmm Regarding the status page for plugins, I tried to reproduce what you did for app ui, but I may forget something, so please check it out.

### DIFF
--- a/data/ui/app_info.ui
+++ b/data/ui/app_info.ui
@@ -28,7 +28,7 @@
                                     <object class="GtkImage" id="app_icon">
                                         <property name="halign">start</property>
                                         <property name="valign">center</property>
-                                        <property name="icon_size">large</property>
+                                        <property name="icon-size">large</property>
                                     </object>
                                 </child>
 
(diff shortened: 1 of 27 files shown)


wwmm:
Good work! Thanks! I had these tasks in my mind but with the start of the new academic semester at my work and the usual problems and classes that come with it it could have taken a lot of time until I could look at this.

>but first the translations have to be updated

I will update the translation template now.

>the default app icon has to be tested on Flatpak because my icon theme does not show it.

If this become a problem we can always consider to embed this one like it was done to some of the others. Talking about that I will remove https://github.com/wwmm/easyeffects/blob/master/data/icons/scalable/emblems/ee-plugins-symbolic.svg as it is not used anymore.

Digitalone1:
@vchernin can you test this?

wwmm:
>I think a new release could be prepared

I will do it tomorrow morning just in case some last moment change is needed. I have already updated the translation template.

vchernin:
> @vchernin can you test this?

image

Using VLC flatpak build, looks good!

Digitalone1:
> > I think a new release could be prepared
> 
> I will do it tomorrow morning just in case some last moment change is needed. I have already updated the translation template.

Wait for me to update the Italian translation. I will notify when it's completed.

wwmm:
>Wait for me to update the Italian translation. I will notify when it's completed.

Ok.

PR #1507, opened 2022-05-01
Plugin stack: Start/endpoint labels

Digitalone1:
This will be more user friendly and will resolve the longstanding issue of not understanding where the stream comes from and where it ends up.

Maybe it has to be refined since `navigation-sidebar` class isn't useful on GtkBox, so we have to set manual margins to align the labels to the effects. On my system it's quite aligned, but I don't know if on other systems with different resolutions it's the same.

It can be work out, but it's a good starting point anyway.

### DIFF
--- a/data/easyeffects.gresource.xml
+++ b/data/easyeffects.gresource.xml
@@ -54,7 +54,6 @@
         <file>ui/shortcuts.ui</file>
         <file>ui/stereo_tools.ui</file>
         <file>icons/scalable/emblems/ee-arrow-down-symbolic.svg</file>
-        <file>icons/scalable/emblems/ee-square-symbolic.svg</file>
         <file>icons/scalable/emblems/ee-spectrum-symbolic.svg</file>
         <file>icons/scalable/emblems/ee-drag-handle-symbolic.svg</file>
     </gresource>
--- a/data/icons/scalable/emblems/ee-square-symbolic.svg
+++ b/data/icons/scalable/emblems/ee-square-symbolic.svg
@@ -1,1 +1,0 @@
-<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16"><path d="M4.002 4h8v8h-8z" style="marker:none" color="#000" overflow="visible" fill="#474747"/></svg>
--- a/data/ui/plugin_row.ui
+++ b/data/ui/plugin_row.ui
@@ -44,6 +44,11 @@
                 <property name="valign">center</property>
                 <property name="opacity">0</property>
                 <property name="icon-name">ee-drag-handle-symbolic</property>
+                <property name="cursor">
+                    <object class="GdkCursor">
+                        <property name="name">grab</property>
+                    </object>
+                </property>
             </object>
         </child>
     </object>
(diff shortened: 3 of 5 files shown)


wwmm:
Although this makes the code simpler as there is no need to switch icons I am not sure it is visually better
Screenshot from 2022-05-01 15-08-50
Do we really need to do this? I think things are fine as they are. And at least to me these top/bottom entries give the weird feeling that clicking on them should reveal widgets on the right side like clicking on the plugins name do.

Digitalone1:
I thought about [this](https://gitlab.gnome.org/Teams/Circle/-/issues/96#note_1393920):

> Overall, pages etc probably need some text explaining what's going on. For example, the Input plugins are applied before going into the recorder.

Despite all the efforts and the arrows, some users still can't understand what's going on.

I know, seeing those labels alone without effects is weird, but they are useful to understand how EasyEffects works when the stack is not empty. Maybe we can hide them when the plugin stack is empty. We already check for empty stack, it's not difficult to do this.

For the feeling of revealing something in the plugin view, I agree. I had this feeling too, but it's not our issue.

The problem is GTK can't distinguish between clickable and non-clickable widgets. HTML and CSS as web standards have the cursor property enabled by default on the links so that the pointer changes into an hand to suggest the users they can click on that item. I wonder why GTK does not do something like this by default.

Maybe there's a specific property we can enable to show the hand cursor to distinguish between effects in the plugin stack and start/endpoint labels. I will check it out.

It should be refined, but I think it's useful anyway.

Digitalone1:
Okay, I managed to do it. Now the start/endpoint labels are not shown when the stack is empty.

`pointer` cursor is set on clickable widgets in the left sidebar. With pointer cursor is more intuitive to understand which object is interactive.

However it would be better to implement the cursor property inside the xml builder, but I don't know how to do it properly because the property expects a gdk cursor object: https://docs.gtk.org/gtk4/property.Widget.cursor.html

I never found a documentation on how to write xml builders.

Anyway, @wwmm tell me what do you think about that.

Digitalone1:
I set half opacity on the start/endpoint labels. This make clickable objects still more intuitive.

I wanted to set the `dim-label` style, but it works only for labels, so setting the opacity on both labels and icons is a better idea.

I'm also testing how to implement cursor pointer from the xml builder.

Digitalone1:
Okay, I managed to move the GdkCursor object into the xml builder. I didn't know properties can have an object as child.

Now seems way better than before.

wwmm:
>I didn't know properties can have an object as child.

Not all of them. I think only the ones that receive some kind of object.

>tell me what do you think about that.

It is better now those entries are not shown when there is no filter. But instead of `Playing Apps` and `Recording Apps` I think it is better to just say `Players` and `Recorders` respectively. It will match the name of each tab and is shorter. What is probably going to help translations to not take too much horizontal space. In any case this can be done later. I will merge this.

wwmm:
>Not all of them. I think only the ones that receive some kind of object.

I guess this did not explain much :smile:. What I was thinking about is that in those situations where a widget has objects set through a method we usually are able to do the same in the xml. I did that in those factories that are used by the listviews. The models and the factories can be set in the xml. We do not have to call a C function for this.

Issue #1524, opened 2022-05-14
Can't load presets


Miroslav11:
### EasyEffects Version

6.2.5

### What package are you using?

Flatpak (Flathub)

### Distribution

Linux Mint 20.3

### Describe the bug

Pipewire is installed and it's working fine. I'm trying to deal with "Dynamic range compression" in easyeffects.

I've put the preset "LoudnessEqualizer.json" from (https://github.com/Digitalone1/EasyEffects-Presets) inside ".var/app/com.github.wwmm.easyeffects/config/easyeffects/output", and when I'm trying to load it "Presets -> Load", it's not loading

### Expected Behavior

Should be loaded

### Debug Log

<details><summary>Debug Log</summary><br/>

```
Paste your log here
```
</details>

Digitalone1:
@vchernin Could it be a Flatpak permission issue?

Miroslav11:
Probably, but I don't know which permission?

wwmm:
@Miroslav11 try to get some logs as explained at https://github.com/wwmm/easyeffects/wiki/Reporting-Bugs#basic-logs. Maybe there is a error message when trying to load the preset that could point us in the right direction.

Miroslav11:
[loading_presets.txt](https://github.com/wwmm/easyeffects/files/8693553/loading_presets.txt)

Miroslav11:
@wwmm there you go :)

wwmm:
It is the first time I see an error like this
```
presets_manager: [json.exception.parse_error.101] parse error at line 7, column 1: syntax error while parsing value - invalid literal; last read: '<U+000A><U+000A><U+000A><U+000A><U+000A><U+000A><'

presets_manager: [json.exception.type_error.304] cannot use at() with null
```
I wonder if the preset file was somehow corrupted.

wwmm:
At least in the github page the preset file has nothing unusual at the line `7`

Miroslav11:
@Digitalone1 said it's working fine on non flatpak versions.

I've tried on both of my OS, Pop OS 22.04 (native pipewire), and on Linux mint 20.3. On both systems I can't load

wwmm:
>Digitalone1 said it's working fine on non flatpak versions.

I have just tested one of his presets on my computer (not flatpak) and there was no problem. But I can't see how Flatpak could be causing an `invalid literal` when trying to load the preset. Weird...

[...]

PR #1526, opened 2022-05-15
Fix Maximum History in Autogain

Digitalone1:
I'm trying to fix the issue shown in #1524 regarding the Maxmium History in Autogain.

Meanwhile I uniformed the calls to prepare widgets.

### DIFF
--- a/include/autogain.hpp
+++ b/include/autogain.hpp
@@ -90,7 +90,7 @@ class AutoGain : public PluginBase {
 
   static auto parse_reference_key(const std::string& key) -> Reference;
 
-  void set_maximum_history(const int& value);
+  void set_maximum_history(const int& seconds);
 };
 
 #endif
--- a/include/ui_helpers.hpp
+++ b/include/ui_helpers.hpp
@@ -48,6 +48,12 @@ void remove_from_string_list(GtkStringList* string_list, const std::string& name
 
 template <StringLiteralWrapper sl_wrapper>
 void prepare_spinbutton(GtkSpinButton* button) {
+  if (button == nullptr) {
+    util::warning("Null pointer provided: Spinbutton widget not prepared.");
+
+    return;
+  }
+
   g_signal_connect(button, "output", G_CALLBACK(+[](GtkSpinButton* button, gpointer user_data) {
                      return parse_spinbutton_output(button, sl_wrapper.msg.data());
                    }),
@@ -67,6 +73,8 @@ void prepare_scale(GtkScale* scale) {
   */
 
   if (scale == nullptr) {
+    util::warning("Null pointer provided: Scale widget not prepared.");
+
     return;
   }
 
--- a/src/app_info.cpp
+++ b/src/app_info.cpp
@@ -356,7 +356,7 @@ void app_info_init(AppInfo* self) {
 
   self->app_settings = g_settings_new("com.github.wwmm.easyeffects");
 
-  prepare_spinbutton<"%">(self->volume);
+  prepare_spinbuttons<"%">(self->volume);
 
   self->data->handler_id_enable = g_signal_connect(self->enable, "toggled", G_CALLBACK(on_enable), self);
   self->data->handler_id_volume = g_signal_connect(self->volume, "value-changed", G_CALLBACK(on_volume_changed), self);
--- a/src/autogain.cpp
+++ b/src/autogain.cpp
@@ -148,10 +148,10 @@ auto AutoGain::parse_reference_key(const std::string& key) -> Reference {
   return Reference::geometric_mean_msi;
 }
 
-void AutoGain::set_maximum_history(const int& value) {
+void AutoGain::set_maximum_history(const int& seconds) {
   // The value given to ebur128_set_max_history must be in milliseconds
 
-  ebur128_set_max_history(ebur_state, g_settings_get_int(settings, "maximum-history") * 1000);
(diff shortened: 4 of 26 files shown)


Digitalone1:
So, `g_settings_get_int` is called inside `set_maximum_history` even if it's already called before. I modified the code to use the values passed by reference.

But I can't resolve the issue. Using debug strings I saw the execution stops inside `ebur128_set_max_history`, so it should be an issue related to libebur128.

I tried also to pass the gsettings value "by value" rather than reference, but nothing changes. I ended up only making a static cast since the value expected is unsigned long as reported [here](https://github.com/jiixyj/libebur128/blob/67b33abe1558160ed76ada1322329b0e9e058b02/ebur128/ebur128.h#L206).

Any idea @wwmm?

Digitalone1:
I answer here.

> I have no problems at all with the maximum history control in the autogain window. Even If I change it while audio is being processed. But whatever it is happening in your computers I agree it has nothing to do with the preset bug.

That's really weird. I can always reproduce the issue. I can't change the maximum history value and it's the only parameter giving this problem and getting a segmentation fault.

wwmm:
>That's really weird. I can always reproduce the issue. I can't change the maximum history value and it's the only parameter giving this problem and getting a segmentation fault.

Try to run the executable inside gdb or running `sudo coredumpctl gdb` as soon as EasyEffects crashes. If you can reproduce but I can't my guess is a multithreading problem somewhere.

Digitalone1:
> > That's really weird. I can always reproduce the issue. I can't change the maximum history value and it's the only parameter giving this problem and getting a segmentation fault.
> 
> Try to run the executable inside gdb or running `sudo coredumpctl gdb` as soon as EasyEffects crashes. If you can reproduce but I can't my guess is a multithreading problem somewhere.

```
Thread 1 "easyeffects" received signal SIGSEGV, Segmentation fault.
0x00007ffff6dffbb1 in ebur128_set_max_history () from /usr/lib/libebur128.so.1
(gdb) bt
#0  0x00007ffff6dffbb1 in ebur128_set_max_history ()
    at /usr/lib/libebur128.so.1
#1  0x0000555555612a03 in AutoGain::set_maximum_history(int const&) ()
#2  0x000055555561249d in AutoGain::AutoGain(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, PipeManager*)::{lambda(_GSettings*, char*, void*)#2}::operator()(_GSettings*, char*, void*) const ()
#3  0x000055555561250d in AutoGain::AutoGain(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, PipeManager*)::{lambda(_GSettings*, char*, void*)#2}::_FUN(_GSettings*, char*, void*) ()
#4  0x00007ffff72317f5 in g_cclosure_marshal_VOID__STRINGv ()
    at /usr/lib/libgobject-2.0.so.0
#5  0x00007ffff72506b6 in g_signal_emit_valist ()
    at /usr/lib/libgobject-2.0.so.0
#6  0x00007ffff7250824 in g_signal_emit () at /usr/lib/libgobject-2.0.so.0
#7  0x00007ffff7364637 in  () at /usr/lib/libgio-2.0.so.0
#8  0x00007ffff72f0145 in  () at /usr/lib/libgio-2.0.so.0
#9  0x00007ffff72506b6 in g_signal_emit_valist ()
    at /usr/lib/libgobject-2.0.so.0
#10 0x00007ffff7250824 in g_signal_emit () at /usr/lib/libgobject-2.0.so.0
#11 0x00007ffff7365f4f in  () at /usr/lib/libgio-2.0.so.0
#12 0x00007ffff7360d5f in  () at /usr/lib/libgio-2.0.so.0
#13 0x00007ffff7dd7825 in g_main_context_invoke_full ()
    at /usr/lib/libglib-2.0.so.0
#14 0x00007ffff7361511 in  () at /usr/lib/libgio-2.0.so.0
#15 0x00007ffff24d9f8f in  () at /usr/lib/gio/modules/libdconfsettings.so
#16 0x00007ffff24da2b2 in  () at /usr/lib/gio/modules/libdconfsettings.so
#17 0x00007ffff73b7d8f in  () at /usr/lib/libgio-2.0.so.0
#18 0x00007ffff73718d3 in  () at /usr/lib/libgio-2.0.so.0
#19 0x00007ffff7232f8f in g_closure_invoke ()
    at /usr/lib/libgobject-2.0.so.0
#20 0x00007ffff724eba8 in  () at /usr/lib/libgobject-2.0.so.0

```

wwmm:
Humm... I have an idea. Let try to change the history before allowing the plugin to be initialized. In other words before starting a stream

Digitalone1:
Can you merge and work from here?

wwmm:
>I have an idea. Let try to change the history before allowing the plugin to be initialized. In other words before starting a stream

That is it. Now I see the crash. The problem is that `set_maximum_history` is not checking if `ebur_state` is not a `nullptr`. The pointer `ebur_state` is only initialized after the processing function from the filter is called. What is only going to happen if a stream is being processed in EasyEffects. I probably never tried to change this parameter when nothing was playing.

wwmm:
Fixed in the master branch.

PR #1529, opened 2022-05-15
Fix compilation of pulseaudio-legacy with Boost 1.79.0

DrMcCoy:
Boost.Filesystem in Boost 1.79.0 changed the already-deprecated header
of string_file.hpp to be not included by default anymore in its commit
266e1ac892a6f54d807fb35bf639a9aa1c8b2db1.

Boost.Filesystem's fstream.hpp was never included directly by default,
but string_file.hpp did include it, so this code here in PulseEffects
worked. With Boost 1.79.0, though, it explodes in lin 179.


wwmm:
Thanks!

PR #1542, opened 2022-05-24
Add capture node blocklist

Digitalone1:
Related to #1541 

I think this is a more elegant way to exclude OBS.

But it has to be tested against a localized OBS. The hack for pavucontrol is done in `on_info_node` not only because `PW_KEY_STREAM_MONITOR` is not set in `on_registry_global`, but mainly because pavucontrol app name is localized and different between languages set at system level.

@wwmm Can you run and test OBS in Portuguese? I don't have OBS installed on my system.

Anyway, If this is not working or you think it's better to do it inside `on_node_info`, you can reject the merge request.

### DIFF
--- a/include/pipe_manager.hpp
+++ b/include/pipe_manager.hpp
@@ -228,6 +228,8 @@ class PipeManager {
 
   std::array<std::string, 2U> blocklist_media_role = {"event", "Notification"};
 
+  std::array<std::string, 1U> blocklist_capture_node = {"OBS"};
+
   std::string header_version, library_version, core_name;
   std::string default_clock_rate = "0";
   std::string default_min_quantum = "0";
--- a/src/pipe_manager.cpp
+++ b/src/pipe_manager.cpp
@@ -221,10 +221,8 @@ void on_node_info(void* object, const struct pw_node_info* info) {
   auto* const pm = nd->pm;
 
   if (auto node_it = pm->node_map.find(nd->nd_info->timestamp); node_it != pm->node_map.end()) {
-    bool remove_node = false;
-
     /*
-      This is a workaround for issue #1128.
+      The following is a workaround for issue #1128.
       Sometimes monitor streams like Pavucontrol can't be blocklisted inside on_registry_global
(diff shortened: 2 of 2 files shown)


Digitalone1:
Forgot to say, we have to check also if `PW_KEY_STREAM_CAPTURE_SINK` is set in `on_registry_global`.

sephid86:
내가 테스트 해 본 결과는 잘 작동합니다.
What I've tested works fine.
test1

wwmm:
@Digitalone1 I had to revert the commit. Looking at the code it seemed fine but for some reason it did not work. I guess it is one of those situations where the stream property is only set at a later time. For now it is better to keep the check for OBS inside the node_info callback.

sephid86:
> @Digitalone1 I had to revert the commit. Looking at the code it seemed fine but for some reason it did not work. I guess it is one of those situations where the stream property is only set at a later time. For now it is better to keep the check for OBS inside the node_info callback.



stream_input_effects.cpp

  142 void StreamInputEffects::on_app_added(const NodeInfo node_info) {
  143   const auto blocklist = util::gchar_array_to_vector(g_settings_get_strv(settings, "blocklist"));
  144 
  145   auto is_blocklisted = std::ranges::find(blocklist, node_info.application_id) != blocklist.end();
  146 
  147   is_blocklisted = is_blocklisted || std::ranges::find(blocklist, node_info.name) != blocklist.end();
  148 
~ 149   if (g_settings_get_boolean(global_settings, "process-all-inputs") != 0 && !is_blocklisted || remove_node != "true") {                                                                                      
  150     pm->connect_stream_input(node_info.id);
  151   }    
  152 }

wwmm:
>What do you think about this?

Although it could be done inside `on_app_added` it is not really going to be simpler than the current implementation inside `on_node_info` because our nodes will have to save that property about the capture from monitors so it can be used later. I think it is fine to keep the workaround as it is. The real improvement to this situation would be having no workaround at all. What will only be possible if OBS developers start to tag their streams with an unique id instead of using the media name tag for a task that is not the one the tag was created for.

Digitalone1:
I think this can be done better. I will open a draft.

Digitalone1:
I think this can be done better. I will open a draft.

Issue #1335, opened 2022-01-13
Process all inputs: move only streams recording from selected mic


Bleuzen:
### EasyEffects Version

6.2.1

### What package are you using?

Arch (easyeffects)

### Distribution

Arch Linux

### Describe the bug

Hi, currently there is an option to process all inputs, which could be nice since I want to process my mic for all applications, but it is not usable as is.
The problem is that, when this option is enabled, EasyEffects moves ALL recording streams to the EasyEffects Source.
This also includes streams recording from monitor sources, so one can't capture desktop audio anymore when this option is enabled.
Also in case of having multiple mics this is problematic, since recording multiple of them is not possible anymore, since EasyEffects moves all recording streams to the EE Source, which only processes one mic.

### Additional Information

To solve this, as an idea,
instead of (pseudo code):

```
foreach (inputstream in inputstreams) {
    // move stream to record from EasyEffects source
}
```

something like this could be done:

```
foreach (inputstream in inputstreams) {
    if (inputstream is recording from selected mic in EasyEffects settings) {
        // move stream to record from EasyEffects source
    }
}
```

wwmm:
The "solution" to this problem already exists. Add the streams for which no effects should be applied to the blocklist. That being said there are some applications like OBS #807 for which the blocklist doesn't work. And unfortunately there isn't a way to handle super corner cases like that.

wwmm:
>something like this could be done:

In thery seems simple but that approach also has undesirable side effects. The default configuration is using the system default microphone. It is totally possible that the user has 2 applications recording from the default mic but he wants effects only for one of them. IF we just move to our virtual device anything recording from the default mic just because that is the mic EasyEffects is using there will also be problems.

The current blocklist approach is far from perfect. But I doubt a perfect solution will ever be available considering what is discussed at #807. In general audio applications do not make good use of the tags supported by the audio servers. If they did a more advanced filtering procedure could be implemented.

Bleuzen:
> The "solution" to this problem already exists. Add the streams for which no effects should be applied to the blocklist.

This is not a solution here. I mentioned multiple mics / inputs (also instruments which I don't want to run through EE).
As an example: I want to record in Audacity. Instruments / secondary mics without processing, but main mic with processing. Adding Audacity to the blocklist just won't apply any processing at all, so this isn't doing it. I want processing for Audacity, but only when it tries to record from my main mic, leaving other input streams recording for example instruments as is.



[...]

Issue #1128, opened 2021-09-03
Pavucontrol is shown in input apps


Digitalone1:
I rarely touch the stream input pipeline, but today I found out this:

![Schermata del 2021-09-03 18-58-58](https://user-images.githubusercontent.com/25790525/132041963-adee350f-48a6-400a-abc2-d8e51717748c.png)

Wasn't it supposed to be hard blocklisted? 

@wwmm Am I wrong or this was happening also in PulseEffects because the localized app name skipped the blocklist and you had to exclude apps on other parameters?

wwmm:
>Wasn't it supposed to be hard blocklisted?

I tried it on my computer and there was no entry. But I am not using translations. My window is in English. I think that the string currently used in our built-in blocklist may be translation dependent. IF that is the case we will have to replace it by another that does not change in this situation.

Digitalone1:
That is the case, we have to find a parameter reporting the application name, not the translated title.

wwmm:
The command `pw-cli list-objects Node` will help to find other tags that may not have been translated.

wwmm:
The output for pavucontrol is
```
                id 125, type PipeWire:Interface:Node/3
 		factory.id = "6"
 		client.id = "206"
 		application.name = "PulseAudio Volume Control"
 		node.name = "PulseAudio Volume Control"
 		media.class = "Stream/Input/Audio"
```

At this moment we are checking the `node.name` 
https://github.com/wwmm/easyeffects/blob/e7597e192c1cee2db1f30500c2fa3c5a5ef875fe/src/pipe_manager.cpp#L797
https://github.com/wwmm/easyeffects/blob/e7597e192c1cee2db1f30500c2fa3c5a5ef875fe/include/pipe_manager.hpp#L193

I wonder if `application.name` is not translated.

Digitalone1:
```
	id 54, type PipeWire:Interface:Node/3
 		factory.id = "6"
 		client.id = "51"
 		application.name = "Regolazione del volume PulseAudio"
 		node.name = "Regolazione del volume PulseAudio"
 		media.class = "Stream/Input/Audio"
	id 56, type PipeWire:Interface:Node/3
 		factory.id = "6"
 		client.id = "51"
 		application.name = "Regolazione del volume PulseAudio"
 		node.name = "Regolazione del volume PulseAudio"
 		media.class = "Stream/Input/Audio"
```

So we can't blocklist it. :(

wwmm:
[...]

PR #1545, opened 2022-05-24
Improve internal blocklist handling + App tags

Digitalone1:
Related to #1128, #1335, #1541 and #1546.

I think this is better because we check for Pavucontrol only in `on_node_info` looking at app id rather than trying to do it in `on_registry_global` and then checking for stream monitor key later.

It's better to look for node names since it's done only the first time when the node is added, but this is not reliable for localized app names (Pavucontrol), so I think those changes are a cleaner approach.

Regarding OBS and capture streams, I excluded the latter in `on_node_info` and there we could add a gsettings key to let the user allow/deny this task. We can discuss if it's convenient blocking `OBS` directly checking the name or just excluding all the capture streams.

Since we talked about app id, I wonder why we don't set it for our nodes. Maybe should we do it?

I mark this as draft, fell free to change/improve it.


wwmm:
>Since we talked about app id, I wonder why we don't set it for our nodes. Maybe should we do it?

Unfortunately almost no player sets `application.id` in streams. There isn't much point in adding this property to our nodes as things are right now. And in EasyEffects only the blocklist uses this id if the audio application sets it.

We could create this property in the node info. But it is not going to make much difference.

Digitalone1:
I know, but since we ask to set id for a better blocklist handling, we should set if before as an example to follow.

Digitalone1:
Another thing I'd like to do is getting rid of timestamps and rely only on serials to track nodes, which should be unique unlike node id, as the pipewire doc says.

wwmm:
>Another thing I'd like to do is getting rid of timestamps and rely only on serials to track nodes, which should be unique unlike node id, as the pipewire doc says.

Yes. I think we should also drop any reference to the `object.id`. It may still have some use for PipeWire internally but in our case we want something that is unique for the object. I think I will have time to do something about this tomorrow. But let me know in case you have already started to replace things by `object.serial`.

About `ee_app_id` I think it may be better to define this in a new header file independently from `pipe_manager.hpp`. This id string isn't conceptually tied to the PipeWire class. I also think calls to runtime objects like 
```
self->data->application->pm->ee_app_id.c_str()
```
can probably be avoided because the variable is static anyway. An approach similar to https://github.com/wwmm/easyeffects/blob/master/include/equalizer_tags.hpp may work.

wwmm:
Or maybe just something like `auto constexpr log_tag = "com.github.wwmm.easyeffects";` inside a header only for ids. If the destination location needs it to be a string `using namespace std::string_literals;` could be used as needed.

Digitalone1:
I'm not working on the serial, but I made a new header for app tags. Will make a new commit in the next hours.

Digitalone1:
I think it's way better now. Unfortunately I couldn't make strings as `constexpr` because of a compiler error, but I think it's good as it is.

If you want, I could add a specific config to enable/disable the capture streams.

The config could be named `Exclude capture streams` and it would be `true` as default.

wwmm:
>The config could be named Exclude capture streams and it would be true as default.

Ok. I think I can do this today.

wwmm:
It seems `object.serial` can not be used to reset or set `PW_KEY_NODE_TARGET`. For this one we have to use `object.id`... Weird...

wwmm:
>Exclude capture streams

On a second though I think this may give the wrong idea that all recorder streams will be excluded. Only capture streams that monitor devices will be exclude. Maybe `Exclude Device Monitor Streams` is a better description.

Digitalone1:
> > Exclude capture streams
> 
> On a second though I think this may give the wrong idea that all recorder streams will be excluded. Only capture streams that monitor devices will be exclude. Maybe `Exclude Device Monitor Streams` is a better description.

You're right.

Digitalone1:
I'm working on it. Hope to do a MR in the next hours.

PR #1549, opened 2022-05-27
Track nodes by pw serial

Digitalone1:
It seems to works. It has to be tested for days because this transition is error prone, but at least to me it's working just switching blocklist and presets.

Changed presets many times and only one time the app remained disabled. The rest it was working good (it could be more a pw issue).

I can't test with multiple outputs and inputs devices, so please do it @wwmm. Anyway I switched from speakers to headphone and no issues.

The only issue I had: `assertion 'default_value >= minimum && default_value <= maximum' failed` using `G_MININT64` as minimum parameter in `g_param_spec_uint64`, but I resolved replacing `SPA_ID_INVALID`. Maybe a casting issues, I commented the code reporting the workaround where this happened.

### DIFF
--- a/include/pipe_manager.hpp
+++ b/include/pipe_manager.hpp
@@ -35,8 +35,6 @@
 #include "util.hpp"
 
 struct NodeInfo {
-  long timestamp;
-
   pw_proxy* proxy = nullptr;
 
   uint id = SPA_ID_INVALID;
@@ -190,14 +188,7 @@ class PipeManager {
 
   spa_hook metadata_listener{};
 
-  /*
-    Nodes are tracked in a map and indexed by the creation date UNIX timestamp.
-    This way we should have always a different index for any streams while PipeWire
-    try to reassign already used ID and maybe we should avoid issues with bad
-    applications creating and destroying too many streams in a very short time.
-  */
-
-  std::map<long, NodeInfo> node_map;
+  std::map<uint64_t, NodeInfo> node_map;
 
   std::vector<LinkInfo> list_links;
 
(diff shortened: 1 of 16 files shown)


wwmm:
So far everything is fine. Tomorrow I will take a look at some of the places where we still use the `id` when we actually wnat the device unique number. Like the code handling the autoloading profiles or any other accessing `pm->output_device.id` or `pm->input_device.id`.

wwmm:
@Digitalone1 I have opened an issue in the Piepwire's page that may be of interest to you https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/2409. As thing are right now trying to use object.serial in more cases is a pain. The amount of code we can actually move to object.serial now without a lot of pain is way smaller than what I had imagined. But at the same time we already have bugs related to the usage of `object.id`.

Digitalone1:
I read the issue. The serial is good for streams since we introduced the timestamps to handle those bad apps like audacity that were spawning and closing lots of streams in very short time reusing id. Now with serial we have a new value at every new stream which is good.

But this is not what we want for devices. Did you test if there's a way to have a string uniquely associated to devices and use it to create links?

wwmm:
>Did you test if there's a way to have a string uniquely associated to devices and use it to create links?

Not yet. I think I will have time for this only on Tuesday. The device object we currently use only because of the autoloading profiles has the base name without the things that are usually appended in the node's `object.name`. If the node had a property like this one with the base name it would be definitely viable to use it. But with the property in another object that has its own events I am not sure it is going to be useful.

PR #1557, opened 2022-06-03
App icon refresh

### LINKED ISSUE #1520: New app icon?
The current icon doesn't quite fit the HIG, would you be interested in help with a new one?

Here's a first attempt at making the current metaphor a bit more HIG-compliant, but we could also try something completely different.

image

bertob:
Add new app icon more in line with the guidelines, including a symbolic icon and the source SVG.

The symbolic needs hooking up to the build system, since I don't really know how that works, but maybe that can be done as a followup since currently there's no symbolic anyway.

image

Closes #1520


wwmm:
Thanks! It is a lot better :-). Let's hope QT users have no problems.

I've changed our meson scripts and I think I am installing the symbolic icon in the correct place. Do you know where the symbolic icon is used in the system? I can see there are some application symbolic icons under `/usr/share/icons/hicolor/symbolic/apps/` but I do not remember having seen them being used by something.

vchernin:
> Do you know where the symbolic icon is used in the system?

IIRC usually the symbolic icon is used in the top left in gnome next to the app name (assuming there are no extensions changing the top bar).

vchernin:
Also, I think you might want to check the about EasyEffects page is showing the new icon. I recall it hardcoded `easyeffects`.

wwmm:
>IIRC usually the symbolic icon is used in the top left in gnome next to the app name (assuming there are no extensions changing the top bar).

Oh... I think I have never noticed there was an icon there :smile: 
Screenshot from 2022-06-03 20-32-20

>Also, I think you might want to check the about EasyEffects page is showing the new icon. I recall it hardcoded easyeffects.

Really? I do not remember having the icon in our readme. Or is this another page?

vchernin:
> Really? I do not remember having the icon in our readme. Or do you talk about another page?

I mean in the "about app" page accessed via the hamburger menu, in application itself. 

https://github.com/wwmm/easyeffects/blob/26de9ea5617609182668f86b078ef249d4acc349/src/application.cpp#L417

wwmm:
>I mean in the "about app" page accessed via the hamburger menu, in the application itself.

Of course I forgot about it :smile:. I have sent a new update to the master branch.

PR #1562, opened 2022-06-07
build: Declare required fmt version

jtojnar:
Since https://github.com/wwmm/easyeffects/commit/755595e0e98d35c03558cb19c1950e1c5ecefb75, the build was failing with:

    error: no matching function for call to 'format(std::locale, const char [14], double&, int&, std::__cxx11::basic_string<char>)'


wwmm:
Thanks!

Issue #1309, opened 2022-01-03
Convolver auto-gain


future-figs:
### EasyEffects Version

6.1.5

### What package are you using?

Fedora

### Distribution

Fedora 35

### Describe the bug

When using the convolver, an unpredictable amount of gain is applied to the IR kernel, and therefore the output signal. This makes it difficult to prevent digital clipping in EQ applications without manually reducing the gain to excess.

The convolver has auto-gain functionality built in. It first normalizes the IR kernel to just short of digital clipping, like a CD master [-1.0 +1.0]. Then, if the sum of squares is still less than 1.0, it increases the gain further. This increases the volume level of the kernel in most cases, and can introduce digital clipping even with both preamp sliders left at 0 dB.
https://github.com/wwmm/easyeffects/blob/d820fa80b3b929225a46fbee253dccdff55e2c78/src/convolver.cpp#L357-L389

It's easy to apply some negative gain with the input/output sliders, but there's no way to know how much is enough—cue obsessing over every little pop or crackle.

### Expected Behavior

I can think of two potential ways to improve this behavior:
1. Change the auto-gain to use an absolute-value sum instead (greatest of either channel). If this sum is <=1, then digital clipping is impossible, unless extra gain is applied on top etc. Ideal for EQ applications, but tends to reduce the volume level noticeably. I can submit a PR if this approach is favored.
2. Do nothing. The user can either scale their IR kernels appropriately ahead of time, or else play with the preamp sliders like they currently have to anyway.

wwmm:
>This increases the volume level of the kernel in most cases, and can introduce digital clipping even with both preamp sliders left at 0 dB.

That is weird. None of the impulse responses I have tested lead to an autogain factor higher than 1. In one of my impulse files the output signal is amplified even if the kernel is scaled with an factor equal to `0.614267`. So something else is at play. Part of the amplification seems to be related to the convolution process.

wwmm:
>Then, if the sum of squares is still less than 1.0, it increases the gain further

I am doing some tests and I do not see this happening. When the average power is less than one `std::min` forces the autogain factor to be equal to `1`.

Mr-Figs:
> > Then, if the sum of squares is still less than 1.0, it increases the gain further
> 
> I am doing some tests and I do not see this happening. When the average power is less than one `std::min` forces the autogain factor to be equal to `1`.

[...]

PR #1563, opened 2022-06-08
Toggle convolver autogain

Zhern:
This is related to the issue #1309.
It is an implementation of a simple button in the Convolver plugin that allows to enable or disable the autogain.

I tested my convolution equalization `.wav` file under several applications and OS, in general the gain is not altered.
I find this feature weird for my use case but I guess some people prefer it as is, so I implemented a way to quickly turn it off/on.

### DIFF
--- a/data/schemas/com.github.wwmm.easyeffects.convolver.gschema.xml
+++ b/data/schemas/com.github.wwmm.easyeffects.convolver.gschema.xml
@@ -16,5 +16,8 @@
             <range min="0" max="200" />
             <default>100</default>
         </key>
+        <key name="autogain" type="b">
+            <default>false</default>
+        </key>
     </schema>
 </schemalist>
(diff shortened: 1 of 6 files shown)


Digitalone1:
I think it's better to use an `autogain` private bool property and set it accordingly to the new gsettings key, then use it inside `apply_kernel_autogain` rather than calling `g_settings_get_boolean`.

wwmm:
>I find this feature weird for my use case but I guess some people prefer it as is

When I implemented this convolver I took some inspiration from another one that is available as an LV2 plugin https://github.com/tomszilagyi/ir.lv2. And this convolver applies autogain. After testing with many different impulse files the overall experience was way better with an autogain. Depending on the impulse response the difference in the output volume can be huge without some kind of correction. 

>I think it's better to use an autogain private bool property and set it accordingly to the new gsettings key, then use it inside apply_kernel_autogain rather than calling g_settings_get_boolean.

I agree. This way other additional calls to `g_settings_get_boolean(settings, "autogain")` won't be needed if `apply_kernel_autogain` is invoked again.

PR #1590, opened 2022-06-21
Pipewire dictionary optimizations

Digitalone1:
Since lots of optimizations have been done ultimately, I though this would make the `pipe_manager` more readable without all those conditional statements. At least where it's convenient to use the helper function/template.

@wwmm  Anyway I found something weird. For `device.nick`, did you really want to use `PW_KEY_DEVICE_NAME` or did you mean `PW_KEY_NODE_NICK`?


wwmm:
>did you really want to use PW_KEY_DEVICE_NAME or did you mean PW_KEY_NODE_NICK?

No. It was probably a mistake the fortunately had no consequences. This property is not being used at thins moment. I think I considered using it t some point but I did not. In case I have updated the master branch with the correct value now.

> If we will use string views in the future, this template could be useful.

I had this in mind when I started optimizing our string usage. But once I realized `std::string_view` does not provide a `const char*` I was disappointed. In other to pass it to all these C functions we are forced to interact with it is necessary to create a `std::string` just to have the `const char*`. That is not optimal. I am defining as many strings as possible as `constexpr const char*` exactly to allow the compiler to create them during the compilation. The way it was before with `std::string` was forcing all of them to be created at runtime. Using string views just to be forced to do the same unneeded runtime manipulation in the end would not make sense. So I gave up on them...

There may be some places here an there where we can use string views. But in the vast majority of the places they are not the way to go :cry:

wwmm:
>So you have to generate a path tag for every plugin like it has been already done for schemas?

Unfortunately we can not do this because we intend to support multiple plugins instances. So at the end of the day the final path will have to be generated at runtime because each instance will need its own path. But we only know this once the code is running. 

In any case I think we still can replace the `std::string` in the header by a `constexpr const char*` and let the runtime code to the place where the plugin path is being generated. The problem of having a std::string in the header is that this object is evaluated only at runtime.  And that is why initialization order becomes something to worry about. But like I said before it did not cause any problem as far as I know. It is just to be on the safe side.

wwmm:
>...and let the runtime code to the place where the plugin path is being generated.

The downside is that we can't use a `+` in the const char. So the variable would need to be passed to the constructor of a std::string or a `s` will need to be added at the end of string literals being added to the tag... That is why I am still thinking about what to do with it...

wwmm:
>or a s will need to be added at the end of string literals being added to the tag...

Looking at the code it does not seem so bad. It seems there are only 3 places using this path variable. For some reason I thought there were more...

Digitalone1:
It's unclear to me why const chars can't be summed in C.

Javascript and PHP can simply make `"string1" + "string2"`. Maybe you can define an util function to perform the concatenation?

wwmm:
>It's unclear to me why const chars can't be summed in C.

Being honest I never searched why it can't be done. Maybe because they are similar to an array of numbers but using characters. We also can not do `a + b` if a and b are array of numbers.

>Maybe you can define an util function to perform the concatenation?>

Something I am tempted to do is the solution proposed at https://stackoverflow.com/questions/38955940/how-to-concatenate-static-strings-at-compile-time. It concatenates const chars* when the code is compiled instead of doing it at runtime. But as it is implemented with string_views and we have to deal with tons of C functions it may not be convenient in the end...

Digitalone1:
I think that a solution can be found, but if the strings have to be generated runtime when we implement the multiple instances, the issue remains.

wwmm:
>I think that a solution can be found, but if the strings have to be generated runtime when we implement the multiple instances, the issue remains.

For now I have updated the master branch with the string literal a approach `s` appended to one of the strings being added to the path tag. It is a concatenation that can be done when the code is compiled. But the rest of the concatenation has to be done at runtime because the path is of the plugin is dynamic. Only the schema id is know at compilation.

wwmm:
I think the next thing I will try to do is some small changes to `src/effects_base.cpp`, `stream_output_effects.cpp` and `src/stream_input_effects.cpp` that will be required for multiple filters instances. Instead of creating all the instances in the effects_base constructor we have to create them ondemand. And I already see this is going to be annoying XD. Besides the fact we have to make sure the `plugins map` is properly updated there are also signals that have to be connected and disconnected every time one of those classes are created/destroyed. This is going to take some time :smile:

PR #1610, opened 2022-06-29
Autogain: fix display of `Internal output gain`

LebedevRI:
The `AutoGain::process()` is treating `internal_output_gain` as
a linear gain:
https://github.com/wwmm/easyeffects/blob/28d1172819424b16b8df4f274387cf6604ca95da/src/autogain.cpp#L322-L324
... yet the display code thought that it was in dB's:
https://github.com/wwmm/easyeffects/blob/28d1172819424b16b8df4f274387cf6604ca95da/src/autogain_ui.cpp#L96-L97

I don't know which of the two need to be adjusted, i fixed this one.

### DIFF
--- a/src/autogain_ui.cpp
+++ b/src/autogain_ui.cpp
@@ -93,8 +93,9 @@ void setup(AutogainBox* self, std::shared_ptr<AutoGain> autogain, const std::str
         gtk_level_bar_set_value(self->l_level, util::db_to_linear(loudness));
         gtk_label_set_text(self->l_label, fmt::format("{0:.0f}", loudness).c_str());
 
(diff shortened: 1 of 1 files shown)


wwmm:
>I don't know which of the two need to be adjusted, i fixed this one.

The one in `autogain_ui` is wrong. While rewriting the window code I probably did not notice this while doing some copy and paste.

wwmm:
So it is probably better to fix in the gtk code instead of converting to db just to go back to linear again.

LebedevRI:
Sure, done.
Thanks!

LebedevRI:
Thank you!

Issue #1607, opened 2022-06-28
Showing info messages to the user


Digitalone1:
Sooner or later we should add a way to show the user some messages/information like in a dialog box.

We could use Adwaita [Toast](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/1-latest/class.Toast.html) or [ToastOverlay](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/1-latest/class.ToastOverlay.html). We already use the warning messages, but those are for debugging purpose and the user usually don't read them because they did not start the application through the command line.

This could be useful for various things we already discussed in other issues. The first ones that comes in my mind:

- informing about a failing in loading a preset
- informing about a wrong locale which causes the inability of all spinbuttons

The problem is that they should have a visible window to be shown, but we might trigger them while the service is running and the window is hidden. So we have to find a way to detect if the window is visible or not. I though something like 

```
if (window_pointer == nullptr) {
  return;
}

show_toast();
```

But don't know if it's really possible,

Regarding the Toast construction, we could define an helper which takes a string as an argument and construct the widget at the moment. That string will be also translatable. Maybe additional optional parameters could be added if we need to automatically hide the Toast (specifying the timeout) or let it close by the user.

wwmm:
>The problem is that they should have a visible window to be shown, but we might trigger them while the service is running and the window is hidden.

Assuming we find a decent way to make the signal connections it should not be a problem because the connections would be made only if the window is created. The problem is where to make the connections...

vchernin:
Toasts are nice but I think for errors requiring the user to be aware, like a preset not loading maybe a [GtkMessageDialog](https://docs.gtk.org/gtk4/class.MessageDialog.html) would be better. It is much more visible and they can show more information. I used one for the the portal errors that can occur in the Flatpak with the background portal.

Digitalone1:
> > The problem is that they should have a visible window to be shown, but we might trigger them while the service is running and the window is hidden.
> 
> Assuming we find a decent way to make the signal connections it should not be a problem because the connections would be made only if the window is created. The problem is where to make the connections...

The main window should be involved somehow. But why the signal? Can't we use a method call? That would be a metter of exposing the object of the main window. Sorry if I'm wrong, but on GTK windowing system I still have to learn.

Digitalone1:
[...]

PR #1616, opened 2022-07-01
Test Message Dialogs

Digitalone1:
Related to #1607 

I wanted to show the Toasts on preset loading error, but looking at their implementation I found out that they need a GTkOverlay like the AdwStatusPage, but this way they would be covered by the Preset Menu widget because that widget is on top of the main window where we are supposed to show the toast.

So I though we need something that stays on top of everything even if it's constructed from the main window which could be covered by other widgets (preset menu, blocklist menu, etc...).

So I chose the AdwMessageDialog, but only during compilation I found out it is available from libAdwaita 1.2 (we're at 1.1 I think). Anyway I implemented the GtkMessageDialog. It's not bad. I tested it. It's working and it appears on top of everything. As suggested by the docs, I set the modal and transient property.

Initially I had issues in getting the signal from the preset manager, but I managed to use the realize function in application_ui. If there is a better and safe method, @wwmm you can refine it. In the message dialog setup I added a null check on the parent widget: maybe it's useless, you can remove it.

In preset manager notify_error I pass the preset name, but unfortunately it's the tag for the json preset. Can it be translated somehow to a user friendly name like in the Effects List? In the message dialog it's shown `multiband_compressor: Wrong Format`, it's not the best, `Multiband Compressor: Wrong Format` is more appropriate.

I forgot to test while the window is hided. You can do it, I will try tomorrow. The preset autoload on the device switch is the only way to trigger the change while the window is hided?

### DIFF
--- a/include/presets_manager.hpp
+++ b/include/presets_manager.hpp
@@ -71,6 +71,9 @@ class PresetsManager {
     plugin_generic
   };
 
+  // signal sending title and description strings
+  sigc::signal<void(const std::string, const std::string)> preset_load_error;
+
   auto get_names(const PresetType& preset_type) -> std::vector<std::string>;
 
   auto search_names(std::filesystem::directory_iterator& it) -> std::vector<std::string>;
@@ -166,5 +169,5 @@ class PresetsManager {
 
   auto load_blocklist(const PresetType& preset_type, const nlohmann::json& json) -> bool;
 
-  void notify_error(const PresetError& preset_error);
+  void notify_error(const PresetError& preset_error, const std::string& plugin_name = "");
 };
--- a/include/presets_manager.hpp
+++ b/include/presets_manager.hpp
@@ -62,6 +62,15 @@ class PresetsManager {
   auto operator=(const PresetsManager&&) -> PresetsManager& = delete;
   ~PresetsManager();
 
+  enum class PresetError {
+    blocklist_format,
+    blocklist_generic,
+    pipeline_format,
+    pipeline_generic,
+    plugin_format,
+    plugin_generic
+  };
+
   auto get_names(const PresetType& preset_type) -> std::vector<std::string>;
 
   auto search_names(std::filesystem::directory_iterator& it) -> std::vector<std::string>;
@@ -156,4 +165,6 @@ class PresetsManager {
   void save_blocklist(const PresetType& preset_type, nlohmann::json& json);
 
   auto load_blocklist(const PresetType& preset_type, const nlohmann::json& json) -> bool;
+
+  void notify_error(const PresetError& preset_error);
 };
--- a/src/application_ui.cpp
+++ b/src/application_ui.cpp
@@ -34,6 +34,8 @@ struct Data {
   GApplication* gapp;
 
   GtkIconTheme* icon_theme;
+
+  std::vector<sigc::connection> connections;
 };
 
 struct _ApplicationWindow {
@@ -87,6 +89,32 @@ auto setup_icon_theme() -> GtkIconTheme* {
   gtk_icon_theme_add_resource_path(icon_theme, tags::resources::icons);
 
   return icon_theme;
+}
+
+void setup_simple_message_dialog(GtkWidget* parent, const std::string& title, const std::string& descr) {
+  if (parent == nullptr) {
+    return;
+  }
+
+  // Modal flag prevents interaction with other windows in the same application
+  auto* dialog = gtk_message_dialog_new(GTK_WINDOW(parent),
+                                        static_cast<GtkDialogFlags>(GTK_DIALOG_DESTROY_WITH_PARENT | GTK_DIALOG_MODAL),
+                                        GTK_MESSAGE_ERROR, GTK_BUTTONS_CLOSE, title.c_str());
+
+  gtk_message_dialog_format_secondary_text(GTK_MESSAGE_DIALOG(dialog), descr.c_str());
+
+  // Destroy the dialog when the user responds to it
+  g_signal_connect(dialog, "response", G_CALLBACK(gtk_window_destroy), NULL);
+
+  // Keep the dialog on top of the main window, or center the dialog over the main window
+  gtk_window_set_transient_for(GTK_WINDOW(dialog), GTK_WINDOW(parent));
+
+  /* Version with Adw.MessageDialog available from libAdwaita 1.2
+  auto* dialog = adw_message_dialog_new(GTK_WINDOW(parent), title.c_str(), descr.c_str());
+
+  adw_message_dialog_add_response(ADW_MESSAGE_DIALOG(dialog), "close", "OK"); */
+
+  gtk_window_present(GTK_WINDOW(dialog));
 }
 
 void apply_css_style() {
@@ -158,9 +186,23 @@ void realize(GtkWidget* widget) {
   ui::effects_box::setup(self->soe_ui, app::EE_APP(self->data->gapp), PipelineType::output, self->data->icon_theme);
   ui::effects_box::setup(self->sie_ui, app::EE_APP(self->data->gapp), PipelineType::input, self->data->icon_theme);
   ui::pipe_manager_box::setup(self->pm_box, app::EE_APP(self->data->gapp));
+
+  self->data->connections.push_back(
+      app::EE_APP(self->data->gapp)
+          ->presets_manager->preset_load_error.connect([=](const std::string title, const std::string descr) {
+            setup_simple_message_dialog(widget, title, descr);
+          }));
 }
 
 void unrealize(GtkWidget* widget) {
+  auto* self = EE_APP_WINDOW(widget);
+
+  for (auto& c : self->data->connections) {
+    c.disconnect();
+  }
+
+  self->data->connections.clear();
+
   g_signal_handlers_disconnect_by_func(gtk_native_get_surface(GTK_NATIVE((widget))),
                                        reinterpret_cast<gpointer>(surface_state_changed), widget);
 
--- a/src/presets_manager.cpp
+++ b/src/presets_manager.cpp
@@ -326,17 +326,15 @@ auto PresetsManager::load_blocklist(const PresetType& preset_type, const nlohman
       } catch (const nlohmann::json::exception& e) {
         g_settings_reset(sie_settings, "blocklist");
 
-        util::warning(e.what());
+        notify_error(PresetError::blocklist_format);
 
-        util::warning(
-            "A parsing error occurred while trying to load the blocklist inside the preset. The file could be invalid "
-            "or corrupted. Please check its content.");
+        util::warning(e.what());
 
         return false;
       } catch (...) {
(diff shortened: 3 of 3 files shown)


Digitalone1:
Oh, next step would be to show messages when the load fails on APO preset, Rnnoise model and Convolver impulse response loading.

@wwmm considering the work on plugin multiple instances, do you have recommendations on their implementation?

wwmm:
When I tried to generate an Arch Linux package there was a compilation error
```
error: format not a string literal and no format arguments [-Werror=format-security]
```
It was the first time I saw something like this. The crash only happened when the pkgbuild tried to compile. It was fine outside of it. I imagine it is some kind of security check Arch Linux enables. Fortunately clang-tidy also complained about the security problem and suggested an easy fix. It is fine now in the master branch.

>but this way they would be covered by the Preset Menu widget because that widget is on top of the main window where we are supposed to show the toast.

I did not think about this. Toasts are not going to play nice with popover menus in this situation.

>So I chose the AdwMessageDialog, but only during compilation I found out it is available from libAdwaita 1.2 (we're at 1.1 I think). 

As libadwaita 1.2 is not released yet it is better not using this class for now.

>Initially I had issues in getting the signal from the preset manager, but I managed to use the realize function in application_ui. If there is a better and safe method, @wwmm you can refine it.

I think it is as good as it can get. For some reason some of the needed objects are not available in the widget init function.

>Can it be translated somehow to a user friendly name like in the Effects List?

If I understood what you want we already do something like this in the plugins menu https://github.com/wwmm/easyeffects/blob/dfaaf9076f78d6669c99849322038659a31d85d7/src/tags_plugin_name.cpp#L24

>The preset autoload on the device switch is the only way to trigger the change while the window is hided?

No. Presets can be loaded from the command line too `easyeffects -l preset_name`

>considering the work on plugin multiple instances, do you have recommendations on their implementation?

Unless I have missed something I do not think it matters. The same approach should work for multiple instances too. Something that is unlikely to happen but that may happen is that different plugins try to show the dialag warning at the same time or in a row. For example if the convolver fails and the rnnoise too. I do not remember how gtk deals with the idea of multiple attempts to show a dialog at the same time. But if we are lucky it does something smart.

In any case even without multiple instances this can happen.

Digitalone1:
> I do not remember how gtk deals with the idea of multiple attempts to show a dialog at the same time. But if we are lucky it does something smart.

I already reproduced this kind of behavior. If I don't close the dialog and I trigger another load error, another dialog is placed on top of the previous. Then I should close two dialogs. Not a big issue I think.

Digitalone1:
> Unless I have missed something I do not think it matters. The same approach should work for multiple instances too.

Are you sure? I have issues in catching the plugin signals inside the application_ui class. 

For example, If I have the signal emitting in the compressor, I could find a way to get it inside the application_ui, but how this will work with multiple instances? Now we have the compressor pointer which is only one, but how can I connect the signals when there will be multiple compressor instances in the future?

Maybe making the signals as static members will help?

wwmm:
Without having multiple instances working (we are very close to it) it is a little hard to be sure. But I think it is just a matter of connecting the signals in the for loop inside plugins_box.cpp where the plugins interface is created. The idea is that inside this for loop each instance will make its own signal connection.

Digitalone1:
So everytime the plugin stack is changed, the signals have to be disconnected and reconnected?

I don't know, there's not a more efficient way?

I implemented the alert construction in the application_ui, but from application_ui there's know way I can know if the plugins are rearranged.

With plugin manager has been easy because it's just one instance.

wwmm:
>So everytime the plugin stack is changed, the signals have to be disconnected and reconnected?

In this particular case we can just call the signal `clear` method inside the plugin ui `dispose` method. This should remove any slot connected to the signal.

>I don't know, there's not a more efficient way?

I don't know. But I think the first question is if connecting/disconnecting is a heavy operation. I never looked at the libsigc++ implementation but I imagine this should not be resource intensive.

>I implemented the alert construction in the application_ui, but from application_ui there's know way I can know if the plugins are rearranged.

I don't think `application_ui` has to know about this when signals are used. When the plugin is created it will connect its signal to a slot provided by application_ui. And once the plugin is gone it clears this connection calling `clear`. The application_ui code does not need to know about this as long as the signal gives all the necessary information when the slot is called.

wwmm:
>Application UI is connecting and disconnection the signal which is emitted by Plugin Manager. 

Oh... I see. I agree that if application_ui is the one connecting it will be harder. But hopefully it won't have to be this way with the plugins instance. 

>I also think Application UI is not the right place to have the dialog construction. Maybe it's better in UI Helpers?

Assuming that the helper will have an easy way to get the main window pointer I do not see a problem in constructing the dialog somewhere else.

wwmm:
>Is there a way I can pass the window pointer to plugin_base?

Ideally the plugin_base class should not be touching the window pointer. It is better to define a signal in plugin base that will be inherited by all the plugin classes and connecting this signal to a slot that can see the main window pointer returned by `gtk_application_get_active_window`

Digitalone1:
It's very tough, I'm not sure about this and I discovered also a serious bug. When the error happens at the autoloading at window launch, two alerts are shown, but I can't close the first one with the left click. The only way to close it is typing the Enter key.

I don't know the reason of this behavior but it's serious since lots of users could have this bug with the new Gate which will fail to load the preset at window launch. I'm sure the previous time I tested I could close two alerts with the left click, but this time I can only close the first with the Enter key, otherwise it stays there blocking the whole app. 

It could be a GTK issue, but I'm considering to give up and remove dialog messages. It's to difficult and GTK is not helping us on this issue. Will find another easy way or we will just stay without them. :(

wwmm:
> but this time I can only close the first with the Enter key, otherwise it stays there blocking the whole app.

This is a problem indeed. Something else I did not think before is that gtk4 still has problems with dialog windows on Xorg. It may be wise to delay the use of these messages. If you want we can keep the code you have already written there but keep it disabled until the situation is more stable.

Digitalone1:
Sure I will disable it.

wwmm:
Something else we could consider is using the overlay messages we gave up on using before. Maybe there is a way to make some warnings "permanent" until the user closes then by clicking on the typical "x button". This way he/she would se the warning even if initially the preset menu is the one in the foreground.

Digitalone1:
I could try to make a specific toast inside the preset menu. I gave up on toasts because they were covered by the preset menu if they were shown on the main window, but maybe it's doable inside the preset menu.

Anyway I'm afraid the toast will stretch the plugin menu until it's visible. Not a big result if it's so.

wwmm:
>I gave up on toasts because they were covered by the preset menu if they were shown on the main window, but maybe it's doable inside the preset menu.

They will be partially covered. I am not sure about how you are visualizing them but I thought about something like this https://developer.gnome.org/hig/patterns/feedback/info-bars.html being show in an overlay near or over the bottom bar. The user should be able to see it even if the menu is visible.

>Anyway I'm afraid the toast will stretch the plugin menu until it's visible. Not a big result if it's so.

I think that the advantage of having this in the main window bottom is that there is the chance we can show any warning there. If we do it on the preset menu we will be following a path were each warning will have its own solution. In the end it may require more work.

Digitalone1:
The overlay solution could be an option, but at this point I'd like to avoid to rework all these things if I have to make an attempt to not disable this feature. So I thought about two workarounds:
- defining a callback to hide the alert when clicking the close button - maybe this works
- do not modify the code and change only the button label as `Close (press ESC)` -  this works

If we are lucky when libadwaita 1.2 will be released, we could use AdwMessageDialog which could be more stable.

wwmm:
>If we are lucky when libadwaita 1.2 will be released, we could use AdwMessageDialog which could be more stable.

Maybe we should wait until libadwaita 1.2 is released before putting energy in these messages.

Digitalone1:
So a callback is already defined to destroy the window. The remaining option is to change the button label `Close (Press ESC)`. This way the user know he/she has to press ESC if the dialog does not hide.

It's good to me not putting energy in these dialogs, but the work is done, at this point it's worth to keep them with the custom label, maybe with time passing GTK will fix the issue itself or migrating to AdwMessageDialog with libAdwaita 1.2 will make things more stable.

image


But it's up to you. If you don't like it, I can remove everything.

Digitalone1:
Tested on X11, it's working good there. Clicking on the button closes the most outer dialog, so I can close the other one too. It's an issue only on Wayland and only on the most outer dialog.

Anyway I wonder why two dialogs are shown at startup. One for the loading of the latest settings and one for the autoloading of the preset based on the device used? Can we reduce this steps at startup loading only the autoloading preset? Only if it's set, otherwise we rely on the latest used settings...

Digitalone1:
I nearly did the toast for Equalizer APO, but the opacity is not full when I set the overlay type. Why?

image

If I remove the overlay type I get the toast as a simple children which is hidden when I close it, but I'd like to get it over the widget with full opacity. 

image

I will make the MR, maybe you can refine it

wwmm:
>I nearly did the toast for Equalizer APO, but the opacity is not full when I set the overlay type. Why?

Initially I tried with a light theme. So it did not look so bad. I agree that there is too much transparency. Maybe it is worth to open an issue in the libadwaita page asking about this so we can solve it in the proper way.

wwmm:
>Anyway can we do something to load only the preset autoloading at startup? It's not for this bug (it has to be fixed by GTK devs). It's that having two loads does not seem efficient to me.

I did not understand. Two loads? Initially the widgets will be initialized with the last used settings. If a device with an autoloading profile is available the corresponding preset will be triggered at some point. In the first case these settings do not come from a preset being loaded but from the binding to the GSettings database.

Digitalone1:
> > Anyway can we do something to load only the preset autoloading at startup? It's not for this bug (it has to be fixed by GTK devs). It's that having two loads does not seem efficient to me.
> 
> I did not understand. Two loads? Initially the widgets will be initialized with the last used settings. If a device with an autoloading profile is available the corresponding preset will be triggered at some point. In the first case these settings do not come from a preset being loaded but from the binding to the GSettings database.

You're right, so why we have two dialogs at startup when the preset load fails? It seems two presets are loaded. And I have presets only for the output steam. I'll try to investigate on this.

wwmm:
>so why we have two dialogs at startup when the preset load fails?

Good question. What kind of changes are you doing to the preset file to force a failure?

wwmm:
The reason why the property is being set to an empty string when EE starts is https://github.com/wwmm/easyeffects/blob/f4ed2005477672bda8232df71bbf6c6fa9663c1f/src/stream_output_effects.cpp#L25. The logic in this part of our code is this way since Pulseaudio and it has to change. We do not know the default device at this stage. Maybe we did in Pulseaudio but it is not the case in PipeWire...

Digitalone1:
Nice you fixed the issue. The problem of the dialogs is present only on Wayland at startup when two dialogs are shown.

If you autoload a corrupted preset switching from speakers to headphones many times, you can spawn multiple dialogs, but you can close them with the click normally (even on Wayland). The issue is only at startup.

PR #1622, opened 2022-07-03
Multiband gate: fix annoying schema bug

LebedevRI:
The enum `com.github.wwmm.easyeffects.multibandgate.detection.enum`
is defined in this schema, but the enum `com.github.wwmm.easyeffects.gate.detection.enum`
is being used instead :) Thus, the schema is not hermetic,
and leads to really hard-to-diagnose bugs (if you don't know where to look)
when changing the plain gate scheme.

This took me way too long to understand...


wwmm:
Thanks! Sometimes it is amazing how much small mistakes can be done when copying and pasting code. And also how they can silently work :smile:

LebedevRI:
Thank you.

Yes, this was not fun at all to diagnose :/

PR #1623, opened 2022-07-03
Various improvements

Digitalone1:
Changes to preset error dialogs and locale handling.

Added also a `str_contains` helper.


Digitalone1:
@wwmm I saw you used also `rfind` comparing against 0. Maybe it's better to use `starts_with`?

PR #1627, opened 2022-07-04
`LSP Gate` implementation (Fixes #1596)

### LINKED ISSUE #1596: Feature request: LSP Gate
Hi!

Now that easyeffects has been packaged in debian, i was finally able to migrate
to pipewire and not-an-ancient version of easyeffects :)

The current `Gate` Effect is based on Calf Studio Gear.
LSP also implements a Gate Effect: https://lsp-plug.in/?page=manuals&section=gate_stereo

While i don't know which one is better/worse, there is one 'big' thing
i'm missing in the current one - separate `Threshold`s for `Attack` and `Release`.
While i have filed an upstream issue https://github.com/calf-studio-gear/calf/issues/314,
it may take an unspecified amount of time for that to be implemented.

At the same time, i do believe LSP *does* support that already:
```
* Hysteresis L, R - enables additional hysteresis curve:
* * Thresh - threshold (relative to basic curve), the level of input signal at which the gate starts to close.
* * Zone - the size of the transition zone below the hysteresis threshold.
```

I see that the `Compressor` was recently ported to use LSP: https://github.com/wwmm/easyeffects/pull/1056
What are general thoughts on doing the same for `Gate`?

LebedevRI:
This ended up being rather boringly straight-forward
after figuring out https://github.com/wwmm/easyeffects/pull/1622.
I started by fully copying the `Compressor` effect,
which is also LSP-based, and is pretty modern,
and then changed things.

The `Sidechain`/`Sidechain Filters` tabs are exactly identical
as they are in the `Compressor`, and in general most of the code
is rather exactly identical to `Compressor`.
Seems like some common boilerplate could/should be extracted.

I'm sure i've missed something, but this is basically it.
image

Fixes https://github.com/wwmm/easyeffects/issues/1596

### DIFF
--- a/data/schemas/com.github.wwmm.easyeffects.gate.gschema.xml
+++ b/data/schemas/com.github.wwmm.easyeffects.gate.gschema.xml
@@ -1,12 +1,26 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <schemalist>
-    <enum id="com.github.wwmm.easyeffects.gate.detection.enum">
-        <value nick="RMS" value="0" />
-        <value nick="Peak" value="1" />
+    <enum id="com.github.wwmm.easyeffects.gate.sidechain.input.enum">
+        <value nick="Internal" value="0" />
+        <value nick="External" value="1" />
     </enum>
-    <enum id="com.github.wwmm.easyeffects.gate.stereo.link.enum">
-        <value nick="Average" value="0" />
-        <value nick="Maximum" value="1" />
+    <enum id="com.github.wwmm.easyeffects.gate.sidechain.mode.enum">
+        <value nick="Peak" value="0" />
+        <value nick="RMS" value="1" />
+        <value nick="Low-Pass" value="2" />
+        <value nick="Uniform" value="3" />
+    </enum>
+    <enum id="com.github.wwmm.easyeffects.gate.sidechain.source.enum">
+        <value nick="Middle" value="0" />
+        <value nick="Side" value="1" />
+        <value nick="Left" value="2" />
+        <value nick="Right" value="3" />
+    </enum>
+    <enum id="com.github.wwmm.easyeffects.gate.filter.mode.enum">
+        <value nick="off" value="0" />
+        <value nick="12 dB/oct" value="1" />
+        <value nick="24 dB/oct" value="2" />
+        <value nick="36 dB/oct" value="3" />
     </enum>
     <schema id="com.github.wwmm.easyeffects.gate">
         <key name="bypass" type="b">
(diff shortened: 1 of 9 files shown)


Digitalone1:
Very nice @LebedevRI 

At first sight I would make a change to the UI aligning all the spinbuttons and letting the Curve and Hyteresis buttons to be above them. I will make a review after the merge.

@wwmm I suggest to merge this in a separate branch and wait to fix the locale issue for a new 4.6.7 release. Then, after a period of testing, we release 4.7 with the new Gate implementation. Maybe also with the Multiband Gate from LSP. What do you think?

LebedevRI:
> At first sight I would make a change to the UI aligning all the spinbuttons and letting the Curve and Hyteresis buttons to be above them.

I'm not sure i understand. The current UI is the way it is intentionally,
Essentially, there is only a single spinbutton for Attack time / Release Time / Reduction / Makeup,
BUT there **are** two spinbuttons per the Curve and Hysteresis _groups_,
so that is why i have grouped Threshold and Zone under the appropriate label.
Likewise, only Hysteresis is a button, Curve is not a button.

Digitalone1:
> I don't want to sound dismissive, but this seems a bit bikesheddy, because there is currently no such handling anywhere in EE, and i think introducing it should not be jumbled together with implementation of a new effect. 

Why not? We would introduce it anyway, so starting from a new effect implementation wouldn't be a bad idea. 

> Presumably it would be provided for _all_ effects, not only the ones for which the underlying implementation provides it

I don't think so. We rely on lv2 controls except for output and input gains, so if there are dry/wet controls we expose them, otherwise not.

wwmm:
>let me know if/how i should change the UI.

I am doing some tests and I think it is fine to keep it as it is. I will merge it to the master branch.

I do not see any problem in having wet/dry controls only in some of the plugins. For the ones that do not have this natively like the noise remover we can always try to add them later.

LebedevRI:
> > I don't want to sound dismissive, but this seems a bit bikesheddy, because there is currently no such handling anywhere in EE, and i think introducing it should not be jumbled together with implementation of a new effect.
> 
> Why not? We would introduce them anyway, so starting from a new effect implementation wouldn't be a bad idea.

Because it is orthogonal to the issue at hand, and comes with a bag of new considerations:
* how do we display those controls?
* where do we display those controls?
* do we only support those if the underlying implementation supports them? why/why not?
* do we really want to just pass-through them to the underlying implementation? Why aren't we doing the same for input/output gain? 
* If we want to support them for all effects, don't we always want to deal with them manually, like we deal with gain?
* If so, how do we implement them?

It's a huge mess, i don't see why it's productive to deal with that here.

> > Presumably it would be provided for _all_ effects, not only the ones for which the underlying implementation provides it
> 
> I don't think so. We rely on lv2 controls except for output and input gains, so if there are dry/wet controls we expose them, otherwise not.
> 
> Edit: @wwmm could find a way to expose them also for plugins which do not provide such controls, but if the plugin already provide them internally, we just use them.

LebedevRI:
@wwmm thank you! Let me know which changes i should make.
FWIW, i would like to help further, i just don't like convoluted overextended/overstretched messes..

wwmm:
>FWIW, i would like to help further, i just don't like convoluted overextended/overstretched messes..

No problems. My point is that depending on how we look at it EasyEffects already has some "level of mess". So a little more of it isn't a big deal in my opinion :smile:. Help is always welcome. Specially in the issues where I do not have a good idea about how to solve :-)

LebedevRI:
One thing i didn't really try to do is, we could display 3 more meters:
https://github.com/lsp-plugins/lsp-plugins-gate/blob/d4b1df23d5b49bb7272f9eb5cd1122ff1cf23d78/src/main/meta/gate.cpp#L134-L136
* `Zone start` (which is just `Curve Threshold` + `Curve Zone`)
* `Hysteresis threshold start` (which is just `Curve Threshold` + `Hysteresis Threshold`)
* `Hysteresis zone start` (which is just `Curve Threshold` + `Hysteresis Threshold` + `Hysteresis Zone`)

Doing so could potentially result in nicer UX, but i didn't try yet so see where to display them.

wwmm:
>Doing so could potentially result in nicer UX (just reading the label is simpler than doing math in the head :)),
but i didn't try yet so see where to display them.

It makes sense to display these levels. At least initially I would just put them in a second line of labels with levels. Currently there are four and the second line would have only 3. So there is some asymmetry... But I think it is the easier thing to do.

LebedevRI:
> > Doing so could potentially result in nicer UX (just reading the label is simpler than doing math in the head :)),
> > but i didn't try yet so see where to display them.
> 
> It makes sense to display these levels. At least initially I would just put them in a second line of labels with levels. Currently there are four and the second line would have only 3. So there is some asymmetry... But I think it is the easier thing to do.

Done, https://github.com/wwmm/easyeffects/pull/1629. Unlike the original line of labels, these are only displayed for the `Gate` page.

Digitalone1:
> * how do we display those controls?

With spinbuttons. Not a big deal.

> * where do we display those controls?

Somewhere in the plugin interface.

> * do we only support those if the underlying implementation supports them? why/why not?
> * do we really want to just pass-through them to the underlying implementation? Why aren't we doing the same for input/output gain?
> * If we want to support them for all effects, don't we always want to deal with them manually, like we deal with gain?
> * If so, how do we implement them?

Input and output gains are different. They are essential. Dry and wet are two controls not really useful (in fact we never had them) which we can expose for the sake of giving the user more options.

If I don't remember bad, we used the internal controls for input/output gains with PulseEffects, but they where different across plugins because certain were -/+ 20 db, others -+ 60 db, so making them on their own added a sort of uniformity. But this is not really needed for dry/wet, so if they are available internally, it's worth to use them as they are,

LebedevRI:
@Digitalone1 my point being, all these are design considerations that need to be decided upon.
"Just implement it along the way" doesn't seem feasible.

PR #1629, opened 2022-07-05
LSP Gate: display more meters

LebedevRI:
I believe, exposing these labels leads to more understandable UI.
* `Attack Zone Start` (which is `Curve Threshold` + `Curve Zone`)
  The Gate begins to open when the `Sidechain` level becomes above the `Attack Zone Start` level.
* `Attack Threshold` (which is just `Curve Threshold`)
  The Gate fully opens when the `Sidechain` level becomes above the `Attack Threshold` level.
* `Release Zone Start` (which is either just `Curve Threshold` or `Curve Threshold` + `Hysteresis Threshold`)
  The Gate begins to close when the `Sidechain` level becomes below the `Release Zone Start` level.
* `Release Threshold` (which is either `Curve Threshold` + `Curve Zone`, or  `Curve Threshold` + `Hysteresis Threshold` + `Hysteresis Zone`)
  The Gate fully closes when the `Sidechain` level becomes below the `Release Threshold` level.

https://github.com/lsp-plugins/lsp-plugins-gate/blob/d4b1df23d5b49bb7272f9eb5cd1122ff1cf23d78/src/main/meta/gate.cpp#L134-L136

image

### DIFF
--- a/data/ui/gate.ui
+++ b/data/ui/gate.ui
@@ -354,6 +354,98 @@
                                         </child>
                                     </object>
                                 </child>
+
+                                <child>
+                                    <object class="GtkBox">
+                                        <property name="hexpand">1</property>
+                                        <property name="vexpand">0</property>
+                                        <property name="homogeneous">1</property>
+                                        <property name="margin-top">4</property>
+                                        <property name="margin-bottom">4</property>
+
+                                        <child>
+                                            <object class="GtkBox">
+                                                <property name="halign">center</property>
+                                                <child>
+                                                    <object class="GtkLabel" id="zone_start_title">
+                                                        <property name="halign">end</property>
+                                                        <property name="xalign">1</property>
+                                                        <property name="label" translatable="yes">Zone start</property>
+                                                    </object>
+                                                </child>
+
+                                                <child>
+                                                    <object class="GtkLabel" id="zone_start_label">
+                                                        <property name="width-chars">6</property>
+                                                        <property name="label">0</property>
+                                                    </object>
+                                                </child>
+
+                                                <child>
+                                                    <object class="GtkLabel">
+                                                        <property name="halign">start</property>
+                                                        <property name="label">db</property>
+                                                    </object>
+                                                </child>
+                                            </object>
+                                        </child>
+
+                                        <child>
+                                            <object class="GtkBox">
+                                                <property name="halign">center</property>
+                                                <child>
+                                                    <object class="GtkLabel" id="hysteresis_threshold_start_title">
+                                                        <property name="halign">end</property>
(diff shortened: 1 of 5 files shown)


LebedevRI:
On a second thought, the nomenclature seems contrived and backwards here.
What if we instead diverge from what LSP provides, and display 4 labels:
* `Attack Zone start` ("the gate begins to open", which is the current `Zone start` label, i.e. `Curve Threshold`+`Curve Zone`)
* `Attack Threshold` ("the gate fully opens", i.e. just literally the `Curve Threshold`)
* `Release Zone start` ("the gate begins to close", which is the current `Hysteresis threshold start` label, i.e. `Curve Threshold` + `Hysteresis Threshold`)
* `Release Threshold` ("the gate fully closes", which is the current `Hysteresis zone start` label, i.e.  `Curve Threshold` + `Hysteresis Threshold` + `Hysteresis Zone`)

?

LebedevRI:
I've tentatively applied that change, but in a separate commit, so it can be undone easily..

Digitalone1:
It's nice having them because I always had complications in understanding those controls. For replicating the Calf behavior of my main preset in LSP version I had to compare the curve in Calf native UI to look similar in the LSP counterpart.

Anyway It would be nice if you copy those information inside the manual (if you didn't do it already), so the user can read them.

LebedevRI:
@Digitalone1 Thank you for taking a look!

> It's nice having them because I always had complications in understanding those controls. For replicating the Calf behavior of my main preset in LSP version I had to compare the curve in Calf native UI to look similar in the LSP counterpart.
 
> Anyway It would be nice if you copy those information inside the manual (if you didn't do it already), so the user can read them.

As it can be seen from the diff, the manual is up-to-date, as far as i'm concerned at least..

Digitalone1:
Okay. I didn't see the part relative to the manual. That's nice.

Only one thing. We have the convention of showing the labels with all uppercase, so please change with `Attack Zone Start`. Thanks. :)

Digitalone1:
I'm testing it on my system. I have the following crash:

```
(easyeffects:3938): GLib-GIO-ERROR **: 21:07:33.460: Settings schema 'com.github.wwmm.easyeffects.gate' does not contain a key named 'curve_threshold'
```

I see we have the underscore as separator, but we usually don't use it in presets?

```
update_key<double>(json.at(section).at("gate"), settings, "hysteresis_threshold", "hysteresis-threshold");
```

I don't remember well these functions and their meaning, but watching other presets we use the hyphens as separator.

LebedevRI:
Posted https://github.com/wwmm/easyeffects/pull/1630

LebedevRI:
@wwmm thank you!

PR #1630, opened 2022-07-05
Gate: fix preset handling

LebedevRI:
Actually store `hysteresis` as a boolean
Don't use `_` separators, but `-`.

As noted in https://github.com/wwmm/easyeffects/pull/1629#issuecomment-1175411474.

### DIFF
--- a/src/gate_preset.cpp
+++ b/src/gate_preset.cpp
@@ -36,15 +36,15 @@ void GatePreset::save(nlohmann::json& json, const std::string& section, GSetting
 
   json[section]["gate"]["release"] = g_settings_get_double(settings, "release");
 
-  json[section]["gate"]["curve_threshold"] = g_settings_get_double(settings, "curve-threshold");
+  json[section]["gate"]["curve-threshold"] = g_settings_get_double(settings, "curve-threshold");
 
-  json[section]["gate"]["curve_zone"] = g_settings_get_double(settings, "curve-zone");
+  json[section]["gate"]["curve-zone"] = g_settings_get_double(settings, "curve-zone");
 
-  json[section]["gate"]["hysteresis"] = g_settings_get_boolean(settings, "hysteresis");
+  json[section]["gate"]["hysteresis"] = g_settings_get_boolean(settings, "hysteresis") != 0;
 
-  json[section]["gate"]["hysteresis_threshold"] = g_settings_get_double(settings, "hysteresis-threshold");
+  json[section]["gate"]["hysteresis-threshold"] = g_settings_get_double(settings, "hysteresis-threshold");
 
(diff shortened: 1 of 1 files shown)


src/gate_preset.cpp
```diff
@@ -36,15 +36,15 @@ void GatePreset::save(nlohmann::json& json, const std::string& section, GSetting
...
   json[section]["gate"]["release"] = g_settings_get_double(settings, "release");
 
-  json[section]["gate"]["curve_threshold"] = g_settings_get_double(settings, "curve-threshold");
+  json[section]["gate"]["curve-threshold"] = g_settings_get_double(settings, "curve-threshold");
 
-  json[section]["gate"]["curve_zone"] = g_settings_get_double(settings, "curve-zone");
+  json[section]["gate"]["curve-zone"] = g_settings_get_double(settings, "curve-zone");
 
-  json[section]["gate"]["hysteresis"] = g_settings_get_boolean(settings, "hysteresis");
+  json[section]["gate"]["hysteresis"] = g_settings_get_boolean(settings, "hysteresis") != 0;
```
LebedevRI, line 43: @Digitalone1 this was the real bug, this would store an integer, because GLib is C and not C++,
and `g_settings_get_boolean()` returns `gboolean` which is `typedef gint   gboolean;`.

LebedevRI:
Looking at the output, this results in the following json:
```
        "gate": {
            "attack": 20.0,
            "bypass": false,
            "curve-threshold": -24.0,
            "curve-zone": -6.0,
            "hpf-frequency": 11.0,
            "hpf-mode": "12 dB/oct",
            "hysteresis": false,
            "hysteresis-threshold": -12.0,
            "hysteresis-zone": -6.0,
            "input-gain": 1.0,
            "lpf-frequency": 19998.0,
            "lpf-mode": "12 dB/oct",
            "makeup": 0.0,
            "output-gain": 6.3,
            "reduction": -24.0,
            "release": 100.0,
            "sidechain": {
                "input": "External",
                "lookahead": 0.01,
                "mode": "Low-Pass",
                "preamp": 0.1,
                "reactivity": 10.01,
                "source": "Side"
            }
        },
```
So even though there's `sidechain` group, the low-/high- pass filtering of the sidechain is not in the `sidechain` group.
I assume it was an oversight in original Compression Effect. Should this be rectified, at least for the Gate?

wwmm:
>I assume it was an oversight in original Compression Effect. Should this be rectified, at least for the Gate?

Honestly I do not remember why I did it this way. Maybe because in LSP these controls were explained in a separated section. In any case I think it can stay as it is. Even if someone is editing the preset file manually it should still be easy to find the parameters.

LebedevRI:
Okay then.
Thank you!

PR #1637, opened 2022-07-06
Improve Gate UI and implement Toast for Equalizer UI

### LINKED ISSUE #1633: Improve Gate layout
image

@wwmm @LebedevRI What do you think? 

Curve and Hysteresis spinbuttons grouped with `linked` Adw style. The labels above enclosed in an homogeneous box. I like it more.

If it's good for you, I'll make the MR.

Digitalone1:
- Added workaround for preset error message dialogs #1616
- Implemented a Toast for APO load error in Equalizer UI
- Improved Gate UI @LebedevRI Fixes #1633

As discussed in #1633 I like the spinbuttons with the same height in one row. But I didn't implement it here. @wwmm please decide which is the better thing to do. 

Schermata del 2022-07-06 16-10-49

As discussed in #1616 I tested AdwToast and finally found a way to show them properly. I put the constructor in the ui_helpers, so we don't have to rewrite the method to show them over and over inside every plugin ui. 

image

It's good how it behave, but maybe it would be better if the background is full opaque? I don't know how to change it, and maybe other people like it this way. @wwmm let me know what do you think.

### DIFF
--- a/data/ui/equalizer.ui
+++ b/data/ui/equalizer.ui
@@ -5,377 +5,392 @@
         <property name="margin-end">6</property>
         <property name="margin-top">6</property>
         <property name="margin-bottom">6</property>
-        <property name="spacing">12</property>
         <property name="orientation">vertical</property>
         <child>
-            <object class="GtkBox">
-                <property name="spacing">12</property>
+            <object class="GtkOverlay" id="overlay">
+                <child type="overlay">
+                    <object class="AdwToastOverlay" id="toast_overlay">
+                        <property name="valign">start</property>
+                    </object>
+                </child>
+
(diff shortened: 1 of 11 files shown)


wwmm:
>t's good how it behave, but maybe it would be better if the background is full opaque? I don't know how to change it, and maybe other people like it this way. @wwmm let me know what do you think.

I think it is fine as it is. It is probably possible to force it to be fully opaque messing with css properties but I think that if we are going to use libadwaita it is better to just follow its default behavior. If we start to customize it we may make things look weird in user themes that expect libadwaita css to be in a certain way when customizing it.

Issue #1628, opened 2022-07-05
Exclude util/NEWS from translation templates


Digitalone1:
@wwmm @vchernin I already asked this, but was told it was used by Flatpak.

Unfortunately it's the nth time the translation templates are generated after the release, so the latest strings are still provided in English untranslated. 

We submit all those changelog strings which are useless to translate in my opinion. It's also discouraging for the potential volunteers because those strings are too many. We need more translations of control parameters, not the changelog.

It's the same issue about the documentation. It would be better to have it translated in all languages, but we know that's won't going to happen and maintaining all those translation templates is annoying. Presenting it to the volunteers on Weblate is also more discouraging because the amount is huge.

My advice is to remove the NEWS file from the translation templates, the number of translations will reduce dramatically and users will be more attracted to translate what really matters.

vchernin:
Note the util/NEWS file itself is not used by the translation template or by Flatpak, `data/com.github.wwmm.easyeffects.metainfo.xml.in` file is where the translatable strings are stored.

Also, this NEWS file or metainfo is not specific to Flatpak. Having good metainfo is useful since it shows up in software stores/viewers like the Flathub website or gnome software.

In my opinion it is good to have the metainfo file translated, at least the primary sections of it (i.e. everything except the release notes). The description of the app will be the first thing users will see when finding EasyEffects in their software store, it is good to have it translated.

I was thinking perhaps the translatable strings could be added long before release (from the news file or otherwise). The translatable file could also be automatically updated. This would mean there’d be less untranslated strings every release.

wwmm:
>The translatable file could also be automatically updated. This would mean there’d be less untranslated strings every release.

It would help with the current problem that is "me not remembering" to run the command that updates the template. But there would still be the problem about the amount of things needing translation. Based on the website traffic most countries are getting the changelogs in English anyway.

wwmm:
I think that for the changelogs translation to be actually helpfully we would need a team of dedicated translators working synchronized with us. But in practice it may take months until some strings are translated. By the time they are there isn't much point at reading the old changelog.

Even with weblate the only translator working ahead of the releases is @Digitalone1.

Digitalone1:
I don't see the point of having them translated. Who reads them? The description of the application is useful, but not the changelogs.

Digitalone1:
[...]

PR #1652, opened 2022-07-15
Convolver dialog and RNnoise toast

Digitalone1:
* Dialogs are shown when an impulse import operation fails inside the Convolver.
This can be tested importing an empty .irs file or a non-stereo .wav (I took from [here](http://www-mmsp.ece.mcgill.ca/Documents/AudioFormats/WAVE/Samples.html)).
* RNnoise now updates the labels reporting if the selected mode has been loaded.
In case of failing, an autohiding toast is also shown to inform the user that the format may be unsupported and the model has fallen back to the Standard one.  
* Fixed spinbuttons alignment in Gate UI.

@wwmm after this what do you think of a new release? Especially to fix the missing exception handling on locale object construction. 

Anyway before the new release, #1628 has to be resolved.

### DIFF
--- a/data/ui/gate.ui
+++ b/data/ui/gate.ui
@@ -38,54 +38,41 @@
                                         <property name="halign">center</property>
                                         <property name="column-spacing">18</property>
 
+                                        <!-- Attack/Release section -->
                                         <child>
-                                            <object class="GtkLabel" id="attack_time_label">
-                                                <property name="label" translatable="yes">Attack</property>
+                                            <object class="GtkBox">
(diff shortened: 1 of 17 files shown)


wwmm:
>after this what do you think of a new release? Especially to fix the missing exception handling on locale object construction.

I do not see any problem. We will never know the real source of the problem but at least we can avoid the unhandled exception.

>Anyway before the new release, https://github.com/wwmm/easyeffects/issues/1628 has to be resolved.

I have an idea. I will talk about it there.

PR #1654, opened 2022-07-17
Use yaml for release notes script

vchernin:
This should avoid any split strings ending up in the metainfo and subsequently the translation template. Now it works just like markdown does, to make a new bullet point/list entry you must put a new hyphen, a newline is not detected as a new bullet point.

I also made a bunch of other random fixes while I was at it, but the script still behaves overall the same as before.


vchernin:
Tested this a decent amount, it seems to work as expected

Digitalone1:
Nice, maybe to be merged after the new release, then we can work with the split of translation UI template and metainfo template.

wwmm:
>Tested this a decent amount, it seems to work as expected
>Nice, maybe to be merged after the new release, then we can work with the split of translation UI template and metainfo template.

Ok. I had no time at all to work on EasyEffects today. Maybe tomorrow's night or tuesday. As soon as I can I will do a new release.

wwmm:
After doing a release this PR conflicted with the master branch. I did some changes to fix the conflict but I see the following when running the update script in the `non release mode`
```
Create a new release? (y/n): n
[ INFO ] Not making a new release, just refreshing changelog and metainfo with current releases.
[ INFO ] The top entry of the NEWS.yaml file will not be used for refreshing, since that is supposed to be the next release.
[ INFO ] Removing template part from temporary NEWS.yaml file, since not generating new release
[ INFO ] Converting news to metainfo
[ INFO ] Succesfully converted news to metainfo
[ INFO ] Checking appstreamcli validate --pedantic
[ ERROR ] appstreamcli validate --pedantic failed 
[ ERROR ] appstreamcli: W: com.github.wwmm.easyeffects.desktop:48: tag-empty ul/li
   The mentioned tag is empty, which is highly likely not intended as it should have content.

✘ Validation failed: warnings: 1 
```
Any ideas @vchernin ?

vchernin:
It’s complaining about this empty bullet point: https://github.com/wwmm/easyeffects/blob/0612371dfd2e5807e80f3cadf2017ac1b7e44873/util/NEWS.yaml#L31. Remove and it shouldn’t complain.

wwmm:
>Remove it (and the above other notes header) and it shouldn’t complain.

Thanks!

PR #1656, opened 2022-07-18
LSP Limiter: threshold's step is 0.01

LebedevRI:
https://github.com/lsp-plugins/lsp-plugins-limiter/blob/d024e23f84c66b422d53490b314a3f06bc8525c5/include/private/meta/limiter.h#L67

Given that the output gain's step is also limited to 0.1,
you can't really steparound this.


LebedevRI:
Thanks!

Issue #1671, opened 2022-07-22
Add Dry/Wet for plugins which already have those knobs


Digitalone1:
We discussed about this in other topics.

![Schermata del 2022-07-22 20-10-12](https://user-images.githubusercontent.com/25790525/180500217-0be873c9-aa0f-4bdf-a4ff-63477ab4644e.png)

Differently by #1589 which misses them and it could have a solution made by us, other plugins (I think all of LSP, not Calf) already ship dry/wet and we could just expose them in our UI.

Anyway I see an issue in the way we will use to implement them. LSP uses -infinite dB for dry in the default state which means no input unprocessed signal is used. This could also be set for wet output processed signal. But we can't replicate these states because we use finite dB values for our settings.

We already had this kind of issue in the past with release threshold in the Compressor which is -inf as default in LSP. We couldn't make it and I set -120 db as default, while the lowest finite value in LSP UI is -79,9 dB (anyway from Ardour/Carla settings I think a more precise linear value could be set manually).

LSP uses a workaround to pass from -79,9 to -inf dB, but we can't do that. This means if we introduce those controls we will always have dry and wet mixed together, even if one of them is imperceptible. I think this is not the best solution, so we should find a better implementation.

Maybe we could try with percentages rather than dB levels? @wwmm Any idea?

Digitalone1:
Actually we already have those controls for Reverb, which I think is one of the few Calf plugins to have them.

```
	Port 11:
		Type:        http://lv2plug.in/ns/lv2core#ControlPort
		             http://lv2plug.in/ns/lv2core#InputPort
		Symbol:      amount
		Name:        Wet Amount
		Minimum:     0.000000
		Maximum:     2.000000
		Default:     0.250000

	Port 12:
		Type:        http://lv2plug.in/ns/lv2core#ControlPort
		             http://lv2plug.in/ns/lv2core#InputPort
		Symbol:      dry
		Name:        Dry Amount
		Minimum:     0.000000
		Maximum:     2.000000
		Default:     1.000000

```

Both can have -inf but the minimum we have is -100 db

wwmm:
>Maybe we could try with percentages rather than dB levels? @wwmm Any idea?

My initial though was just using a linear scale between `0.0` and `1.0` and moving on with our lives. In theory it may be possible to do some clever mapping in the spinbutton label formatting callback to show the `-Inf`. But I do not think it is worth the effort. In my opinion linear values with a step of `0.01` should be more than enough.

[...]

PR #1684, opened 2022-07-25
Modify dry/wet handling in Reverb and prepare for adding the same controls in other plugins

Digitalone1:
Related to #1671

This introduces a new percentage value for dry/wet controls. It has been implemented for Reverb which already have those knobs exposed in our UI, but it can be replicated for other plugins, especially ones who have the relative ports and we only have to expose them.

This substantially replicates what is done in LSP interfaces for some knobs, including dry/wet controls. We can adjust the values, but also set linear 0 (-infinity) excluding the signal totally. We can't replicate the text `-inf` in the spinbutton, but 0% gives the idea that nothing of the signal is used.

Mapping -100 db to 0 db range on a percentage scale should be quite straightforward of what is happening. 1% variation corresponds to 1 db variation. 100% is 0 db gain, 50% is -50 db, but 0% is -infinity.

Since we don't want to revolutionize the existing layout of the plugins, I just put the dry/wet spinbuttons on top of input/output gain sliders. This is what most of the native UI of lv2 plugins is doing, generally dry/wet knobs are packed near the in/out gain controls.

@wwmm let me know what do you think. Anyway this has to be merged AFTER the release because we want to test further, even if I measured the output gain with the LSP spectrum analyzer and 0% of dry and wet was reporting no signal at all.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 __pycache__
 _build
 build
+build-dir
 locale
 gschemas.compiled
 obj-*-linux-gnu
@@ -58,3 +59,7 @@ util/flatpak/build-dir
 util/flatpak/repo
 util/flatpak/.flatpak-builder
 .flatpak
+.flatpak-builder
+
+# explictly include flatpak patch directory which includes usually ignored patch files
+!util/flatpak/patch/**/*.patch
--- a/data/schemas/com.github.wwmm.easyeffects.reverb.gschema.xml
+++ b/data/schemas/com.github.wwmm.easyeffects.reverb.gschema.xml
@@ -35,13 +35,13 @@
             <range min="0" max="1" />
             <default>0.5</default>
         </key>
-        <key name="amount" type="d">
-            <range min="-100" max="6" />
-            <default>-12</default>
-        </key>
         <key name="dry" type="d">
-            <range min="-100" max="6" />
-            <default>0</default>
+            <range min="0" max="106" />
+            <default>100</default>
+        </key>
+        <key name="wet" type="d">
+            <range min="0" max="106" />
+            <default>88</default>
         </key>
         <key name="predelay" type="d">
             <range min="0" max="500" />
--- a/data/ui/reverb.ui
+++ b/data/ui/reverb.ui
@@ -193,77 +193,10 @@
                 </child>
 
                 <child>
-                    <object class="GtkLabel">
-                        <property name="label" translatable="yes">Wet</property>
-                        <layout>
-                            <property name="column">2</property>
-                            <property name="row">0</property>
-                        </layout>
-                    </object>
-                </child>
-                <child>
-                    <object class="GtkSpinButton" id="wet">
-                        <property name="halign">center</property>
-                        <property name="orientation">vertical</property>
-                        <property name="width-chars">10</property>
-                        <property name="digits">1</property>
-                        <property name="update-policy">if-valid</property>
-                        <property name="adjustment">
-                            <object class="GtkAdjustment">
-                                <property name="upper">106</property>
-                                <property name="value">100</property>
-                                <property name="step-increment">0.1</property>
-                                <property name="page-increment">1</property>
-                            </object>
-                        </property>
-                        <layout>
-                            <property name="column">2</property>
-                            <property name="row">1</property>
-                        </layout>
-                        <accessibility>
-                            <property name="label" translatable="yes">Wet Level</property>
-                        </accessibility>
-                    </object>
-                </child>
-
-                <child>
-                    <object class="GtkLabel">
-                        <property name="label" translatable="yes">Dry</property>
-                        <layout>
-                            <property name="column">3</property>
-                            <property name="row">0</property>
-                        </layout>
-                    </object>
-                </child>
-                <child>
-                    <object class="GtkSpinButton" id="dry">
-                        <property name="halign">center</property>
-                        <property name="orientation">vertical</property>
-                        <property name="width-chars">10</property>
-                        <property name="digits">1</property>
-                        <property name="update-policy">if-valid</property>
-                        <property name="adjustment">
-                            <object class="GtkAdjustment">
-                                <property name="upper">106</property>
-                                <property name="step-increment">0.1</property>
-                                <property name="page-increment">1</property>
-                            </object>
-                        </property>
-                        <layout>
-                            <property name="column">3</property>
-                            <property name="row">1</property>
-                        </layout>
-                        <accessibility>
-                            <property name="label" translatable="yes">Dry Level</property>
-                        </accessibility>
-                    </object>
-                </child>
-
-                <child>
                     <object class="GtkLabel" id="bass_cut_label">
                         <property name="label" translatable="yes">Bass Cut</property>
                         <layout>
-                            <property name="column">4</property>
+                            <property name="column">2</property>
                             <property name="row">0</property>
                         </layout>
                     </object>
@@ -285,7 +218,7 @@
                             </object>
                         </property>
                         <layout>
-                            <property name="column">4</property>
+                            <property name="column">2</property>
                             <property name="row">1</property>
                         </layout>
                         <accessibility>
@@ -298,7 +231,7 @@
                     <object class="GtkLabel" id="treble_cut_label">
                         <property name="label" translatable="yes">Treble Cut</property>
                         <layout>
-                            <property name="column">5</property>
+                            <property name="column">3</property>
                             <property name="row">0</property>
                         </layout>
                     </object>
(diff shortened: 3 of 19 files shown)


Digitalone1:
Screenshot:

image

wwmm:
It is working. But I think there is no need to have decibel values in the way if in the window we are showing percentage values. In the file `/usr/lib/lv2/calf.lv2/Reverb.ttl` there is
```
     ] , [
        a lv2:InputPort ;
        a lv2:ControlPort ;
        lv2:index 11 ;
        lv2:symbol "amount" ;
        lv2:name "Wet Amount" ;
        lv2:default 0.250000 ;
        lv2:minimum 0 ;
        lv2:maximum 2 ;
    ] , [
        a lv2:InputPort ;
        a lv2:ControlPort ;
        lv2:index 12 ;
        lv2:symbol "dry" ;
        lv2:name "Dry Amount" ;
        lv2:default 1 ;
        lv2:minimum 0 ;
        lv2:maximum 2 ;
    ] , [
```
We can have the linear minimum and maximum values directly in our schema file and just calculate the percentage when showing them in the window. This way there will be no need to convert values inside `bind_key_double_percent_db`. And in this case this function can be renamed to something like `bind_key_double_percent`.

In other words there is no need for `util::db_percent_to_linear` in `lv2_wrapper`. Only the GtkSpinButton would need something like this when converting the linear values between `lv2:minimum` and `lv2:maximum` to `0% - 100%`.

wwmm:
>And in this case this function can be renamed to something like bind_key_double_percent.

Better than this. It can be replaced by `bind_key_double`. because from the database to the lv2_wrapper we can use only the linear values.

wwmm:
One disadvantage of doing the conversion from linear to percentage values and vice versa in the ui side is that the preset file would probably have a linear value instead of the percentage. It is not the end of the world but maybe it will be better to convert from percentage to linear inside lv2_wrapper.

In this case the schema would have the maximum and minimum values in percentage and no conversion for the GtkSpinButton would be needed. Only in lv2_wrapper to make the percentage linear.

Digitalone1:
@wwmm we never saved the linear value in gsettings, why should we do this for dry/wet? I think it's more straightforward having in the preset what is shown in the UI.

Take also in consideration that not working with the dB in this case could be misleading for the user because 100% means 6 dB gain while one might think it's just the plain signal with no gain.

Besides the for gain 0 db you have to set 50% which is way more misleading because that seems half the original signal, but it's just the plain signal.

Digitalone1:
If we want to map linear values reflecting 100% on 1 as 0 dB gain, we have something like this:

```
200% ==> +6 dB
100% ==> 0 dB
18% == around - 15 dB
```

It's very misleading. So if I want to set low levels round 30-40 db I have to tackle on 12-14% and it's not as precise as the current solution.

Digitalone1:
> It is ok. But I think I would prefer them in the same line as the others with that `linked` class to join them.

I can do it, but are you really sure? Having them linked will make the -/+ buttons of the dry value near the wet level. It could be misleading. Having them separate is more straightforward, but let me know, if you still prefer the linked class, I'll add it.

wwmm:
>because 100% means 6 dB gain while one might think it's just the plain signal with no gain.

From where is coming this "6 dB" gain? Is the Calf source code doing this? This is not how I would expect dry/wet processing to be done. In most definitions dry and wet controls just means that the output signal can be composed by a fraction of the input signal and the processed signal. Let's say I want an output signal that is `80% dry` and `20% wet`. I would expect it to do something like this with the linear values `output = 0.8 * dry + 0.2 * wet`. But now that there is a logarithmic scale in the middle something else is going to happen... 

Honestly I think that what you want to achieve only makes sense if we show decibel units in our window. I will always look at `50 %` there as meaning `0.5 * signal`. Having an invisible decibel scale behind the percentage value is not a good idea.

wwmm:
>Having them linked will make the -/+ buttons of the dry value near the wet level.

I think we may have visualized this in a different way. I was thinking about using the linked class with there controls in the vertical position like the others. Or in other words to do the same thing we did to the `Reduction` and `Makeup` controls in the gate window.

Digitalone1:
> From where is coming this "6 dB" gain? Is the Calf source code doing this? 

Dry and wet knobs in Reverb native UI are in dB. And, as I saw few days ago, the same is for LSP plugins as you can see the screenshot I posted in the relative issue.

You can check your own in carla-rack.

Schermata del 2022-07-26 15-06-04

I set 6 dB as maximum at both dry and wet.

Schermata del 2022-07-26 15-06-42

As you can see above, the parameters in the relative ports are linear and the max is 2. Just as the calculation done in our `db_to_linear` function.

> This is not how I would expect dry/wet processing to be done. In most definitions dry and wet controls just means that the output signal can be composed by a fraction of the input signal and the processed signal.

But it's not like you say for Calf Reverb and, as I saw few days ago, the same is happening in LSP plugins. The value is in dB. In Calf Reverb native UI you go down to -59,9 dB and the last step is `-inf` db (if you want any dB value lower than -60 db, you have to manually set the relative linear one in the parameters menu section provided by carla). Please just try it out.
 
> Honestly I think that what you want to achieve only makes sense if we show decibel units in our window. I will always look at `50 %` there as meaning `0.5 * signal`. Having an invisible decibel scale behind the percentage value is not a good idea.

The solution I did it's to allow the user to completely exclude the signal, which means setting the `-infinity` value. I'm not pretending my solution is the best, but since we can't show infinity on the spinbutton, I thought it would be good to map [-100,0] dB range on a percentage scale letting coincide 0% with -infinity db/0 linear.

If you think there's a better solution, I'm open to new ideas. Anyway it's sure the Calf Reverb expects linear values for those ports, so 2 is 6 dB, 1 is 0 dB and 0 is -infinity db.

Digitalone1:
@wwmm I just checked for LSP and also there the values are reported as dB inside the UI

Schermata del 2022-07-26 15-28-06

For LSP you go down -79,9 and the last step is -inf, the max is 20 db (for Calf is 6 instead).

You could also say, let just show them in dB as we already did for all other parameters, but I don't think it's a good idea because I want to completely exclude the dry in certain cases. And if we won't allow this, other people will complain in the issue section in the future, that's guaranteed.

So we have to find a solution to this issue.

wwmm:
>As you can see above, the parameters in the relative ports are linear and the max is 2. Just as the calculation done in our db_to_linear function.

That is not the problem. We are showing in the window percentage values for controls that have as their purpose to create a new signal based on a combination of processed and unprocessed signals. When I say `take 50 % of the signal` I expect half of it to be used in the computation and not something else.

Calf and LSP are showing decibels to the user. And this is fine because there is no percentage there. So we understand that `0 dB` in the dry control and let's say `-12 dB` in the wet control means "take the the whole input signal and mix it with with approximately 0.25 * (processed signal)". But in our case we have to remember that we are talking about a percentage of an invisible logarithmic scale. That is not the usual workflow.

Either we do the same Calf and LSP are doing or we show percentages of the linear values that are actually going to be used to multiply the wet/dry signals and computing the output signal. An invisible mapping from percentages to logarithmic values when we expect fractions of the signals to be combined will catch a lot of people off guard.

wwmm:
>The solution I did it's to allow the user to completely exclude the signal, which means setting the -infinity value.

But this is totally possible mapping the percentage value directly to the linear value given to the plugin. `0%` will be `0` what means no signal. There is no need for a logarithmic scale in the middle when showing percentages.

>Oh, I understood. Okay, I can do it, but which is the advantage of having them linked vertically?

I am not sure "advantage" is the better word. Almost everywhere we have the labels either above or before the widget. But the `Wet Level` label is coming after. It gave a strange feeling when compared to the others.

>Please take in consideration my idea of having them linked horizontally, which would consume less vertical space. But if you don't like it, I will change them to vertical as already done for makeup/reduction. Let me know.

Oh, no. I was thinking about having them in the same line as the other controls that are already in the vertical orientation. I agree that if we keep them next to the gain controls the vertical orientation will waste a lot of space. I get that most plugins keep these controls next to each other but do we really need to do the same?

wwmm:
>Either we do the same Calf and LSP

By the way why do we still think we can't do it? Thinking about this again we may be able to do the same through a custom parser like the ones we already have https://github.com/wwmm/easyeffects/blob/ebd13c7161e9fb29f5b70937582c9e01fa1eb133/src/ui_helpers.cpp#L100. I think that in the past we did not have any of those. Maybe that is why we have in our subconscious that we can't do it.

wwmm:
>Thinking about this again we may be able to do the same through a custom parser like the ones we already have 

Let me explain this a little more. If we use the linear values everywhere and do the decibel conversion only in the GtkSpinButton custom parsers I think there is a chance that we can map the `-Inf` label displayed by the spinbutton to the `zero` linear value.

wwmm:
But I still think that percentage values directly mapped to the linear port values are totally enough for what it is done with the wet and dry controls... Something similar already happens in the `Stereo Tools` plugin. The `balance control` is just a float in the range `[-1, 1]` and it is totally fine.

Digitalone1:
Okay, I will try to define a new parser having the spinbutton and the gsettings same as linear values in the plugins. Will let you know how it goes.

So, I think you should drop my commits because unfortunately this has to be reworked from scratch.

Digitalone1:
> Oh, no. I was thinking about having them in the same line as the other controls that are already in the vertical orientation. I agree that if we keep them next to the gain controls the vertical orientation will waste a lot of space. I get that most plugins keep these controls next to each other but do we really need to do the same?

Not really needed, but it's easier to put them there rather than changing the layout of the existing UI.

Anyway this is not a big issue. Just focus on the values handling. At the layout we will think about it later.

wwmm:
>So, I think you should drop my commits because unfortunately this has to be reworked from scratch.

Are you sure it is better to undo everything? I think most of what you did can be reused.

wwmm:
>I will try to define a new parser having the spinbutton and the gsettings same as linear values in the plugins.

Do you mean a new parser trying to map `0` to `-Inf` or just a parser to map the linear port values to percentage values? For the last case the parsers we already have are enough.

Digitalone1:
> Do you mean a new parser that will map `0` to `-Inf` or just a parser to map the linear port values directly to percentage values? For the last case the parsers we already have are enough.

Not sure yet, I will try different solutions. I will let you know which is more suitable. In case of issues, I'll report in #1671.

Digitalone1:
> Are you sure it is better to undo everything? I think most of what you did can be reused.

I won't delete the branch. I will do a new one where I could drop some commits. So I won't loose anything. It's better if you drop my commits here otherwise there will be conflicts on the new merge request.

wwmm:
>I won't delete the branch. I will do a new one where I could drop some commits. So I won't loose anything. It's better if you drop my commits here otherwise there will be conflicts on the new merge request.

Oh, I see. I will revert it then.

PR #1688, opened 2022-07-26
Create codeql-analysis.yml

wwmm:


vchernin:
FYI @wmmm you can just push changes to an existing branch and the tests would be rerun with the new workflow.

Since the actions runners either use ubuntu-20.04 or ubuntu-22.04 you will need to explictly build probably with an arch image or reusing the flatpak somehow.

vchernin:
> I've noticed that I would need to somehow put a docker Arch Image there but as I am not used to this I did not really understand how this could be done.

Yeah the arch docker image is used, the image itself is reasonably easy, the problem is all the small stupid issues. Eg makepkg not wanting to run as root, while the arch image by default gives you a root user.

wwmm:
>Eg makepkg not wanting to run as root, while the arch image by default gives you a root user.

I think I did not need the `makepkg` command in the circleci image. What did you need it for?

vchernin:
> What did you need it for?

My idea was to make/use a arch docker image for both producing the arch build (ie with makepkg), and to also have a up-to-date build environment to run codeql. 

I’m not sure how circleci works but unless you declared dependencies explicitly it would have needed to read the pkgbuild.

wwmm:
>My idea was to make/use a arch docker image for both producing the arch build (ie with makepkg), and to also have a up-to-date build environment to run codeql.

I see. I did not use our pkgbuild there. I just enter the build folder and run `ninja install`.

PR #1700, opened 2022-07-28
Add Dry and Wet for Delay Comp

### LINKED ISSUE #1671: Add Dry/Wet for plugins which already have those knobs

Digitalone1:
Fixes #1671 

Now the only one remaining is the LSP MB Gate which has to be implemented in place of the existing Calf version.
For the others, Dry and Gate are not included and has to be custom made by us (if we have the knowledge to do it).

### DIFF
--- a/data/schemas/com.github.wwmm.easyeffects.delay.gschema.xml
+++ b/data/schemas/com.github.wwmm.easyeffects.delay.gschema.xml
@@ -12,6 +12,22 @@
             <range min="-36" max="36" />
             <default>0</default>
         </key>
+        <key name="dry-l" type="d">
+            <range min="-100" max="20" />
+            <default>-100</default>
+        </key>
+        <key name="dry-r" type="d">
+            <range min="-100" max="20" />
+            <default>-100</default>
+        </key>
+        <key name="wet-l" type="d">
+            <range min="-100" max="20" />
+            <default>0</default>
+        </key>
+        <key name="wet-r" type="d">
+            <range min="-100" max="20" />
+            <default>0</default>
+        </key>
         <key name="time-l" type="d">
             <range min="0" max="1000" />
             <default>0</default>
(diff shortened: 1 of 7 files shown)


wwmm:
>if we have the knowledge to do it

I think that in plugins like the Bass Enchancer and the Exciter it will be just a matter of adding the 2 signals based on each configured fraction. But in plugins that may add latency like the noise remover or the convolver I am still not sure about what I am going to do. The problem with latency is that we have to pay attention we will be mixing audio buffers that are in the same position of the time axis. I think I will have to do some buffering for the input signals just to be able to add then to the processed signals.

wwmm:
At least for now my intention is doing some code refactoring in the preset classes so that they can be used with multiple filter instances. Maybe while I do that things will get clearer in my head about what to do with the dry/wet controls in the other plugins.

PR #1715, opened 2022-08-02
Implement LSP Sidechain Multiband Gate

### LINKED ISSUE #1672: Implement LSP Multiband Gate
Since we have the LSP Gate, it makes much more sense replacing the Calf MB Gate with the LSP one.

@LebedevRI do you have plans to import it?

I might suggest to replicate the MB Comp layout having Sidechain controls inside an expander widget rather than the switcher.

Digitalone1:
Fixes #1672

Tested on my system, it seems to work, but something may still be missing.

@wwmm @vchernin @LebedevRI please test.

After the merge, please update translation templates.


wwmm:
@Digitalone1 as far as I could see everything is fine.

@vchernin thanks for the implementation of these github artifacts. All I had to do to test was downloading the Arch Linux package generated by them and installing it. Super useful :-)

Digitalone1:
@wwmm please don't forget the translation template.

PR #1718, opened 2022-08-02
tests: Analyze c++ with codeql by minifying xml

vchernin:
Please wait for CI to pass on this one, it finds an error, it would be good to decide to do something with it or not.

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -117,17 +117,13 @@ jobs:
         path: ${{ steps.makepkg.outputs.pkgfile0 }}
         if-no-files-found: error
 
-  analyze:
-    name: Analyze
+  codeql-analyze:
+    name: CodeQL Analyze
     runs-on: ubuntu-20.04
     permissions:
       security-events: write
     container:
       image: archlinux
-    strategy:
-      fail-fast: false
-      matrix:
-        language: [ 'python' ] # cpp should also be here, but the codeql analysis fails due to a strange internal semmle error https://github.com/github/codeql/issues/9929
 
     steps:
     - name: Checkout repository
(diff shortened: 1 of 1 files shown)


vchernin:
If you click details next to code scanning results down below it will show you the complaint about potential overflow. It also shows up if you click on "files changed". I think it will be happy if we just define `usable_height` as a double?

wwmm:
>I think it will be happy if we just define the relevant variables as doubles?

It is possible. In this case these variables do not need to be float. I will change them to double and see what happens.

wwmm:
Doesn't it do the same when a commit is done? Is the analysis shown only for pull requests? I saw in the actions tab that it did the analysis for my last commit but I do not se if it changed the warnings or not. The output in the actions tab has a lot of things.

vchernin:
It is done for pushes to master and pull requests (per the top of CI.yml).

from the logs I spotted there seems to be 5 warnings since your commit attempting to fix it. Annoyingly I can’t see the full report which is in the security tab > code scanning alerts. It must be private for security reasons, I’m not sure if it can be made public (of course I can eventually see the results in my fork, but that’s an extra step).

vchernin:
I looked around, all I can find was this https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/managing-code-scanning-alerts-for-your-repository#viewing-the-alerts-for-a-repository which confirms only those with write access can see the proper reports in the security tab, while the pull request annotations (that we saw here) are public.

wwmm:
>I looked around, all I can find was this https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/managing-code-scanning-alerts-for-your-repository#viewing-the-alerts-for-a-repository which confirms only those with write access can see the proper reports in the security tab, while the pull request annotations (that we saw here) are public.

Oh... I forgot about the security tab :smile: . They are there indeed. Honestly I would prefer this to be public so people can help to fix thee security warnings. But if there is no other way I guess we just live with it.

vchernin:
> Honestly I would prefer this to be public so people can help to fix thee security warnings. But if there is no other way I guess we just live with it.

At least everyone who goes through a pr will be warned if they introduce some security warning. What might also help is to receive alerts, in the [repo watch button](https://docs.github.com/en/account-and-profile/managing-subscriptions-and-notifications-on-github/setting-up-notifications/configuring-notifications#configuring-your-watch-settings-for-an-individual-repository) you can enable notifications for security alerts, e.g. when you push directly and wouldn’t necessarily see the tests. Or when codeql learns about a new security warning that it didn’t find before.

wwmm:
>What might also help is to receive alerts, in the [repo watch button](https://docs.github.com/en/account-and-profile/managing-subscriptions-and-notifications-on-github/setting-up-notifications/configuring-notifications#configuring-your-watch-settings-for-an-individual-repository) you can enable notifications for security alerts, e.g. when you push directly and wouldn’t necessarily see the tests. Or when codeql learns about a new security warning that it didn’t find before.

I already have `All Activity` enabled. So I imagine I will receive security notifications too.

Issue #1698, opened 2022-07-28
Missing parameters into Equalizer and Loudness


Digitalone1:
![image](https://user-images.githubusercontent.com/25790525/181625024-1b64b8bc-2f5d-4931-a134-3ee514827bfe.png)

Should we add Balance and Pitch L/R?

Digitalone1:
Also for Loudness Compensator:

```
	Port 10:
		Type:        http://lv2plug.in/ns/lv2core#ControlPort
		             http://lv2plug.in/ns/lv2core#InputPort
		Symbol:      hclip
		Name:        Hard-clipping enable
		Minimum:     0.000000
		Maximum:     1.000000
		Default:     0.000000
		Properties:  http://lv2plug.in/ns/lv2core#toggled


	Port 11:
		Type:        http://lv2plug.in/ns/lv2core#ControlPort
		             http://lv2plug.in/ns/lv2core#InputPort
		Symbol:      hcrange
		Name:        Hard-clipping range
		Minimum:     0.000000
		Maximum:     24.000000
		Default:     6.000000
		Properties:  http://lv2plug.in/ns/ext/port-props#hasStrictBounds
```

wwmm:
I think it makes sense to add them. The balance control isn't really needed because the stereo tools plugins provides one. But on the other hand putting one more plugin in the pipeline just to do balance when the user is using only the equalizer sounds like an overkill.

Digitalone1:
Any idea where to place those balance and pitch spinbuttons?

Having them in the same layout as mode/band parameters does not seem a good idea to me because the will enlarge the already existing buttons (especially with some translations).

I was thinking of doing a layout of the label on top of the spinbutton rather than having both on the same line. But don't know if this is appreciated.

Another option could be having only buttons at the right and organize spinbuttons (horizontal) and relative labels below the band sliders box.

I'm open to suggestions.

wwmm:
>I was thinking of doing a layout of the label on top of the spinbutton rather than having both on the same line. But don't know if this is appreciated.

It is the kind of thing we would have to see how it looks. At this moment I have the feeling it may become a little confusing alternating labels and spinbuttons in the vertical.

Something else to consider is doing something similar to the one done in the autogain. But in this case we lose some vertical space.

Digitalone1:
I think I'll go with leaving all the buttons in the right side and placing the spinbuttons at the top

Digitalone1:
[...]

PR #1720, opened 2022-08-03
Equalizer: Add Balance and Pitch parameters

Digitalone1:
Related to #1698 

@wwmm I expected the pitch parameter to change much the voice, but it's quite imperceptible. The balance works as expected.

I recommend to merge this before the weblate request, otherwise there could be some conflicts.


wwmm:
>I expected the pitch parameter to change much the voice, but it's quite imperceptible. The balance works as expected.

So far I could not notice any difference in pitch. I wonder if we have misunderstood what this control is really doing or if the name in the LSP window is talking about a different kind of pitch. The balance is easily noticeable as expected.

wwmm:
Based on its docs `Pitch Left - the frequency shift for all filters of the left channel, in semitones.`. I think this is doing something completely different from what the pitch plugin does. What is good because we are actually exposing new features. But I am not sure I understand what this is achieving. Maybe the reason why I do not notice anything is because my current band values do not have a frequency where this shift causes noticeable difference.

wwmm:
Oh... I did a test now with only one band. I set the type to `high pass` and the frequency to `32 Hz`. Depending on the "pitch value" I can completely kill the signal in the corresponding channel. It is the same effect as setting a higher frequency. It is like the manual says. It is shifting the frequency of the band to higher or lower values depending on whether positive or negative values are set in the "pitch".

The pitch plugin is shifting the pitch of the voice. But in the Equalizer shifting the pitch shifts the band frequency.

PR #1732, opened 2022-08-06
Show missing plugin alerts

### LINKED ISSUE #1730: Inform the user about missing plugin
As reported in https://github.com/wwmm/easyeffects/issues/1685#issuecomment-1194723147

Since we already have UI util to show toasts in the plugin interface, this should be easily implemented.

The big work is adding an overlay to all existing effects. At the moment only few of them have it for showing toasts.

We should also format the message including the package needed for the effects. Something like:

```c++
_("%s is not installed on the system. This effect in not available.", "Linux Studio Plugins");
```

Digitalone1:
What I did:
- an adw toast overlay widget has been added to all plugins ui xml: this may not be needed, but maybe in the future we would have the need to show toasts inside the plugin interface (at moment only equalizer and rnnoise show them for errors related to loading apo/model profiles), so having them already there could be useful.
- a new tag has been added for package names and the plugins are initialized with one of them so a plugin knows its name along with the package providing its functionality: this could be useful to get or emit the package name if it's missing.
- a new `package_installed` bool flag is added to plugin base so we pass it the value of `lv2_wrapper->found_plugin` and it knows if the package is installed.
- plugins which rely on external packages have a `AdwToastOverlay* toast_overlay`: this could be removed if it's not used for showing missing plugin alerts.
- a new ui helper has been added to format the text for missing plugin and showing it inside the provided toast: as I said in the ticked, to format the translated text having the package name string I had to use `fmt::runtime()` bacause gettext() does not provide a constexpr string.

Now we have to decide what to do.
* Showing an adw toast overlay with the text saying to the user that the selected plugin is not available because the specified package is not installed (I failed doing this emitting the signal when `lv2_wrapper->found_plugin` was checked false, so I removed that signal). We may find another way getting the `package_installed` bool value from user interface.
* @wwmm As you suggested, not creating the plugin controls, but showing another box with the message that the package is not available (in this case the ui helper for missing plugin toast has to be removed, and also the `AdwToastOverlay* toast_overlay` could be removed from ui c++ files, but it can remain in the xml builder for future use).

Don't know what is better. We have to decide also taking in consideration the multiple plugin instances for the future.

I also made some fixes regarding the signals which had non references binding to references or different types leading to implicit conversions, but this is not related to the missing plugin alerts.

Edit:
Fixes #1730

### DIFF
--- a/data/ui/autogain.ui
+++ b/data/ui/autogain.ui
@@ -5,549 +5,564 @@
         <property name="margin-end">6</property>
         <property name="margin-top">6</property>
         <property name="margin-bottom">6</property>
-        <property name="spacing">12</property>
         <property name="orientation">vertical</property>
         <child>
-            <object class="GtkGrid">
-                <property name="halign">center</property>
-                <property name="row-spacing">6</property>
-                <property name="column-spacing">24</property>
-                <property name="margin-bottom">6</property>
-                <child>
-                    <object class="GtkLabel" id="label_target">
-                        <property name="label" translatable="yes">Target</property>
-                        <layout>
-                            <property name="column">0</property>
-                            <property name="row">0</property>
-                        </layout>
-                    </object>
-                </child>
-                <child>
-                    <object class="GtkSpinButton" id="target">
-                        <property name="halign">center</property>
-                        <property name="valign">center</property>
-                        <property name="width-chars">10</property>
-                        <property name="adjustment">
-                            <object class="GtkAdjustment">
-                                <property name="lower">-100</property>
-                                <property name="value">-23</property>
-                                <property name="step-increment">0.1</property>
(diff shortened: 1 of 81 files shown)


wwmm:
>a new package_installed bool flag is added to plugin base so we pass it the value of lv2_wrapper->found_plugin and it knows if the package is installed.

For RNNoise we will have to set `package_installed` to true or false using a preprocessor `#ifndef`. The decision is about whether rnnoise will be used or not is done when the code is compiled.

Digitalone1:
@wwmm It seems okay. Uninstalled LSP on my system and the status page is showing. Please give a look at it, especially at the preprocessor for rnnoise and test on your system.

Fixes #1730

PR #1736, opened 2022-08-09
Code optimization

Digitalone1:
I think this has a better fix to the codeql issue about double/float casting.

I also optimized the chart code to try to avoid implicit casting and remove some unneeded static casting. Please take a look.

Where can I look at codeql issues? Can you link the page in the Readme? Some other projects use these services to analyze the code and have the link in the Readme. Thanks.

### DIFF
--- a/include/chart.hpp
+++ b/include/chart.hpp
@@ -60,7 +60,7 @@ void set_color(Chart* self, GdkRGBA color);
 
 void set_axis_labels_color(Chart* self, GdkRGBA color);
 
-void set_line_width(Chart* self, const double& value);
+void set_line_width(Chart* self, const float& value);
 
 void set_draw_bar_border(Chart* self, const bool& v);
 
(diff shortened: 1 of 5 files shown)


wwmm:
>Where can I look at codeql issues? Can you link the page in the Readme? Some other projects use these services to analyze the code and have the link in the Readme. Thanks.

I wonder how they are putting this in the readme. Probably manually because github only gives access to the security tab to the repository owner. I did not see in in the repository settings how to make the security warnings public.

In any case as the codeql analysis for the pull requests are public here while the pull request was not accepted we will know if a PR is introducing security issues visible to codeql. We just have to pay attention to the analysis.

My own pushes can introduce security warnings. But in this cases github warns me directly anyway. At this moment we have no warning in the security tab.

Digitalone1:
Oh maybe it's just codeql like this. I'm contributing to another project in javascript which has codacy and the issue page is public.

vchernin:
Yes this seems to be a codeql specific quirk since it is primarily intended for security issues. Codacy from my experience seems more oriented around code quality/linting. 

I think it would be nice to have more linting checks but I am not sure what good tool would be to use. Codacy doesn’t even have any C or C++ tools it seems.

I tried looking into enabling more warnings with `-Wall -Wextra` but those seem to produce a ton of warnings that will need to be sorted through. Maybe this would be good for new issue.

Issue #1738, opened 2022-08-09
cppcheck config


vchernin:
Experimenting with `cppcheck` it seems this is a decent config:
```
cppcheck --enable=all --enable=information --inconclusive --error-exitcode=1 --suppress=missingIncludeSystem --suppress=unusedStructMember --suppress=unusedFunction --suppress=unknownMacro --inline-suppr --library=gtk --library=posix  ./src 2> err.txt
```
Unfortunately the upstream gtk config will need a lot of work so we can stop suppressing unusedStructMember,unusedFunction and unknownMacro, as those seem to result in lots of unhelpful warnings.

But the rest seem like potentially real complaints.

<details><summary>cppcheck</summary><br/>

```cpp
[1msrc/application.cpp:83:99: [31mperformance:[39m Function parameter 'name' should be passed by const reference. [passedByValue][0m
  self->data->connections.push_back(self->pm->new_default_sink_name.connect([=](const std::string name) {
                                                                                                  ^
[1msrc/application.cpp:91:101: [31mperformance:[39m Function parameter 'name' should be passed by const reference. [passedByValue][0m
  self->data->connections.push_back(self->pm->new_default_source_name.connect([=](const std::string name) {
                                                                                                    ^
[1msrc/application.cpp:247:14: [31mstyle:[39m Consider using std::accumulate algorithm instead of a raw loop. [useStlAlgorithm][0m
        list += name + ",";
             ^
[1msrc/application.cpp:255:14: [31mstyle:[39m Consider using std::accumulate algorithm instead of a raw loop. [useStlAlgorithm][0m
        list += name + ",";
             ^
[1msrc/application_ui.cpp:166:78: [31mperformance:[39m Function parameter 'title' should be passed by const reference. [passedByValue][0m
          ->presets_manager->preset_load_error.connect([=](const std::string title, const std::string descr) {
                                                                             ^
[1msrc/application_ui.cpp:166:103: [31mperformance:[39m Function parameter 'descr' should be passed by const reference. [passedByValue][0m
          ->presets_manager->preset_load_error.connect([=](const std::string title, const std::string descr) {
                                                                                                      ^
[1msrc/compressor.cpp:222:21: [31mstyle:[39m Unused variable: serial [unusedVariable][0m
  for (const auto& [serial, node] : pm->node_map) {
                    ^
[1msrc/compressor_ui.cpp:136:21: [31mstyle:[39m Local variable 'serial' shadows outer variable [shadowVariable][0m
[...]

PR #1739, opened 2022-08-10
lint: Fix std:source_location clang-tidy errors by ifdefing for clang

vchernin:
Related to: https://github.com/wwmm/easyeffects/issues/1738

There doesn't seem to be a way to supress clang errors for clang-tidy, so we must fix the actual build error well enough for clang-tidy
This usage of an experimental source_location will eventually fail once experimental source_location has been removed from stable clang, when that happens this change should be reverted

### DIFF
--- a/include/util.hpp
+++ b/include/util.hpp
@@ -34,6 +34,12 @@
 #include <string>
 #include <thread>
 #include <vector>
+// this usage of an experimental source_location will eventually fail once experimental source_location has been removed from stable clang
+// when that happens source_location (a c++20 feature) has probably made it to stable clang, 
+// so we should remove the ifdefs and only use std::source_location for both gcc and clang.
+#ifdef __clang__
+#include <experimental/source_location>
+#endif
 
 namespace util {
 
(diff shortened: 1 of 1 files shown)


wwmm:
I think we can make this work. Although having to also make changes to util.cpp is annoying I think we can make things a little better through a typedef that will define a new type based on `std::source_location` or `std::experimental::source_location` depending on whether clang is being used. I will see what happens.

vchernin:
> I think we can make this work. Although having to also make changes to util.cpp is annoying I think we can make things a little better through a typedef that will define a new type based on `std::source_location` or `std::experimental::source_location` depending on whether clang is being used. I will see what happens.

The problem is I found even more weird compile errors with clang. I am not at my computer with the logs currently but the import of `<set>` and I think `<range>` in some headers causes a long clang error since some functions are apparently missing from where clang expects them.

I need to confirm it actually works but I think it would be better to use `clang-tidy` directly which doesn’t seem to actually compile the code. And then we can formally warn in meson if someone tries to build with clang.

wwmm:
I tried to compile with clang now and I saw erros like
```
../src/effects_base.cpp:230:31: error: invalid operands to binary expression ('std::map<std::string, std::shared_ptr<PluginBase>>' (aka 'map<basic_string<char>, shared_ptr<PluginBase>>') and 'const _Elements<1UL>')
  for (auto& plugin : plugins | std::views::values) {
```
It seems clang does not have full support for c++20 `ranges` yet. In this case there is no point in trying to use it at this moment.

vchernin:
> It seems clang does not have full support for c++20 ranges yet. In this case there is no point in trying to use it at this moment.

Indeed, and since there isn't really a good way to hide the build errors, and it is very slow so it is a pain to find the latest errors after trying to fix them, I will abandon the idea of clang-tidy in CI for now.

Issue #1681, opened 2022-07-25
Show debug info in about page


vchernin:
It would be helpful for debugging to show the commit of the build in a debug section/page in the about window. We could also show things like distro or whether this is the flatpak build. 

We could ask in the bug report template to copy paste from this window to ensure we have accurate info to start with.

wwmm:
>It would be helpful for debugging to show the commit of the build in a debug section/page in the about window. We could also show things like distro or whether this is the flatpak build. Potentially info about plugins and so on could also be added.

I think that as long as the issue is not one of those where EasyEffects crashes before its window is shown it would help. But I am not sure about how to get some of these information in a way that works both on Flatpak and non Flatpak builds.

Maybe we should try to get some of this information in the logs first and after this considering if it will be worth the effort of having them in the about window.

vchernin:
FYI we can get some distro information with `/etc/os-release`, annoyingly only fields like `ID` are actually guaranteed to be there: https://www.freedesktop.org/software/systemd/man/os-release.html. Not to mention that it's useless in flatpak.

We can show if this is the flatpak build with a runtime check using libportal or just checking for `/.flatpak-info`, but for other builds I don't think there is a good way without editing all the distro package files and adding some `-Ddistro=`. So all we'd be able to say is `other` and `flatpak` for the distro.

I think showing information about plugins or pipewire itself would be more useful. For inspiration where I got this idea from originally is gnome builder: https://gitlab.gnome.org/GNOME/gnome-builder/-/blob/main/src/libide/gui/ide-support.c

vchernin:
What do you think of just recording all the already existing logs using the internal log helper functions (all of them meaning as if `G_MESSAGES_DEBUG=easyeffects` was passed), and copying them into the troubleshooting section of the about window? So then users can just copy paste from the app to get logs when they see an issue.

This would make it easier to get logs, since users don’t need to open the command line and then copy paste or save stdout and stderr to a file manually. Also they don’t need to restart the app, which is a problem because easyeffects needs to be fully quit if shutdown on window close is disabled, and that is easy to miss.

I’m sure it’s possible the logs could balloon, for that case there could be a hardcoded limit on kept log size. Most of the time I imagine the logs aren’t that massive?

I don’t think there should really be a performance impact since the codepath for logs is mostly executed normally anyhow (just the logs don’t get printed) I think?

[...]

PR #1750, opened 2022-08-14
Add development build profile

vchernin:
This adds a meson option `-Ddevel=true|false`. @wwmm please test building locally to make sure this makes sense. I tested in fedora using just meson and ninja, and with the flatpak build.

This adds a optional new app id `com.github.wwmm.easyeffects.Devel`. This is especially useful for Flatpak users as apps with same ID always share their user data. Now if you build the flatpak using the manifest here you will get a development build which doesn't share data with the stable build. So e.g. you can leave the stable build and its configs untouched while you experiment and maybe break things in the devel build.

This also adds a devel css class which adds a construction tape looking header bar, to make it obvious in screenshots if this is a development build. There is a corresponding devel icon as well.

Notably gsettings/gschemas internally are left always using `com.github.wwmm.easyeffects` as it is useful to not impact the settings in case someone wants to copy/edit them (but this does not impact the flatpak builds being isolated).

The main possibility for error is if the wrong id is used say for gsettings, so I specifically named them `APP_ID_BUILD` (everything but gsettings) and `tags::app::id_canonical` (gsettings). 

Another drawback is you will never see "there is nothing to do" from ninja since meson must always check the commit before it starts the build, it has no reliable way to watch git itself. But it shouldn't have a serious impact.

Also this includes the commit in the about page for https://github.com/wwmm/easyeffects/issues/1681, but that is all for now.


wwmm:
>The main possibility for error is if the wrong id is used say for gsettings, so I specifically named them APP_ID_BUILD (everything but gsettings) and tags::app::id_canonical (gsettings).

Tomorrow I will try to do some tests. But looking at the code things seems mostly fine. The renaming from `id` to `id_canonical` that feels a little weird. It isn't really clarifying things. As most of the code using this variable is gsettings code and both releases will use the same gsettings path I think it is better to keep the name as `id` and use `conf.set_quoted` to define a bool variable named like `IS_DEVEL_BUILD` or similar to select the correct string where it is necessary or in situations like `if (tags::app::id_canonical != app_id_build)`.

vchernin:
> The renaming from `id` to `id_canonical` that feels a little weird. It isn't really clarifying things. As most of the code using this variable is gsettings code and both releases will use the same gsettings path I think it is better to keep the name as `id` and use `conf.set_quoted` to define a bool variable named like `IS_DEVEL_BUILD` or similar to select the correct string where it is necessary or in situations like `if (tags::app::id_canonical != app_id_build)`.

I can revert the `id` -> `id_canonical` commit easily enough. 

I originally had a conf.set_quoted for a devel bool like you suggest but I removed it since we only need to check it in one spot, but it could be readded. I wasn't sure what the sane way to read the `IS_DEVEL_BUILD` variable would be since it would be "true" or "false" and reading from a file.

wwmm:
>I wasn't sure what the sane way to read the IS_DEVEL_BUILD variable would be since it would be a string "true" or "false", reading from config.h.

DEfiningas a string would not be good indeed. I think Meson has another function similar to `set_quoted` that is able to define an integer. Maybe it is `conf.set`. I would have to check in its docs but I am almost sure I used it before.

wwmm:
There is a way to define a boolean https://mesonbuild.com/Configuration.html
```
conf_data.set10(token, boolean_value)
# The line above is equivalent to this:
if boolean_value
  conf_data.set(token, 1)
else
  conf_data.set(token, 0)
endif
```

wwmm:
Not related to the PR but almost 60 minutes have passed and the aarch64 CI is still compiling the code :smile:

vchernin:
> There is a way to to do it without strings https://mesonbuild.com/Configuration.html

Thanks, this worked much better. 

> Not related to the PR but almost 60 minutes have passed and the aarch64 CI is still compiling the code

It is technically related to the PR since I changed some build configs to speed up some dependency builds, but that then requires a rebuild. I will disable the aarch64 builds anyhow as it is obvious it will just sometimes be very slow since it is not terribly uncommon to need to edit the Flatpak dependencies. At least with Flatpak I have never seen any differences build wise with aarch64.

PR #1756, opened 2022-08-14
Don't translate appname

vchernin:
The comment looks right with poedit on my system. I couldn't find a way to mark it as not translatable at all in the desktop file.


vchernin:
it seems some of the arch mirrors are down, not much we can do.

PR #1763, opened 2022-08-16
Some miscellaneous changes

vchernin:
Please check over the license commit, I am no authority on licenses but I think it makes sense to explictly state/repeat what the source files say in the readme.

This is a draft (CI will fail) since I set the warning level to 2, and enabled `-Werror` on devel builds. I suppressed some seemingly useless warnings, but it found two minor warnings, one I wasn't sure what the actual fix is:
https://github.com/wwmm/easyeffects/blob/7158390d6e6a2f4da55d68b7e007f1e1713ff8eb/src/convolver_menu_combine.cpp#L74 Since n and m are both unsigned ints that first comparison will always be true. Should that comparison just be removed?


Digitalone1:
> Since n and m are both unsigned ints that first comparison will always be true. Should that comparison just be removed?

It's not safe to remove the comparison. I made a fix for it. Please test it.

wwmm:
It seems that instead of suppressing the warning about unused variables it would be better to include the headers with `-isystem` instead of `-I` https://stackoverflow.com/questions/15053776/how-do-you-disable-the-unused-variable-warnings-coming-out-of-gcc-in-3rd-party-c. The question is if there is a way to tell this to Meson.

wwmm:
It seems that declaring the dependencies with `include_type: 'system'` is the way https://mesonbuild.com/Reference-manual_functions.html#arguments17

wwmm:
Although this works the amount of unused parameters is so huge that is better to suppress. The problem is that many callbacks force us to have arguments that are not necessarily used. And we can't change this. I wonder if there is a way to ignore unused arguments from functions arguments... Removing unused variables is a good thing. But as the warning is right now it is useless to look at it indeed.

wwmm:
Nevermind. `Wunused-parameter` is designed to show unused parameters in functions. So it will always print that insane amount of warnings. It is a pity. This could be useful in some situations. But with so many callbacks it will have to be suppressed.

vchernin:
force pushed and rebased to fix conflicts since some warnings changes were brought to master.

vchernin:
it looks like all the warnings are fixed, so this won't cause CI build failures.

Note when new warnings are added changed this might break the devel builds, but I am not too concerned since the compiler even tells you it is failing due to the warnings, and it is only for devel mode.

PR #1764, opened 2022-08-16
Fix Codeql uint issue

Digitalone1:
As reported [here](https://github.com/wwmm/easyeffects/pull/1763#issue-1339849820).


vchernin:
FYI this wasn’t from codeql, but from the built in GCC warnings.

vchernin:
Tested locally this adds 2 new warnings (which -Werror turns into errors):
```
../src/convolver_menu_combine.cpp: In lambda function:
../src/convolver_menu_combine.cpp:74:23: error: comparison of integer expressions of different signedness: ‘int’ and ‘std::vector<float>::size_type’ {aka ‘long unsigned int’} [-Werror=sign-compare]
   74 |     for (int m = 0; m < b.size(); m++) {
      |                     ~~^~~~~~~~~~
../src/convolver_menu_combine.cpp:75:35: error: comparison of unsigned expression in ‘>= 0’ is always true [-Werror=type-limits]
   75 |       if (const auto z = n - m; z >= 0 && z < a.size() - 1) {
      |                                 ~~^~~~
```

Digitalone1:
@vchernin I changed it. I never saw those warnings while compiling. Maybe I don't have them enabled. I never tweaked with the compiler.

vchernin:
FYI @Digitalone1 now that the changes have been merged you should see warnings while compiling, and if you set `-Ddevel=true`, i.e. with `meson build -Ddevel=true` when creating the build directory it will error on all warnings.

Digitalone1:
> FYI @Digitalone1 now that the changes have been merged you should see warnings while compiling, and if you set `-Ddevel=true`, i.e. with `meson build -Ddevel=true` when creating the build directory it will error on all compiler warnings.

I tried this, but I do not see errors, neither warnings. Do you see any?

vchernin:
> I tried this, but I do not see errors, neither warnings. Do you see any?

There are none now because all the warnings/errors were fixed in master.

Try revert f942147e33b3faa1fd94dab68662f2d3a10f4d9b and then you will see warnings or errors (depending on what whether you have passed `-Devel=true` or not).

PR #1769, opened 2022-08-18
Minor fixes

vchernin:
Made some opinionated readme improvements along with some wiki tweaks. I think it is clearer/more readable while still making sure all the info is there.

### DIFF
--- a/src/effects_box.cpp
+++ b/src/effects_box.cpp
@@ -310,7 +310,7 @@ void setup(EffectsBox* self, app::Application* application, PipelineType pipelin
 
   adw_view_stack_page_set_icon_name(self->plugins_box_page, "emblem-music-symbolic");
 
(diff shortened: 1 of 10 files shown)


src/effects_box.cpp
```diff
@@ -310,7 +310,7 @@ void setup(EffectsBox* self, app::Application* application, PipelineType pipelin
 
   adw_view_stack_page_set_icon_name(self->plugins_box_page, "emblem-music-symbolic");
 
-  // setting up the boxes we added t othe stack
+  // setting up the boxes we added t other stack
```
vchernin, line 313: Not sure what this sentence was supposed to mean, but this typo fix seems correct in itself.

wwmm: It should have been "to the stack" :smile:

PR #1783, opened 2022-08-27
flatpak: Fix autostart patch conflicts

vchernin:


vchernin:
Revisiting upstreaming this patch.. Having 2 code paths for just writing the file and another code path for libportal is still messy due to the gsettings key being needed for libportal.

However, provided you have `xdg-desktop-portal` > 1.14.0 at runtime (which most distros shipping recent easyeffects seem to have), and `libportal-gtk4` for building (which is also common), it should work even on the host. I tried testing on my system but in a container it didn't work, but I assume it should work outside a container.

@wwmm could you try testing it again by applying the patch and installing `libportal-gtk4`? Some of the patch itself is fairly over complicated, but that is mostly because I tried to make it robust/work well without knowing what I was doing. But it can be simplified now that I understand it better.

wwmm:
>@wwmm could you try testing it again by applying the patch and installing libportal-gtk4?

Later today or tomorrow morning I will do some tests to see what happens. For some reason I already have `libportal-gtk4` installed.

Do you remember what was the problem the last time? At some point there were compilation errors but I do not remember if this was the last problem or if we solved this and something else happened.

vchernin:
> Do you remember what was the problem the last time? At some point there were compilation errors but I do not remember if this was the last problem or if we solved this and something else happened.

The problem was in non-flatpak builds you can easily just write to `~/.config/autostart/com.github.wwmm.easyeffects.desktop` and use the existence of that file to track the state of autostart. But in Flatpak we can only write to the Flatpak container's path which is useless. So we need to add a gsettings key `enable-autostart` to keep track of the autostart state.

Said gsettings key would be useless for non-Flatpak builds, but there is no obvious way to conditionally include some gsettings keys for some builds. Even if it was possible, a bunch of ifdefs everywhere is ugly. 

The reason we didn't upstream the patch in the first place was because on the host `xdg-desktop-portal` was unable to securely find the app id, so it made invalid autostart files. However, since `xdg-desktop-portal` > 1.14.0 that bug should now be fixed.

wwmm:
@vchernin it is in a state that it can be tested. I will update our master branch with it. I think that the situation outside of Flatpak is still the same even with the newer portals. It creates the desktop file
```
wallace@wwmm ~/myfiles/code/go/nibraden $ cat ~/.config/autostart/.desktop 
[Desktop Entry]
Type=Application
Name=
Exec=easyeffects --gapplication-service
X-Flatpak=
```
but as you can see it is created without a name. I imagine that inside Flatpak it will work so I will update our master branch so you can do some tests.

wwmm:
>Run-time dependency libportal found: NO (tried pkgconfig and cmake)

The Flatpak build does not have libportal??

vchernin:
> I imagine that inside Flatpak it will work so I will update our master branch so you can do some tests.

Yeah, given your test it will probably work. But the fact it doesn't work for you sounds like a bug, maybe I will try reproduce it on fedora.

The flatpak build needs its libportal updated (albeit it isn't an important fix), I have a patch locally. Libportal is intended to be API/ABI stable although it not a guaranteed currently so I will configure updates for it.

PR #1791, opened 2022-09-01
Portal cleanup

vchernin:
Here are some minor fixes I spotted, the only serious one was the switches not being disconnected when the preferences window was disposed. This compile flag now seems to work identically compared to the previous patch. One potential issue is existing flatpak user's autostart gsettings key will be lost since it has been moved, I am not sure if there is a good way to handle that (the gnome documentation is down so I can't even look there :D).

Note I did change some uses of `== 1` and `== 0` to be `== true` and `== false` as I thought that looked cleaner, but I don't mind if they should be changed back.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -43,7 +43,7 @@ localedir = join_paths(prefix, get_option('localedir'))
 include_dir = include_directories('include')
 config_h_dir = include_directories('.')
 
-libportal = dependency('libportal', include_type: 'system', required: false)
+libportal = dependency('libportal-gtk4', include_type: 'system', required: false)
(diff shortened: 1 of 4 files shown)


wwmm:
>Note I did change some uses of == 1 and == 0 to be == true and == false as I thought that looked cleaner, but I don't mind if they should be changed back.

clang-tidy complained about the implicit type casting from bool to int. There won't be a problem because of this but using 0 and 1 has the advantage of not letting us forget we are dealing with C and not C++ functions. After making a new release I take a look at this.

PR #1797, opened 2022-09-04
Fix 6.3.0 changelog

Digitalone1:
Added missing changes to the changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,15 @@
 # Changelog
+
  
 ## 6.3.0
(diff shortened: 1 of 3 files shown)


Digitalone1:
@vchernin saving the NEWS.yaml with geany removed all whitespaces at the end of line.

It would be better in the future to trim the end of lines of the CHANGELOG file with an additional sed command.

I'm not really good at it, but you can take inspiration from [here](https://stackoverflow.com/a/20521915), if you want.

PR #1851, opened 2022-09-27
equalizer_ui: remove spacing between bands

ryuukk:
Otherwise you can't edit without having to scroll like it's a mobile UI, wich it already is tbh

### DIFF
--- a/src/equalizer_ui.cpp
+++ b/src/equalizer_ui.cpp
@@ -600,7 +600,9 @@ void build_channel_bands(EqualizerBox* self, const int& nbands, const bool& spli
 
     bands_box = self->bands_box_right;
   }
-
+  
+  gtk_box_set_spacing(bands_box, 0);
+  
   for (int n = 0; n < nbands; n++) {
     auto* band_box = ui::equalizer_band_box::create();
 


Digitalone1:
Hello @ryuukk. I don't think this could be accepted as it is at the moment because we usually set ui properties like spacing in the xml builder file, not the C++ source code.

Besides, in the past (PulseEffects), we used to have no spaces at all between bands, but an issue was opened since changes in gain and frequency labels lead to different space between bands, so we managed to set a space that is not affected by other components. 

Because of this, the PR has to be discussed.

Digitalone1:
GTK and Gnome complaints are useless and out of place here.

What you showed is a graphical equalizer, not a parametric one, am I right? Things are different.

We show sliders and quality also. Frequencies have units while JamesDSP does not include them. We also show two decimal digits for gain, James only one.

We could remove units and write frequencies shorter (like 999 and 1k). Anyway removing space as we are now is not a good option. Changes have to be done on other components, then we can think about reducing space. Otherwise you will see floating bars on changing label width.

wwmm:
>The UI is optimized for mobile UI/UX, that's why spacing is required everywhere

GTK has a bias for mobile screens. This does not mean that EasyEffects is. I do not have a mobile screen after. I use an ultrawide monitor.

>...thanks Gnome and GTK, they are destroying Linux desktop

Do not do this kind of thing here. Even considering that in the past I used KDE and QT for many years and I have my complaints against both toolkits I do not have much patience for this kind of discussion that goes nowhere. I can't force GTK to use less space just as I can't force QT to do some things I think GTK does better. Some things are just out of our hands.

>You can't even tweak 16 bands with a maximized windows at 1080p, you have to scroll..

And even with your workaround you will still have to scroll if you use 32 bands... So it is not like the real problem is being fixed...

>In JamesDSP, you can do that with ease:

Sure. Showing less controls and information as well as using a totally different equalizer and window layout you can distribute things in a different way. But their approach isn't really a good match to what has to be handled by EasyEffects.

>I don't think this could be accepted as it is at the moment because we usually set ui properties like spacing in the xml builder file, not the C++ source code.

Yes. This kind of thing is better in the xml.

>Besides, in the past (PulseEffects), we used to have no spaces at all between bands, but an issue was opened since changes in gain and frequency labels lead to different space between bands, so we managed to set a space that is not affected by other components.

Yes. PulseEffects had this annoying problem. But I am doing some tests now and it does not seem to be happening in EasyEffects. So much has changed along the years both on EasyEffects and GTK that I am not sure about why things seem different now. @Digitalone1 are you still able to notice changes in the space between bands when changing the frequency and gain after removing the space between bands? So far I can't.

In any case in my monitor I do not see a big advantage in removing the space. When setting the number of bands to 16 and increasing the window width just enough to show all the bands when there is no space only one band is hidden behind the scroll once the space is removed. 

And on top of this there is the following question. What is going to happen if we remove the space and the user chooses a theme that is not the default adwaita theme? At least in my monitor and with the default theme there is still some space between bands even after removing the space between bands. I am not sure if this is going to happen in all setups.

Digitalone1:
> @Digitalone1 are you still able to notice changes in the space between bands when changing the frequency and gain after removing the space between bands? So far I can't.

I can't test right know. Anyway it depends o the width-char property. I should be larger enough to avoid the floating sliders. If it larger enough, the spacing could be set to 0.

I thought one possible solution:

- [ ] A slim left column where we show units (`Hz`, `Q` and `dB`).
- [ ] Remove units from quality and frequencies.
- [ ] Show only one decimal for quality (in the band settings two can still be set).
- [ ] Format the frequency to show `ddd.d` for < 1000 and `dd.d`k for > 1000
- [ ] Set width-char as 7 which is less than the actual value I think, so we will have less space between bands avoiding the floating sliders

wwmm:
> Format the frequency to show ddd.d for < 1000 and dd.dk for > 1000

I am not sure how I feel about it... I think I prefer to have the unit there.

>Show only one decimal for quality (in the band settings two can still be set)

Maybe...

> Format the frequency to show ddd.d for < 1000 and dd.dk for > 1000

This already happens. But with the unit at the end.

Digitalone1:
> This already happens. But with the unit at the end.

With a separating space and `Hz` suffix to the unit.

Digitalone1:
Didn't have the time to look at this yet. But we have to consider that the longest string in a single band should be a negative gain less than -9.

1 symbol (minus sign) plus 2 digits plus 1 point plus 2 decimal digits plus 1 space plus 2 letters for dB unit, we have 9 chars. I'm pretty sure 10 is set as width-char since an additional char is considered to avoid expanding width on label changes.

So if 10 is  set for width-char and there's so spacing between bands, there's nothing we could do. Otherwise we have to accept floating bands when label width changes.

Digitalone1:
Besides I noticed that we should change frequency format because there's no point in showing the decimal digit on frequencies less than 1000 when none is shown for frequencies greater than 1000.

Digitalone1:
So we have a spacing of 6: https://github.com/wwmm/easyeffects/blob/master/data/ui/equalizer.ui#L213-L235

We could try to set it as 0. Will test on my system when have time.

wwmm:
>We could try to set it as 0. Will test on my system when have time.

That is what I did. At first sight it did not cause any side effects. But it did not really save a lot of space.

Digitalone1:
For strange reasons I don't know, I though we were showing units on gains. I was wrong. So the longest label is the frequency with 2 digits + 1 decimal separator + 1 decimal digit + 1 space + 3 letters for unit (kHz). 

It's 8. I set it and there's less space.   

Can't do better that this:

image

It seems GTK still adds some margin or the width-char take in account monospace width while the font used is narrow. Anyway I don't like to set a width-char lower that the max chars.

I will make the PR.

wwmm:
Thanks.

Issue #1733, opened 2022-08-06
Feature: Sort equalizer bands by frequency, and/or manually reorder?


nyanpasu64:
### EasyEffects Version

6.2.8

### What package are you using?

Flatpak (Flathub)

### Distribution

Void Linux

### Describe the bug

When adjusting a complex equalization preset, I often like to create fine-tuned bands at specific frequencies and widths, rather than purely adjusting the amplitude of preset bands. As a result, over time I often end up with bands out of frequency order, making it hard to see at a glance whether low or high frequencies are brought up or down.

### Expected Behavior

Ideally there would be a button to sort bands in frequency order (after adding bands), and perhaps even a way to move undesired bands to the end of the list (either setting the frequency to 20 KHz then sorting, or drag-and-drop bands to reorder them) so you can decrease the band count.

### Debug Log

<details><summary>Debug Log</summary><br/>

```
Paste your log here
```
</details>

wwmm:
This will require some serious refactoring in the equalizer window code. For performance reasons I already intended to actually use a listview container instead of a GtkBox. But I have no idea when I will be able to focus on this. With a listview handling the bands sorting may become possible. Doing this with a GtkBox holding the band widgets will be a pain.

nyanpasu64:
Is it possible to sort the underlying data structure holding bands (as if you exported to JSON and sorted the bands there), then rebuild the UI from the sorted bands (as if you loaded the preset from disk)?

I looked into sorting the JSON presets by hand, but it seemed like a nontrivial amount of work I'd have to write a JSON-parsing program for, rather than quickly sorting by hand in a text editor (the JSON contains band0, band1, then band10 through band19, _then_ band2 and so on, plus you duplicate bands for left and right channels). I may get around to it at some point, I hope, but I'm not super well versed on JSON libraries.

Digitalone1:
We have an APO format import feature. Could you do it in APO format? Its syntax is not too hard to write.

Maybe from the documentation is seems hard:

https://sourceforge.net/p/equalizerapo/wiki/Configuration%20reference/

But if you download an APO preset and look inside it, the format is pretty straightforward.

https://sourceforge.net/projects/peace-equalizer-apo-extension/files/Configurations%20%28presets%29/

wwmm:
>plus you duplicate bands for left and right channels).

[...]

PR #1872, opened 2022-10-04
Sort imported equalizer band filters

bhack:
https://github.com/wwmm/easyeffects/issues/1733#issuecomment-1267240702

### DIFF
--- a/src/equalizer_ui.cpp
+++ b/src/equalizer_ui.cpp
@@ -506,6 +506,8 @@ auto import_graphiceq_preset(EqualizerBox* self, const std::string& file_path) -
   if (bands.empty()) {
     return false;
   }
+  
+  std::ranges::stable_sort(bands, {}, &APO_Band::freq);
 
   // Apply GraphicEQ parameters obtained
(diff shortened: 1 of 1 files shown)


Digitalone1:
Thanks @bhack 

For our code style there's no need of the `else` statement after a `return`.

Please, write it as:

```c++
  if (filter_type.empty()) {
    return false;
  } 

  std::ranges::stable_sort(bands, {}, &APO_Band::freq);
```

Digitalone1:
And, please, can you do the same also for GraphicEq import? Thanks.

Here: https://github.com/wwmm/easyeffects/blob/master/src/equalizer_ui.cpp#L507

bhack:
Done

bhack:
Sorting it could go to impact the UI in the case of reduction to a subset N filters:

https://github.com/jaakkopasanen/AutoEq/tree/master/results/oratory1990/harman_over-ear_2018/1MORE%20Triple%20Driver%20Over%20Ear#parametric-eqs

PR #1907, opened 2022-10-23
Speex noise reduction support

### LINKED ISSUE #1792: Feature request: Add Speex Noise supression
OBS has this filter for noise supression that works really well for just removing constant background noise without much loss in quality. 
As far as I'm aware, EasyEffects only has AI noise reduction which really tanks audio the quality with lower-end setups; it would be really nice to have this alternative.

hch12907:
Closes #1792.


Digitalone1:
Nice. A document for the manual should be added also.

hch12907:
Added a manual page. Also added an option to the GUI for configuring noise suppression level.

PR #1922, opened 2022-10-28
Update NEWS.yaml

Digitalone1:
Update news for Speex.

The manual misses Speex new controls anyway.


wwmm:
A few lines later there is some information about the new plugin. If we just merge this there will be duplicated information. It is probably better to fuse the two lines.

Digitalone1:
Surprisingly I didn't notice it. Fixed.

PR #1944, opened 2022-11-06
about: Port to AdwAboutWindow

TheEvilSkeleton:
**Disclaimer: I know very little about C++!**

I only grabbed users with the most amount of contributions, but I didn't add everyone.

People to credit:
- @vchernin
- @Digitalone1

### DIFF
--- a/src/application.cpp
+++ b/src/application.cpp
@@ -399,15 +399,19 @@ void application_init(Application* self) {
 
   entries[2] = {"about",
                 [](GSimpleAction* action, GVariant* parameter, gpointer gapp) {
-                  std::array<const char*, 2> authors = {"Wellington Wallace", nullptr};
+                  std::array<const char*, 2> authors = {"Wellington Wallace <wellingtonwallace@gmail.com>", nullptr};
 
-                  gtk_show_about_dialog(
-                      gtk_application_get_active_window(GTK_APPLICATION(gapp)), "program-name", APP_NAME, "version",
-                      VERSION, "comments", _("Audio effects for PipeWire applications"), "authors", authors.data(),
-                      "logo-icon-name",
-                      IS_DEVEL_BUILD ? std::string(tags::app::id).append(".Devel").c_str() : tags::app::id,
-                      "license-type", GTK_LICENSE_GPL_3_0, "website", "https://github.com/wwmm/easyeffects",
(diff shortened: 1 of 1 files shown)


wwmm:
I think it may be a little weird to do that without asking these people first. Maybe they do not want their names in the about window. And in some cases like the one from `hosiet`, `Mikhail` and `Vincent`  it is more about maintaining distribution packages. This is something that is mostly done independently of the work in EasyEffects repository. Flatpak is the only exception. Maybe author isn't the best tag for them. At least `artists`, `translators` and `documentators` have their own field in the about window. Hopefully there is one for package maintainers.

I think it is better to wait for each person manifestation and add the credits ondemand. Specially because some of these contributors haven't actively participated for some years.

TheEvilSkeleton:
> In any case I think it is better to wait for each person manifestation and add the credits ondemand. Specially because some of these contributors haven't actively participated for some years.

Good point. I removed everyone. That being said, we can add a credit section via [`add_credit_section`](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/method.AboutWindow.add_credit_section.html).

wwmm:
I wonder if it is worth having an open issue where contributors interested in having their names in the about section could manifest their opinions. A PR page may not have the proper dynamic for this. What do you think @Digitalone1 and @vchernin? You two have contributed a lot to the project. So it is obviously fine to have your names in the about page if you want it.

vchernin:
I think a new issue for this is a good idea, and yeah I would be fine with my name in the about page.

Digitalone1:
Hello, no problem in having `Giusy Digital` in the About section. I don't like to share my real name, but that's a pseudonym on Github, so it's okay to use it for me. :)

TheEvilSkeleton:
@wwmm Should I add @Digitalone1 and @vchernin's names right now or is it better if they add themselves in separate MRs later?

Digitalone1:
Waiting for @vchernin answer, you can add `Giusy Digital`.

TheEvilSkeleton:
@vchernin said that they're okay with adding their name, but I want to know if @wwmm wants the respective contributors to add themselves in the list or if we can add your names in this MR.

I'm asking because I don't want to spam CI artifacts and waste resources for no reason.

wwmm:
I can add them after merging the pull request.

PR #1969, opened 2022-11-11
Update translation template (for Speech Processor)

hch12907:
Turns out that we forgot to include it in POTFILES. Oopsie!


wwmm:
Ia had to revert because the update to the template file created conflicts with the translation updates that come from weblate. Later I will manually add the changes to the potfile.

PR #2074, opened 2023-01-15
flatpak: do not bundle calf plugins

Bleuzen:
EasyEffects could just use the calf plugins already packaged on flathub (`org.freedesktop.LinuxAudio.Plugins.Calf`)
see https://github.com/flathub/com.github.wwmm.easyeffects/issues/169


vchernin:
Thanks for the pr, please also adjust `util/flatpak/com.github.wwmm.easyeffects.Devel.json` as done in https://github.com/flathub/com.github.wwmm.easyeffects/pull/170 as the version here is meant to be the main copy of everything. You can also delete `util/flatpak/shared-modules/linux-audio/fluidsynth2.json` here.

I think downstream in the flathub repo it would be best to drop the submodule since that approach didn't work too well, instead basically just copy the files in `util/flatpak` there and make a few small adjustments. But can wait until after this PR.

Bleuzen:
> please also adjust `util/flatpak/com.github.wwmm.easyeffects.Devel.json`

> delete `util/flatpak/shared-modules/linux-audio/fluidsynth2.json`

done :)

PR #2089, opened 2023-01-24
Fix excessive whitespace on property strings

hch12907:
Without the change, the tooltip looks like this:

image


wwmm:
Thanks! Somehow I did not notice this regression.

PR #2171, opened 2023-03-11
Sort Equalizer bands on GSettings

### LINKED ISSUE #2162: Sort equalizer bands internally
So now that I have time, I'd like to address this issue.

First let's do a recap. After making the import of GraphicEQ presets, a user complained about the bands not sorted by frequency. So we used the c++ `stable_sort` to sort them but this solution had two issues:

1. It could only sort imported presets from APO/GraphicEQ formats
2. The performance was not very good.

The first is self-explanatory: at that time if a user made their bands in the GUI and those bands were not sorted, they had to sort them manually (which is quite cumbersome) because we didn't have a method to sort them automatically.

The second point depends on how the sort algorithm works. Not 100% sure on what stable sort does, but the values of the vector should be replaced if needed and, for every replace there are 3 copies (copy in the temporary, then copy the first into the second and the second/temporary into the first). So it depends on how the vector is unsorted. Worse the sort, more the copies, bad the performance.

So @wwmm introduced a new GTK widget for the bands in the GUI. This solution solved the first point, but not the second since I suppose the list of widgets has to be sorted quite the same way as the vector built for APO/GraphicEQ import.

Besides this introduces an inconsistency (in my opinion) because, even if the bands are sorted graphically, they are not sorted internally, so if I make the sort and save the preset, I will have an unsorted preset at the next reload (replacing an existing different preset or at the next service startup).

So I think we should use a different approach sorting directly on the gsettings keys rather than the widgets or custom structures. Especially because the gsettings keys are bound to lv2 plugin values, so they reflects on the audio processor and the GUI, giving also the opportunity to save the preset retaining the sorted bands.

Thinking about this solution I came up with the following pseudo-code. Note that it's a draft (syntax should be not correct, types are wrong...) made only for presentation purpose:

```c++
void sort_bands () {
  struct EQ_Band {
    std::string type;
    std::string mode;
    std::string slope;
    double freq;
    double gain;
    double q;
    bool solo;
    bool mute;
  };

  const auto used_bands = g_settings_get_int(settings, "band_num");
  
  if (used_bands < 1 || used_bands > max_bands) {
    return;
  }
  
  std::multimap<double, struct EQ_Band> sorted_bands;
  
  for (int n = 0; n < used_bands; n++) {
    const double f = g_settings_get_double(settings_left, band_gain[n].data());
    
    sorted_bands.emplace(
      pair<double, struct EQ_Band>(
        f, 
        {
          .freq = f,
          .type = g_settings_get_enum(settings_left, band_type[n].data()),
          .mode = g_settings_get_enum(settings_left, band_mode[n].data()),
          .slope = g_settings_get_enum(settings_left, band_slope[n].data()),
          .gain = g_settings_get_double(settings_left, band_gain[n].data()),
          .q = g_settings_get_double(settings_left, band_q[n].data()),
          .solo = g_settings_get_boolean(settings_left, band_solo[n].data()),
          .mute = g_settings_get_boolean(settings_left, band_mute[n].data())
        }
      )
    );
  }
  
  for (const auto& p : sorted_bands) {
    g_settings_set_double(settings_right, band_frequency[n].data(), p.second.freq);
    g_settings_set_enum(settings_right, band_type[n].data(), p.second.type);
    g_settings_set_enum(settings_right, band_mode[n].data(), p.second.mode);
    g_settings_set_enum(settings_right, band_slope[n].data(), p.second.slope);
    g_settings_set_double(settings_right, band_gain[n].data(), p.second.gain);
    g_settings_set_double(settings_right, band_q[n].data(), p.second.q);
    g_settings_set_boolean(settings_right, band_solo[n].data(), p.second.solo);
    g_settings_set_boolean(settings_right, band_mute[n].data(), p.second.mute);
  }
}
``` 

What this wants to do:
1. Define a struct as helper to aggregate what we need (the data of a single band);
2. Retrieve the number of used bands;
3. Define a `multimap` of the above structure indexed by frequencies (multiple bands can have the same frequency, so we can't use a plain `map`);
4. Loop through used bands and, for each band, retrieve their values from gsettings, then store them into a new structure;
5. Insert that structure in the multimap specifying the frequency as index;
6. Loop through the multimap (which should be sorted by frequency) and store the values of each band into gsettings.

This in theory should have big advantages. As already said before, it reflects on lv2 plugin values and on the GUI. On the performance side, should be faster because we copy the values of the bands only two times: one for the struct construction (with `emplace` should be constructed in place), the other for the storing of the data into gsettings. The sort is made by the multimap internally, but I think it only moves the internal pointers, no other copies involved.

Then I'd like to do it in `equalizer.cpp`, so it could be executed also when the window is not shown (not really needed at the moment, but maybe useful in the future...).

But this only takes into account the situation when the bands are not splitted. For the `split band` scenario we have to make the whole process also for `settings_right`.

@wwmm what do you think? Is it doable or there's something I'm missing?

Digitalone1:
Fixes #2162

Everything explained in #2162 

Moreover I fixed some GTK deprecations (a lot still remain) and made a little refactoring to define bands number constant only once in Equalizer and MB Comp/Gate since it's not a good practice defining the same const in multiple files (now if it will be changed in the future, only one place has to be touched).


wwmm:
As far as I could see it is fine :-). I think we can delete the previous function used for sorting the widgets.

PR #2186, opened 2023-03-19
Overhaul clang tidy

vchernin:
Thanks to clang 16 now being released clang-tidy and clang itself can be used more easily. However Clang 16 has not made it to the arch repos yet, so this should stay as a draft. It is probably worth reviewing some of the changes made by clang tidy in one of the commits anyhow.

I tried to enable as many checks as possible, let clang-tidy fix everything it could, and then marked all the remaining checks as ignored in the `.clang-tidy` file. So now the issues in the `.clang-tidy` file can be fixed one by one, and CI will tell you if they pass.

There is a fairly large diff in here, one part that is a bit questionable is clang-tidy fixes unused variables in parameters by commenting out the variables.

In all the auto fixes it did, it made one change that broke the code at runtime in `pipe_manager.cpp`, along with another one that caused a compile failure, which needed to be overriden. Hopefully that should not happen much though.

Annoyingly clang-tidy is not able to run in parallel by itself, so I brought in a template python script from clang-tidy, along with second wrapper for hat too, so the way to run for all files is `util/clang-tidy.sh`. Be warned checking all files is very very slow. Meson does have a clang-tidy target but it feels a bit clunky from my testing.

It would also be nice to have another script to check only the changed lines in a diff, e.g. for checking PRs quickly. By the time clang 16 makes it to the arch repos I should be able to add that too. And maybe a fancy github pr comment action too: https://github.com/platisd/clang-tidy-pr-comments


wwmm:
>It is probably worth reviewing some of the changes made by clang tidy in one of the commits anyhow.

Definitely. Some of the suggestions to `pass by reference` are dangerous and have caused random crashes in the past. The problem is that clang-tidy does not know about the lifetime of some of these objects. For example the following suggestion is not safe in our situation
>void update(AppInfo* self, const NodeInfo& node_info) {
. The node_info is sent from the PipeWire thread to the gtk thread through a call to `g_idle_add`. In this process a function is schedule to be executed when the gtk thread is "idle". By the time this happens the object that existed in the PipeWire thread may have already been destroyed. So it is safer to pass by copy.

Another potential crash is
>void setup(FilterBox* self, const std::shared_ptr<Filter>& filter, const std::string& schema_path)
We want the `shared_ptr` reference count to be increased. This way we will be sure that every block using it will have a reference to something that is alive. If we pass by reference I think there is a risk that if the scope that created the pointer finishes other blocks in other threads or being called later through `g_idle_add` will access a null pointer.

wwmm:
>In all the auto fixes it did, it made one change that broke the code at runtime in pipe_manager.cpp, along with another one that caused a compile failure, which needed to be overriden. Hopefully that should not happen much though.

>this needs to occur after pw_init(), so putting it before pw_init() in the initializer breaks this
>// NOLINTNEXTLINE(cppcoreguidelines-prefer-member-initializer)
> thread_loop = pw_thread_loop_new("ee-pipewire-thread", nullptr);

Yes. clang-tidy always suggests to move code like this to the class member initializer list. But doing this with pipewire, pulseaudio and gstreamer init functions causes runtime crashes.

vchernin:
I ended up disabling the pass by reference check for `LinkInfo`, `NodeInfo` and shared pointers.

As I realized arch has historically been slow with new compiler releases, I just put the CI build on alpine linux for now. 

Even though this could still use a script that uses the diff to check faster, I think this is ready enough to merge. I haven't seen any weirdness doing basic tests.

wwmm:
Tomorrow I will try to take a closer look at the current changes. But after scrolling through them I noticed another one that is better to tell clang-tidy to ignore too
>self->data->connections.push_back(self->pm->device_output_route_changed.connect([=](const DeviceInfo& device) {

The structure with device object properties also goes from the pipewire thread to the gtk thread through `g_idle_add`. until we have time to properly evaluate if it is safe to pass it by reference it is better to pass a copy.

vchernin:
> But after scrolling through them I noticed another one that is better to tell clang-tidy to ignore too

I added DeviceInfo to the ignored list as well in the latest push.

vchernin:
Given how long [clang-tidy takes without a cache or running only on a diff](https://github.com/wwmm/easyeffects/actions/runs/4463597175/jobs/7838967588), I added a cache for github actions. It seems to at least somewhat work in my tests.

wwmm:
Clang-tidy ignored the `#ifdef RNNOISE_AVAILABLE` in the `rnnoise.cpp` file and moved to the class member initializer list code that should only be enabled when RNNOISE_AVAILABLE is set. I think I will look at changes in this pull request and slowly implement them manually on my computer. Most of what clang-tidy is suggesting is fine. But these subtle mistakes are a little dangerous.

vchernin:
I have removed the big diff from the PR and instead just disabled all the auto fixable checks for now, so they can be fixed one at a time like the other issues. So now this PR doesn't make many code changes (except for some for making clang build).

The big diff can still be seen here: https://github.com/wwmm/easyeffects/commit/d1a5242aa127edb5d2a36410f0584a3cc269baab

vchernin:
The cache setup has been a bit annoying to debug, but hopefully now it will work at least for C++ files. Header files are sort of hacked into this whole thing, and don't seem to work with `ctcache` which caches the clang-tidy invocations since header files are not in the compilation database like C++ files are.

PR #2259, opened 2023-04-09
build: make clang versions lower than 16 not able to build

toukoAMG:
commit https://github.com/wwmm/easyeffects/commit/f43d41a4c10496f94cd13a77d508cd4618d4b92d  makes clang able to build now, but other clang versions still can't, so this commit reproduces the original behaviour but only for clang versions lower than 16

this is also my first time working with meson so there is probably a much more efficient way to do this, but hey my change still works

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -15,6 +15,14 @@ project(
 	meson_version: '>= 0.59.0'
 )
 
+# see https://github.com/wwmm/easyeffects/pull/1739 for more info
+cxx = meson.get_compiler('cpp')
+cxx_version=cxx.version()
+if cxx.get_id() == 'clang'
+  clang_version = cxx_version.version_compare('>=16.0.0')
+  assert(clang_version == true, 'This project only supports gcc or > clang 16 compilers due to usage of c++20 features.')
+endif
+
 suppressed_warnings = [
   '-Wno-missing-field-initializers',
   '-Wno-unused-parameter'


wwmm:
>this is also my first time working with meson so there is probably a much more efficient way to do this, but hey my change still works

Thanks for the patch. I do not think there is a better way to check the compiler version.

meson.build
```diff
@@ -15,6 +15,14 @@ project(
 	meson_version: '>= 0.59.0'
 )
 
+# see https://github.com/wwmm/easyeffects/pull/1739 for more info
+cxx = meson.get_compiler('cpp')
+cxx_version=cxx.version()
+if cxx.get_id() == 'clang'
+  clang_version = cxx_version.version_compare('>=16.0.0')
+  assert(clang_version == true, 'This project only supports gcc or > clang 16 compilers due to usage of c++20 features.')
```
eli-schwartz, line 23: You're asserting `true == true` or potentially `false == true`, for the record, and both evaluate to the value of the first (non-constant) argument. You could just as easily error with:

```meson
assert(clang_version, 'This project only ........')
```

But you could also do this:
```meson
if cxx.get_id() == 'clang' and not cxx_version.version_compare('>=16.0.0')
    error('This project only ................')

wwmm: Thanks @eli-schwartz. I have updated the master branch with your suggestion.

eli-schwartz: Oops, if you are going to drop cxx_version = cxx.version() you also need to inline that. Which to be fair, it does make sense to inline it.

So, `and not cxx_version.version_compare('>=16.0.0')` -> `and not cxx.version().version_compare('>=16.0.0')`

wwmm: Of course! That is what happens when we just look at the code without actually testing it :smile:. Done!

PR #2282, opened 2023-04-20
app icons: handle ALSA playback cases

akay:
When ALSA is used (e.g. no PulseAudio or pipewire-pulse installed, or `media.cubeb.backend` in Firefox is set to `alsa`), the string `alsa_playback` is prepended to the node name, which prevents us from displaying the correct icon for the app:

2023-04-20_17h-29m-23s

This fixes it without changing the displayed node name:

image

...unless that's what we want, since it's redundant?


wwmm:
Thanks!

Issue #2310, opened 2023-04-30
Documentation/manual in github pages?


vchernin:
I think it would be nice to add the documentation to github pages.

It turns out it is easy to export the mallard doc pages by running (install `yelp-tools` for `yelp-build`):

```
mkdir help/html
yelp-build html -o help/html help/C
```

Then the html directory contains web page files with a `index.html` and can be uploaded to github pages using an action (you don't need to maintain a seperate branch).

<details>
<summary>Generic action example</summary>
<br>

```yaml
# Simple workflow for deploying static content to GitHub Pages
name: Deploy static content to Pages

on:
  # Runs on pushes targeting the default branch
  push:
    branches: ["master"]

  # Allows you to run this workflow manually from the Actions tab
  workflow_dispatch:

# Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
permissions:
  contents: read
  pages: write
  id-token: write

# Allow only one concurrent deployment, skipping runs queued between the run in-progress and latest queued.
# However, do NOT cancel in-progress runs as we want to allow these production deployments to complete.
concurrency:
  group: "pages"
  cancel-in-progress: false

jobs:
  # Single deploy job since we're just deploying
  deploy:
    environment:
      name: github-pages
      url: ${{ steps.deployment.outputs.page_url }}
    runs-on: ubuntu-latest
    steps:
      - name: Checkout
        uses: actions/checkout@v3
      - name: Setup Pages
        uses: actions/configure-pages@v3
      - name: Upload artifact
        uses: actions/upload-pages-artifact@v1
        with:
          # Upload entire repository
          path: '.'
      - name: Deploy to GitHub Pages
        id: deployment
        uses: actions/deploy-pages@v2
```

</details>

A website was discussed in #770, but this is only another place for the help pages to make them more accessible. It is easier to click a documentation link on the readme then to open the app and then click the manual button (which I feel is easy to miss if not paying attention). Also you can read it without needing to install the yelp package.

It may be smart to put the version in the help page itself (maybe in the about section?), and only run the deploy workflow on pushes to tags, so the documentation github page will always show the latest version, but not unreleased changes.

wwmm:
>Also you can read it without needing to install the yelp package.

And at least on KDE installiing yelp is pointlessas KDE does not use yelp. And as a result the documentation fails to open. I think it is a good idea to have the documentation also in a website.

PR #2313, opened 2023-05-01
Publish help pages to GitHub pages

vchernin:
Fixes https://github.com/wwmm/easyeffects/issues/2310

The pages are only published when a tag is pushed, but you can manually publish (after merging) by running the "run workflow" button here: https://github.com/wwmm/easyeffects/actions/workflows/deploy-help.yaml. Then the appstream check will be happy.

Note you will want to explicitly set GitHub Pages deployments source to GitHub Actions in repo settings, under Settings -> Pages -> Build and deployment -> Source. Before I changed this I saw in my repo it kept pointlessly trying to deploy the branch.

You will also want to change GitHub environment to allow any branch on this repo to push the help pages. In https://github.com/wwmm/easyeffects/settings/environments change the environment "github-pages" deployment branches to all branches. Otherwise when the action runs on a push to a tag it will not be allowed to publish its changes. You may need to run the workflow once manually before this setting shows up.

I also notably renamed Manual to Help which I think is more clear (and is what Gnome files/Nautilus does). 

Big picture ahead, I think some wiki pages (or at least parts of them) would be better suited in the (now more accessible) help pages. E.g. the command line page, and parts of FAQ. To me some of the information is directly relevant for usage of the app, the wiki in my mind is better suited for things more external to the app.

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -60,7 +60,7 @@ jobs:
         jq '.modules[0]."config-opts" += ["-Dwerror=true"]' util/flatpak/com.github.wwmm.easyeffects.Devel.json > "$updated_manifest"
         mv "$updated_manifest" util/flatpak/com.github.wwmm.easyeffects.Devel.json
 
-    - uses: flatpak/flatpak-github-actions/flatpak-builder@v6
+    - uses: flatpak/flatpak-github-actions/flatpak-builder@v6.1
       with:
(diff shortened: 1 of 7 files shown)


wwmm:
>Big picture ahead, I think some wiki pages (or at least parts of them) would be better suited in the (now more accessible) help pages. E.g. the command line page, and parts of FAQ. To me some of the information is directly relevant for usage of the app, the wiki in my mind is better suited for things more external to the app.

It is a good idea. Most people will probably keeping editing the wiki because it is easier than making a pull request but some of the info the info there should be in a more visible place.

PR #2443, opened 2023-07-12
Flatpak updates

vchernin:
I think it's best to just remove the wrapper script entirely, there isn't a good reason to leave it lying around when it isn't used by the flathub or the github actions build.


vchernin:
@wwmm could you take a look at this PR?

wwmm:
>I think it's best to just remove the wrapper script entirely, there isn't a good reason to leave it lying around when it isn't used by the flathub or the github actions build.

I agree.

Issue #2445, opened 2023-07-13
Community Presets implementation


Digitalone1:
After reading #1771, I wanted to test presets placed in a system config folder.

In addition to the common local config `~/.config/easyeffects`, EE looks into system folders given by `g_get_system_config_dirs` (appending input/output directories): https://github.com/wwmm/easyeffects/blob/master/src/presets_manager.cpp#L34-L38

And also into `etc/easyeffects/input` and `etc/easyeffects/output`: https://github.com/wwmm/easyeffects/blob/master/src/presets_manager.cpp#L41-L42

* The first issue I encountered is that if I copy a preset into a system folder, the new preset name is not shown into the menu. But if I shut down and relaunch the application, the new presets are listed inside the menu. It seems system folders are scanned only at the application startup. @wwmm can you confirm?

* Another issue is that the menu shows the delete button for a system preset, but this does not work because the application does not have root privileges to erase the related file.

System folders could be useful in the future if someone wants to distribute community presets in a separate package. But the issues above should be addressed.

- If the presets are installed when EE is running, they should be immediately visible, not requiring to restart the application.
- System presets should net be removable from the preset menu because they are tied to the external package, so the should disappear only if the user uninstalls the package. Besides the remove button does not work for them and this is confusing for the user. Therefore we should remove the delete button for system preset entries.
- We should test what happens for double presets (files with the same name) placed in both user and system folders. I suppose the ones in user config should take the precedence.

wwmm:
> It seems system folders are scanned only at the application startup. @wwmm can you confirm?

Yes. At this moment we listen to changes only in the user folder and not in the system https://github.com/wwmm/easyeffects/blob/cf37bb3b9c692850b93f299eaeeaff874f4f75ab/src/presets_manager.cpp#L65C13-L65C13. THE problem is that while the user config is in only one place the system preset path can have multiple locations. I am not sure if creating a folder monitor for each possible path is a good approach or even a good idea from a performance point of view.

>System presets should net be removable from the preset menu because they are tied to the external package, so the should disappear only if the user uninstalls the package. Besides the remove button does not work for them and this is confusing for the user. Therefore we should remove the delete button for system preset entries.

That is why I think that community or system presets should be tied to the `import dialog`. This way the workflow forces the user to import everything to a single directory where we can actually delete files. The import dialog could point to a system path by default. And in this system path the community presets would be located for people interested in them.

[...]

PR #2482, opened 2023-08-03
Preparing for Community Presets

Digitalone1:
Implemented some of the changes suggested in #2445 

Old system folders have been removed and multiple selection has been implemented for file import in Presets menu, Convolver and RNNoise plugins.

It misses the setting of initial folder which can be done with the following: https://docs.gtk.org/gtk4/method.FileDialog.set_initial_folder.html

The problem is that I don't know where to pick the folder (which I suppose it has been declared in meson file?).

@wwmm Take a look also if there's some memory leak in the new system for multiple files import.

Fixed also the changelog that excludes the `#` since it's a character for writing comments. It seems there's no way to escape it in the current format, unless the double qouted strings are used, which is not our case because it seems we're using multilines strings.

### DIFF
--- a/include/presets_manager.hpp
+++ b/include/presets_manager.hpp
@@ -135,8 +135,6 @@ class PresetsManager {
 
   std::filesystem::path user_presets_dir, user_input_dir, user_output_dir, autoload_input_dir, autoload_output_dir;
 
-  std::vector<std::filesystem::path> system_input_dir, system_output_dir;
-
   GSettings *settings = nullptr, *soe_settings = nullptr, *sie_settings = nullptr;
 
   GFileMonitor *user_output_monitor = nullptr, *user_input_monitor = nullptr;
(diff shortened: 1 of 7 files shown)


vchernin:
For issue links I recommend putting the full url like `https://github.com/wwmm/easyeffects/issues/1000` instead of just the number. I recall that looked best in both the markdown output (i.e. CHANGELOG.md) and the appstream metainfo output.

wwmm:
>The problem is that I don't know where to pick the folder (which I suppose it has been declared in meson file?).

It is created by MEson at https://github.com/wwmm/easyeffects/blob/320d3d57ff34dd11440f6b11398895fb0fff5243/meson.build#L80. The define is put by MEson in the header `config.h` that is generated at compilation time. It is just a matter of passing `SYSTEM_PRESETS_DIR` as argument.

Digitalone1:
@wwmm I'll try to do it.

@vchernin I'll replace them.

PR #2534, opened 2023-08-23
include <numeric> for std::iota in convolver_menu_combine

nekopsykose:
specifically with clang+libc++:  

```
../src/convolver_menu_combine.cpp:69:8: error: no member named 'iota' in namespace 'std'
  std::iota(indices.begin(), indices.end(), 0U);
  ~~~~~^
1 error generated.
```

### DIFF
--- a/include/convolver_menu_combine.hpp
+++ b/include/convolver_menu_combine.hpp
@@ -21,6 +21,7 @@
 
 #include <adwaita.h>
 #include <execution>
+#include <numeric>
 #include "convolver_ui_common.hpp"
 #include "resampler.hpp"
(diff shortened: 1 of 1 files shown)


wwmm:
Thanks. Somehow gcc does not need. Which clang version is requiring the header?

nekopsykose:
in this case it's using llvm 16. i assume this isn't related to clang though, but rather libc++, since if it was purely clang then clang+libstdc++ would have the same complaint (it doesn't), and these default-header-namespace-includes are instead a c++ stdlib thing as opposed to the compiler

Issue #2535, opened 2023-08-23
failure to build with libc++


q66:
### EasyEffects Version

7.0.7

### What package are you using?

Other (specify below)

### Distribution

Chimera Linux

### Describe the bug

The software does not build, as it currently assumes that `std::from_chars` is implemented for floating point types, which is not the case with libc++ in any current or upcoming version (unfortunately).

### Expected Behavior

The software builds

### Debug Log

https://github.com/chimera-linux/cports/actions/runs/5955398619/job/16153961535?pr=396

### Additional Information

I've written a patch to work around this: https://gist.github.com/q66/0d5a0508f75c8f985c619e8df23158e2

That said, maybe there is a better way, or at least that path perhaps should not be used unconditionally

q66:
After patching this, it also fails to run with errors like `libc++abi: terminating due to uncaught exception of type std::__1::regex_error: One of *?+{ was not preceded by a valid regular expression. `

This appears to be due to stuff like :

```
static const auto re_preamp = std::regex(R"(preamp\s*+:\s*+([+-]?+\d++(?:\.\d++)?+)\s*+db)", std::regex::icase);
```

Is `\s*+` and so supposed to be valid regex? Because I don't think it is (you can only have one quantifier)

vchernin:
I am unsure why our [alpine linux clang CI build](https://github.com/wwmm/easyeffects/blob/32599bdec8eac122a4b767c7898022c5bc1f5910/.github/workflows/CI.yaml#L208) isn't using libc++. Do you happen to know how to tell meson to use libc++ instead of libstdc++?

If we begin to support libc++ it would be good to make sure we can test it.

q66:
Regarding the regex, I notice that you use `+` everywhere after quantifiers in the regexes. This fails on libc++ but for some mysterious reason works with libstdc++. Could you explain what the intent is with these?

Alpine is not using libc++, as it's not built with a clang-based toolchain.

q66:
(that said, if we can get this to work, you will get free testing through my distribution, as it will be included in packages)

wwmm:
>I've written a patch to work around this: https://gist.github.com/q66/0d5a0508f75c8f985c619e8df23158e2
>That said, maybe there is a better way, or at least that path perhaps should not be used unconditionally

[...]

PR #2536, opened 2023-08-23
fix suffixed-+'s on regex's

nekopsykose:
these are not valid, and libc++ fails on these with:

libc++abi: terminating due to uncaught exception of type std::__1::regex_error: One of *?+{ was not preceded by a valid regular expression.

it happens to work on libstdc++, but the intention is the same without the + and works with both

---

part of #2535

### DIFF
--- a/src/equalizer_ui.cpp
+++ b/src/equalizer_ui.cpp
@@ -151,7 +151,7 @@ void on_calculate_frequencies(EqualizerBox* self, GtkButton* btn) {
 auto parse_apo_preamp(const std::string& line, double& preamp) -> bool {
   std::smatch matches;
 
-  static const auto re_preamp = std::regex(R"(preamp\s*+:\s*+([+-]?+\d++(?:\.\d++)?+)\s*+db)", std::regex::icase);
+  static const auto re_preamp = std::regex(R"(preamp\s*:\s*([+-]?\d+(?:\.\d+)?)\s*db)", std::regex::icase);
 
   std::regex_search(line, matches, re_preamp);
 
@@ -166,7 +166,7 @@ auto parse_apo_filter(const std::string& line, struct APO_Band& filter) -> std::
   std::smatch matches;
 
   static const auto re_filter =
-      std::regex(R"(filter\s*+\d*+\s*+:\s*+on\s++([a-z]++(?:\s++(?:6|12)db)?+))", std::regex::icase);
+      std::regex(R"(filter\s*\d*\s*:\s*on\s+([a-z]+(?:\s+(?:6|12)db)?))", std::regex::icase);
 
   std::regex_search(line, matches, re_filter);
 
@@ -175,7 +175,7 @@ auto parse_apo_filter(const std::string& line, struct APO_Band& filter) -> std::
   }
 
   // Possible multiple whitespaces are replaced by a single space
-  auto apo_filter = std::regex_replace(matches.str(1), std::regex(R"(\s++)"), " ");
+  auto apo_filter = std::regex_replace(matches.str(1), std::regex(R"(\s+)"), " ");
 
   // Filter string needed in uppercase for unordered_map
   std::transform(apo_filter.begin(), apo_filter.end(), apo_filter.begin(),
@@ -193,7 +193,7 @@ auto parse_apo_filter(const std::string& line, struct APO_Band& filter) -> std::
 auto parse_apo_frequency(const std::string& line, struct APO_Band& filter) -> bool {
   std::smatch matches;
 
-  static const auto re_freq = std::regex(R"(fc\s++(\d++(?:,\d++)?+(?:\.\d++)?+)\s*+hz)", std::regex::icase);
+  static const auto re_freq = std::regex(R"(fc\s+(\d+(?:,\d+)?(?:\.\d+)?)\s*hz)", std::regex::icase);
 
   std::regex_search(line, matches, re_freq);
 
@@ -209,7 +209,7 @@ auto parse_apo_frequency(const std::string& line, struct APO_Band& filter) -> bo
 auto parse_apo_gain(const std::string& line, struct APO_Band& filter) -> bool {
   std::smatch matches;
 
-  static const auto re_gain = std::regex(R"(gain\s++([+-]?+\d++(?:\.\d++)?+)\s*+db)", std::regex::icase);
+  static const auto re_gain = std::regex(R"(gain\s+([+-]?\d+(?:\.\d+)?)\s*db)", std::regex::icase);
 
   std::regex_search(line, matches, re_gain);
 
@@ -223,7 +223,7 @@ auto parse_apo_gain(const std::string& line, struct APO_Band& filter) -> bool {
 auto parse_apo_quality(const std::string& line, struct APO_Band& filter) -> bool {
   std::smatch matches;
 
-  static const auto re_quality = std::regex(R"(q\s++(\d++(?:\.\d++)?+))", std::regex::icase);
+  static const auto re_quality = std::regex(R"(q\s+(\d+(?:\.\d+)?))", std::regex::icase);
 
   std::regex_search(line, matches, re_quality);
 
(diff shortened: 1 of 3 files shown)


nekopsykose:
well, maybe this was actually intended to be something else? i can't find a reference in libstdc++ for this kind of style being recognised or meaning something special..

wwmm:
I will wait for @Digitalone1 opinion on this as she is the one that worked the most with the regex strings. At least on Arch Linux and `gcc-libs` there is no issue with them.

nekopsykose:
> At least on Arch Linux and gcc-libs there is no issue with them.

it's handled by the c++ stdlib (libc++/libstdc++/...), and yea, the gcc libstdc++ handles this fine. i guess i am curious on _how_ it actually works with libstdc++ at all myself, even if this fix is correct, because it strikes me as bizarre :D

q66:
the `+` suffix is an extension for possessive quantifiers; e.g. for input string `aaa`, regex `a+a` will match (the `a+` will match the first two `a`'s, leaving the final `a`) while `a++a` will not (because `a++` will consume all the `a` characters, leaving nothing else to match)

however, the c++ standard's ecmascript syntax does not specify possessive quantifiers; as far as i can tell, they exist in PCRE and in java, as well as in python (from version 3.11), perl, ruby, and boost, but not in standard c++ or javascript

q66:
more findings: i just tested this with libstdc++ and it does not actually implement the semantics of possessive quantifiers; testing my `a++a` pattern with the `aaa` input string returns a match in libstdc++ (it does not in python, for example)

so this might as well be a bug in libstdc++, as they don't mention the extension anywhere (and perhaps @Digitalone1 implemented it with the assumption that they match possessively, but they do not)

Digitalone1:
> I will wait for @Digitalone1 opinion on this as she is the one that worked the most with the regex strings. At least on Arch Linux and `gcc-libs` there is no issue with them.

I'm against this. Possessive quantifiers are more efficient and safer because if they are not respected, the regex fails faster. More info [here](https://www.regular-expressions.info/possessive.html):

> Possessive quantifiers are a way to prevent the regex engine from trying all permutations. This is primarily useful for performance reasons. You can also use possessive quantifiers to eliminate certain matches. 

> The main practical benefit of possessive quantifiers is to speed up your regular expression. In particular, possessive quantifiers allow your regex to fail faster.

This is because they prevent backtracking. More info [here](https://learn.microsoft.com/en-us/dotnet/standard/base-types/backtracking-in-regular-expressions).

> Backtracking occurs when a regular expression pattern contains optional [quantifiers](https://learn.microsoft.com/en-us/dotnet/standard/base-types/quantifiers-in-regular-expressions) or [alternation constructs](https://learn.microsoft.com/en-us/dotnet/standard/base-types/alternation-constructs-in-regular-expressions), and the regular expression engine returns to a previous saved state to continue its search for a match.

> Backtracking is central to the power of regular expressions; it makes it possible for expressions to be powerful and flexible, and to match very complex patterns. At the same time, this power comes at a cost. Backtracking is often the single most important factor that affects the performance of the regular expression engine.

I know that C++ reference does not mention them and that's surprising to me (I only found it [here](https://riptutorial.com/cplusplus/example/23134/quantifiers)), but it works for most, if not all, Linux distributions and it's good to have them.

A specific patch for Chimera could be made as it was made for `std::from_chars`. But this is only my opinion.

@wwmm you decide if this is worth to merge. The problem supporting other compilers is that maybe today it's only regex (and also std::from_chars), but tomorrow can be another thing that forces us to change plans. Don't know if it's worth to do this if most of Linux distributions work out of the box with libstdc++ in an easier and more efficient way.

q66:
@Digitalone1 libstdc++ literally does not implement them; the semantics are identical to ordinary, just check it

q66:
this is *not* possessive semantics: https://ideone.com/j75Rdp

Digitalone1:
> @Digitalone1 libstdc++ literally does not implement them; the semantics are identical to ordinary, just check it

Already tested with debug messages on the command line when I introduced them.

https://github.com/wwmm/easyeffects/blob/master/src/equalizer_ui.cpp#L428-L431

> this is _not_ possessive semantics: https://ideone.com/j75Rdp

`language: C++14 (gcc 8.3)`

We try to use C++20. gcc is now on [13.2.1](https://archlinux.org/packages/core/x86_64/gcc/)

q66:
```
q66@chimera:~$ c++ --version
c++ (Debian 12.2.0-14) 12.2.0
Copyright (C) 2022 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
q66@chimera:~$ c++ -std=c++20 test.cc -o test
q66@chimera:~$ ./test
aaa
```

at least with gcc 12 it *still* does not match possessively

q66:
gcc 13: https://godbolt.org/z/1WYK417Mz

q66:
additionally, i'm grepping the libstdc++ codebase and there are zero references to any such extension existing in it

q66:
made an upstream report too: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=111129

q66:
@Dgitalone1 GCC upstream says:

```
Confirmed. There is no such extension in libstdc++, this is just a bug.
````

**Comment by Digitalone1** [comment]
That's interesting. I will retest it on my system and let you know the result.

**Comment by jwakely** [comment]
> > @Digitalone1 libstdc++ literally does not implement them; the semantics are identical to ordinary, just check it
> 
> Already tested with debug messages on the command line when I introduced them.

I don't know how you tested it, but that `regex_search` has always returned `true` with all versions of GCC:
https://godbolt.org/z/q3WvTqoMq

> https://github.com/wwmm/easyeffects/blob/master/src/equalizer_ui.cpp#L428-L431
> 
> > this is _not_ possessive semantics: https://ideone.com/j75Rdp
> 
> `language: C++14 (gcc 8.3)`
> 
> We try to use C++20. gcc is now on [13.2.1](https://archlinux.org/packages/core/x86_64/gcc/)

Doesn't matter, the std::regex semantics are unchanged except for the addition of the `multiline` option.

**Comment by Digitalone1** [comment]
> I don't know how you tested it, but that `regex_search` has always returned `true` with all versions of GCC: https://godbolt.org/z/q3WvTqoMq

That's weird. See https://en.cppreference.com/w/cpp/regex/regex_search

> Returns true if a match exists, false otherwise.

**Comment by q66** [comment]
yes, because the semantics are not possessive, so it returns a match

**Comment by jwakely** [comment]
GCC shouldn't even compile the regex, because it's not valid. But we incorrectly accept the invalid regex and also say it matches.

I'm not sure what we do, maybe we just completely ignore the extra `+` quantifiers, so we treat `(a*+a++b)` as `(a*a+b)` which _does_ match `"aaab"`.

**Comment by Digitalone1** [comment]
I'm pretty sure it was returning `false` at the time I tested it. If you say so, I made a big mistake.

**Comment by Digitalone1** [comment]
I tested it another time and I confirm it returns true on the following

```c++
std::string s = "aaa";
std::smatch matches;
auto res = std::regex_search(s, matches, std::regex("(a++a)"));
```

So it was my mistake. I'm sorry. This PR is okay to me.

@q66  Also the lines here can be removed, thanks: https://github.com/wwmm/easyeffects/blob/master/src/equalizer_ui.cpp#L428-L431

jwakely:
> I'm pretty sure it was returning `false` at the time I tested it. If you say so, I made a big mistake.

It returned false with GCC 4.8 when `std::regex` was unimplemented, but `regex_search` also returned false when match "a" against "a", because no matching was implemented yet.

Since GCC 4.9.0 (the first release with working `std::regex`) it returned true.

Digitalone1:
> It returned false with GCC 4.8 when `std::regex` was unimplemented, but `regex_search` also returned false when match "a" against "a", because no matching was implemented yet.
> 
> Since GCC 4.9.0 (the first release with working `std::regex`) it returned true.

It was 2-3 years ago, could not be a version so low. I misunderstood the results of the tests and the compiler was not failing adding the ``+ suffix. That made me think possessive quantifiers were supported.

Anyway, in 2023 not having this feature in C++ is a bit awkward. PHP has it from years.

jwakely:
I think PHP uses PCRE for its regex engine. C++ supports ecmascript and posix regexes, and neither of those supports that feature.

Digitalone1:
> I think PHP uses PCRE for its regex engine. C++ supports ecmascript and posix regexes, and neither of those supports that feature.

So if I want to optimize a regex avoiding the backtracking, is there no way to do it on C++? 

At least atomic groups are supported?

jwakely:
You want to optimize a regex in C++, don't use std::regex. PCRE2 can be used in C++.

Digitalone1:
> You want to optimize a regex in C++, don't use std::regex. PCRE2 can be used in C++.

How? With an external library? It's doable, but I don't like to add an extra dependency on this project. Let's keep it simple, even if it's not optimized...

PR #2540, opened 2023-08-24
only use parallel stl with libcxx when available

q66:
When using the libcxx workarounds option, switch off parallel for_each when using either LLVM older than 17 or when the incomplete pstl flag is set (which is when not using `-fxperimental` at this point). This is the last piece needed to make it build with libcxx.

Ref https://github.com/wwmm/easyeffects/issues/2535

### DIFF
--- a/src/convolver_menu_combine.cpp
+++ b/src/convolver_menu_combine.cpp
@@ -68,7 +68,7 @@ void direct_conv(const std::vector<float>& a, const std::vector<float>& b, std::
 
   std::iota(indices.begin(), indices.end(), 0U);
 
-  std::for_each(std::execution::par_unseq, indices.begin(), indices.end(), [&](const int n) {
+  auto each = [&](const int n) {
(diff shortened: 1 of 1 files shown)


vchernin:
I think this is fairly reasonable, it's too much hassle to build an unreleased libc++ just to get pstl working.

q66:
I have checked the existing pstl support in stable llvm, and it has too many caveats to be practically deployable, and relies on the separate pstl runtime (which implements the algos either with a serial backend, or using tbb or openmp)

in 17 there is a completely new parallel stl infrastructure that doesn't come with separate dependencies (and can use plain old threads for the implementation, and does by default) but it's behind `-fexperimental`; so i think this is the best way to go for now

wwmm:
It is fine. Combine impulse response files will be a lot slower without the parallel call. But it doesn't seem there are many users out there using this feature anyway.

PR #2551, opened 2023-08-26
RNNoise improvements: Add options for VAD and mixing.

WillyPillow:
This patch adds the following options for RNNoise:

1. Gating via the VAD probability produced by RNNoise
2. Mixing the original signal and denoised signal to avoid sounding too "dry"

### DIFF
--- a/data/schemas/com.github.wwmm.easyeffects.rnnoise.gschema.xml
+++ b/data/schemas/com.github.wwmm.easyeffects.rnnoise.gschema.xml
@@ -15,5 +15,17 @@
         <key name="model-path" type="s">
             <default>""</default>
         </key>
+        <key name="vad-thres" type="d">
+            <range min="-0.1" max="1.1" />
+            <default>0.95</default>
+        </key>
+        <key name="wet-ratio" type="d">
+            <range min="-0.1" max="1.1" />
+            <default>1.0</default>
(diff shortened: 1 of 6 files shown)


Digitalone1:
Thanks @WillyPillow 

We use to expose wet/dry controls in the user interface as decibel. So can you please do like wet controls in other plugins? Using the db/linear conversion and preparing the widget to show the -inf under a certain threshold...

You should just copy was has been done on other similar controls. Should not be so hard. Thanks.

Also, I see the release from 0 to 2000, what unit does it use? 

Another thing, we don't use `std::cerr`. Maybe you want to use `util::debug`.

Digitalone1:
Oh, the `Release` uses ms unit, ok, I got it now. 

So you should prepare the SpinButton to show the `ms` unit like release on other plugins (i.e. Compressor).

And I think the same db/linear conversion should be done in `VAD Threshold`.

So, you should also prepare to show `db` unit on `VAD Threshold` and `Wet` (using only Wet and not "Wet Ratio").

wwmm:
>Another thing, we don't use std::cerr. Maybe you want to use util::debug.

As we do not put prints inside the gsettings callbacks it is better to not have any print at all there. But this is something I can remove after the merge.

Issue #2572, opened 2023-08-30
noise reduction chops up my speech


majewsky:
### EasyEffects Version

7.0.8

### What package are you using?

Arch (easyeffects)

### Distribution

Arch Linux

### Describe the bug

I have an input effects chain set up with just the Noise Reduction filter on "Standard Model". This worked like a charm up until 7.0.7. After upgrading to 7.0.8 today, my voice is now appearing chopped up, as if the noise classification is too aggressive. 

### Expected Behavior

Speech should not be filtered out.

### Additional Information

This is unrelated to recording applications: I can reproduce it by just having EasyEffects open, having just the Noise Reduction filter with default settings on, and using the bottom-right corner button that mirrors the processed input onto my headphones.

After some twiddling with the newly added parameters, it appears that turning the "Threshold" down from the default 95% makes it better. But even at 20%, there's still some chopping that the 7.0.7 version does not exhibit.

wwmm:
As the code is right now the voice detection is always applied. It is something that can be improved in the future but I think that setting the threshold to zero should be enough to have the old behavior back. In this case anything should be considered voice. Did you try to set it to zero?

majewsky:
Interesting. At zero, it does seem to work as intended. Would it make more sense to put that as a default then? I don't have insight into the general case, but in my specific case whatever the previous versions did was a rock-solid default, so while I appreciate giving more options, I'm biased towards leaving the defaults as they were before.

wwmm:
>Would it make more sense to put that as a default then?

Then some people will complain that the voice detection does not work by default. It is probably going to be necessary to add an on/off switch for the whole voice detection block.

wwmm:
>What are the settings to match the previous behavior?

Setting the release value to the maximum should help. But I will update our master branch with a toggle to fully disable the VAD.

wwmm:
I have updated our master branch. Users on Arch Linux can test it through the AUR package. IT seems fine as far as I could see in my tests. Let me know if you face any problem.

Ropid:
It seems to me there's a bug with the settings not getting applied after a restart. I've set threshold to 0%, but when I restart then the filter seems to behave like with the default 95% setting. I have to open the easyeffects window and change threshold to something else and then back to 0% to get it to apply the setting.

wwmm:
[...]

PR #2579, opened 2023-09-07
[QoL] Change NoiseReduction threshold from 95% to 0%

zastrixarundell:
Based on #2572

The 95% threshold is too strong. It chops voice inside a room with minimal noise levels while using a Scarlett Solo and a Condenser Microphone. The degree of chopping did not allow for single word sentences, such in saying: 'Hey', 'Wow', etc.

0% resets to the previous behavior before the reimplementation. This only affects the default value and users can set to 95% manually.

### DIFF
--- a/data/ui/rnnoise.ui
+++ b/data/ui/rnnoise.ui
@@ -73,7 +73,7 @@
                                                             <object class="GtkAdjustment">
                                                                 <property name="lower">0</property>
                                                                 <property name="upper">100</property>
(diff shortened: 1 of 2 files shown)


wwmm:
There is already  in the master branch a control that allows the voice detection to be disabled. So setting the threshold to zero by default does not make sense. And doing it without updating the gsettings database schemas is not going to make things work properly.

PR #2604, opened 2023-09-16
filter shema: max frequency is 24k not 20k

LebedevRI:
Having the same default in multiple places is fun :)


LebedevRI:
@wwmm thank you!

wwmm:
>Having the same default in multiple places is fun :)

That is annoying indeed. The `GtkAdjustment` bound to the spinbutton exposes the properties for the upper and lower values. But I think that gsettings does not do the same for the range we set for the keys.

PR #2614, opened 2023-09-24
DeepFilterNet effect

openglfreak:
Using the LADSPA plugin because I'd rather write a LADSPA wrapper than learn how to integrate Rust with C++.


wwmm:
Thanks for the huge contribution. But I am trying to run it and easyeffects process is instantly killed when the plugin tries to run. I am not sure why yet. I am modifying the following pkgbuild https://aur.archlinux.org/cgit/aur.git/plain/PKGBUILD?h=deepfilternet-git to install deepfilternet. Maybe I am doing something wrong...

wwmm:
As soon as `ladspa_wrapper->create_instance` is called the system kills easyeffects process. How slow is the ladspa instance creation process? Maybe it should not be done inside a realtime thread.

wwmm:
>How slow is the ladspa instance creation process? Maybe it should not be done inside a realtime thread.

It is VERY slow. The system stopped killing us when I moved the ladspa wrapper instance creation to the main thread. But even on my Ryzen 7700 the window freezes for almost 1 second during the instance creation process. Hopefully some performance optimization will be possible in the future.

The filter is quite capable when removing noise. But at least with a video at 44.1 kHz the voices become robotic. I have to see what happens at 48 kHz. Maybe the resampling is not totally fine.

I have also renamed the filter name showed to the user. Having one filter named "Noise Reduction" and another "Noise Remover" isn't very enlightening.

wwmm:
>It is VERY slow. 

I was running a binary built with the sanatizer. A normal binary will be faster. Still worth some optimization if possible because usually things do not get so slow when the sanatizer is enabled.

wwmm:
>But at least with a video at 44.1 kHz the voices become robotic. I have to see what happens at 48 kHz. Maybe the resampling is not totally fine.

The voice is not robotic at 48 kHz. So something is not working right in the resampling. I have also noticed that the filter is setting the minimum latency to `200 ms`. But somewhere in deepfilternet readme they talk about a minimum latency of `20 ms`. Any special reason for the larger value @openglfreak ?

vchernin:
I wonder how much more performant using rust to c++ bindings would be. I reailize there is already a working ladspa wrapper, but upstream already [has a crate for python bindings](https://github.com/Rikorose/DeepFilterNet/tree/59789e135cb5ed0eb86bb50e8f1be09f60859d5c/pyDF) There are crates to help with c++ like [cxx](https://crates.io/crates/cxx) which seem promising at least.

wwmm:
>I wonder how much more performant using rust to c++ bindings would be.

I have no idea. And I am not sure if this is where the slowdown is coming from. Maybe the ladspa instance creation process is inherently slow. Fortunately the non sanatized binary seems to run under normal speeds.

openglfreak:
> > But at least with a video at 44.1 kHz the voices become robotic. I have to see what happens at 48 kHz. Maybe the resampling is not totally fine.
> 
> The voice is not robotic at 48 kHz. So something is not working right in the resampling. I have also noticed that the filter is setting the minimum latency to `200 ms`. But somewhere in deepfilternet readme they talk about a minimum latency of `20 ms`. Any special reason for the larger value @openglfreak ?

The 200ms vs 20ms difference is because I'm an idiot who can't even convert ms into seconds properly. Please change it to be the right number.

About the roboticness, I had that bug before, but that was before I added the resampler for non-48khz input, and then I didn't encounter it again, but maybe I just got lucky. However since you moved the resampler initialization to the main thread, it probably doesn't get initialized in time for the first blocks of samples to be processed, you would need to re-add the `resampler_ready` logic.

openglfreak:
> > > But at least with a video at 44.1 kHz the voices become robotic. I have to see what happens at 48 kHz. Maybe the resampling is not totally fine.
> > 
> > 
> > The voice is not robotic at 48 kHz. So something is not working right in the resampling. I have also noticed that the filter is setting the minimum latency to `200 ms`. But somewhere in deepfilternet readme they talk about a minimum latency of `20 ms`. Any special reason for the larger value @openglfreak ?
> 
> The 200ms vs 20ms difference is because I'm an idiot who can't even convert ms into seconds properly. Please change it to be the right number.
> 
> About the roboticness, I had that bug before, but that was before I added the resampler for non-48khz input, and then I didn't encounter it again, but maybe I just got lucky. However since you moved the resampler initialization to the main thread, it probably doesn't get initialized in time for the first blocks of samples to be processed, you would need to re-add the `resampler_ready` logic.

Oh also `ladspa_wrapper->n_samples = n_samples;` might be wrong if rate != 48000, would need to scale by 48000 / rate, I will send a patch.

wwmm:
>The 200ms vs 20ms difference is because I'm an idiot who can't even convert ms into seconds properly. Please change it to be the right number.

>Oh also ladspa_wrapper->n_samples = n_samples; might be wrong if rate != 48000, would need to scale by 48000 / rate, I will send a patch.

Ok.

Issue #2533, opened 2023-08-21
Expose new LSP 1.2.10 features in EasyEffects


Digitalone1:
The new release has a ton of new features. See [here](https://github.com/sadko4u/lsp-plugins/releases/tag/1.2.10).

- [x] Add phase switches to Delay Compensator plugin series.
- [x] Add Allpass filter to the Filter plugin series.
- [x] Add Bandpass, Ladder-pass and Ladder-rej filters to Parametric Equalizer plugin series.
- [x] Introduce Linear Phase mode for multiband plugin series: Multiband Compressor, Multiband Expander, Multiband Gate, Multiband Limiter.
- [x] Add Band Width option (with `oct` unit).
- [x] Implement Stereo Split mode for stereo version of Compressor.
- [x] Implement Stereo Split mode for stereo version of Gate.
- [x] Implement Stereo Split mode for stereo version of Expander.
- [x] Implement Stereo Split mode for stereo version of Multiband Compressor.
- [x] Implement Stereo Split mode for stereo version of Multiband Gate.

wwmm:
I guess it is time to start the work on the `filter` plugin and move from cal to lsp. I will do it this week.

p-chan5:
Thanks @Digitalone1 and @wwmm I was also thinking of making a feature request :D

p-chan5:
> Hope I'm no missing something.

You are forgetting the new Inspect button and Inspected filter identifier (oct) introduced in version [1.2.6](https://github.com/sadko4u/lsp-plugins/releases/tag/1.2.6).

![002](https://github.com/wwmm/easyeffects/assets/123143444/deb22fd7-f98a-437c-82bc-5a4524b75ed2)

I don't know if it's possible, but it would be very useful to see the scale detuned in cents in the equalizer band options; something similar to what happens with Width.

![003](https://github.com/wwmm/easyeffects/assets/123143444/2ebe94fb-72bd-410b-b283-02e3d3ecb091)

wwmm:
>You are forgetting the new Inspect button and Inspected filter identifier (oct) introduced in version

Native window controls like this one should work out of box. Just the controls that make sense to add to EasyEffects custom window and presets that will need some work.

LebedevRI:
Looks like Gain/Envelope/Curve meters in `Compressor` plugin
are no longer working with `LSP 1.2.10` + `EasyEffects 7.0.7`.

I guess ports were renamed?
Probably some other effects are also affected.

[...]

PR #2676, opened 2023-10-18
Equalizer: prepare to expose band width option

Digitalone1:
Related to #2533


Digitalone1:
This only adds the gsettings key and the needed strings tags. The UI implementation will be made in the future.

PR #2678, opened 2023-10-19
Add EQ Band Width and Compressor Split Stereo

Digitalone1:
As the title says. But I have an issue on the sensitive properties of SC Compressor source dropdowns. They are bound to the `split-stereo` gsettings key and I can set the sensitivity of source dropdowns whenever I click on the checkbutton, but the problem is that it's not synchronized at the UI initialization. I can't understand why. Tried both callbacks and gsettings bind solutions.

@wwmm please, take a look.

### DIFF
--- a/data/schemas/com.github.wwmm.easyeffects.compressor.gschema.xml
+++ b/data/schemas/com.github.wwmm.easyeffects.compressor.gschema.xml
@@ -24,6 +24,14 @@
         <value nick="Min" value="4" />
         <value nick="Max" value="5" />
     </enum>
+    <enum id="com.github.wwmm.easyeffects.compressor.sidechain.stereosplit.source.enum">
+        <value nick="Left/Right" value="0" />
+        <value nick="Right/Left" value="1" />
+        <value nick="Mid/Side" value="2" />
+        <value nick="Side/Mid" value="3" />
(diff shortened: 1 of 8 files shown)


wwmm:
>but the problem is that it's not synchronized at the UI initialization.

I think I did not understand what is not working. When I put the compressor in the pipeline with `split-stereo` initially off only the dropdown at the side of `Sidechain` is enabled. And if I put the compressor when `split-stereo` was previously set to on only the dropdown at the bottom is enabled. Isn't this what is supposed to happen? IT seems to be working as intended.

wwmm:
Wait... Maybe it is working for me because of the fix I made a few minutes ago
```
g_settings_bind(self->settings, "stereo-split", self->sidechain_source, "sensitive",
                  static_cast<GSettingsBindFlags>(G_SETTINGS_BIND_DEFAULT | G_SETTINGS_BIND_INVERT_BOOLEAN));
```
But at least on my computer the way ti was before caused a compilation problem.

Digitalone1:
This is weird. I got no compile errors.

wwmm:
>This is weird. I got no compile errors.

The github actions failed to compile too https://github.com/wwmm/easyeffects/actions/runs/6576113496/job/17864774570

Issue #2738, opened 2023-11-10
Add "quit" to main menu


jtl5770:
### EasyEffects Version

7.1.3

### What package are you using?

Flatpak (Flathub)

### Distribution

Fedora 39

### Describe the bug

With Fedora 39, the taskbar button for running applications is gone. So there is no way anymore to easily quit easyeffects (if you have set it to be started automatically and not quit when closed).

### Expected Behavior

Usually, an app can be can be expected to have a dedicated "quit" menu entry in the hamburger menu. My proposal is to add just that.

### Debug Log

<details><summary>Debug Log</summary><br/>

```
Paste your log here
```
</details>

violetmage:
Please see #2739

wwmm:
Implemented in our master branch.

Issue #2697, opened 2023-10-23
Crashes randomly or when firefox is restarted after a video is played.


aspizu:
1. Open firefox
2. Play video
3. Close firefox
4. Open firefox

EE will crash.

```
fish: Job 1, 'easyeffects --gapplication-serv…' terminated by signal SIGKILL (Forced quit)
```

wwmm:
Try to run it in debug mode so we can have some idea about what is being done when the crash happens https://github.com/wwmm/easyeffects/wiki/Reporting-Bugs#basic-logs. Remember to kill EasyEffects first. I use Firefox everyday to watch videos and no crashes happens on my side. So more information will be needed to understand what is going on.

aspizu:
it stopped crashing

aspizu:
```
(easyeffects:63392): easyeffects-DEBUG: 19:19:33.542:   pipe_manager.cpp:213    Stream/Output/Audio 121 Firefox has been removed
(easyeffects:63392): easyeffects-DEBUG: 19:19:43.476:   stream_output_effects.cpp:159   No app linked to our device wants to play. Unlinking our filters.
(easyeffects:63392): easyeffects-DEBUG: 19:19:43.476:   stream_output_effects.cpp:317   disconnecting the equalizer filter from PipeWire
(easyeffects:63392): easyeffects-DEBUG: 19:19:43.478:   pipe_manager.cpp:213     158 ee_soe_equalizer has been removed
(easyeffects:63392): easyeffects-DEBUG: 19:19:54.961:   pipe_manager.cpp:1167   Stream/Output/Audio 167 Firefox with serial 5412 has been added
(easyeffects:63392): easyeffects-DEBUG: 19:19:54.963:   pipe_manager.cpp:900    new metadata property: 167, target.node, Spa:Id, 141
(easyeffects:63392): easyeffects-DEBUG: 19:19:54.963:   pipe_manager.cpp:900    new metadata property: 167, target.object, Spa:Id, 5281
(easyeffects:63392): easyeffects-DEBUG: 19:19:54.970:   pipe_manager.cpp:1207   Firefox port 150 is connected to easyeffects_sink port 124
(easyeffects:63392): easyeffects-DEBUG: 19:19:54.970:   pipe_manager.cpp:1207   Firefox port 143 is connected to easyeffects_sink port 71
[...]

PR #2739, opened 2023-11-10
Add "Quit" menu option and "--version" command line option

violetmage:
As per recent requests in #2738 and #2697 , add a `--version` command line option, and add the "quit" application action to the primary_menu hamburger menu in the GUI.


wwmm:
Thanks! :-)

PR #2765, opened 2023-11-19
Add g_object_ref and g_object_unref around util::idle add sections

violetmage:
Intended to fix use-after-free discovered in #2752

It will be a lot of work to add the correct g_object_ref and g_object_unref statements, as well as refactor the logic in each util::idle_add, so that g_object_unref always happens; but hopefully this will eliminate annoying edge cases.


src/convolver_ui.cpp
```diff
@@ -303,18 +303,18 @@ void get_irs_spectrum(ConvolverBox* self, const int& rate) {
...
-      return;
-    }
-
-    if (gtk_toggle_button_get_active(self->show_fft) != 0) {
+    
+    if (  
+      ( self != nullptr ) &&
+      ( ui::chart::get_is_visible(self->chart) ) &&
+      ( gtk_toggle_button_get_active(self->show_fft) != 0 )
+    ) {
```
wwmm, line 313: This check will crash if `self == nullptr` because the check `ui::chart::get_is_visible(self->chart)` will be executed anyway and `self->chart` will point to null. Same issue on the other check.

wwmm: Something like this will have to be done
```
if (self == nullptr) {
     g_object_unref(self);
      return;
  }

if (!ui::chart::get_is_visible(self->chart)) {
  g_object_unref(self);
  return;
}
```
and in the end the `g_object_unref(self);` you have already added. This one is executed in case the other checks fail and the callback code goes to the end.

violetmage: I thought conditions with an && operator are checked in order until one returns not true; so that in this case, if self == nullptr, evaluation would stop there, skip the rest of the checks, and only execute g_object_unref

violetmage:
I'm starting to wonder if this should be implemented using weak references (GObject weakref) instead of strong references...

Will do more research while I'm on the bus today.

src/convolver_ui.cpp
wwmm: >I thought conditions with an && operator are checked in order until one returns not true;

I am almost 100% sure I tried this in the past and the code crashed. Or maybe at the time it was an `||` and not a `&&`. In any case it is probably better to play safe and keep them separated just to avoid being surprised by a different compiler behavior.

wwmm:
>I'm starting to wonder if this should be implemented using weak references (GObject weakref) instead of strong references...

It is probably safer to call `g_object_unref`. It is containers usually do to their children.

violetmage:
My reasoning is that, we don't actually need to do anything to a widget that's been freed, and weak references would be a thread safe way to know if the object has been freed.

wwmm:
>My reasoning is that, we don't actually need to do anything to a widget that's been freed

I think that this will probably require more work and will be harder to handle than just trying to keep the object alive while something still tries to use it.

>and weak references would be a thread safe way to know if the object has been freed.

Unless I misunderstood something the docs say they are not thread safe https://docs.gtk.org/gobject/method.Object.weak_ref.html. What makes sense considering that GTK as a whole is not thread safe.

The docs talks about a `GWeakRef` to try to make things thread safe but I imagine some kind of synchronization will have to happen. What probably will make things a lot more complicated than just keeping the object alive.

src/convolver_ui.cpp
violetmage: [this so andwer](https://stackoverflow.com/questions/7925479/if-argument-evaluation-order) says that the left to right semantics of && is standardized

wwmm: Ok. I did a quick test and at least on gcc it seems to be the case.

violetmage:
So do you think that I should proceed using g_object_ref and g_object_unref like this, to fix this bug?

wwmm:
So do you think that I should proceed using g_object_ref and g_object_unref like this, to fix this bug?

Yes. I think it will probably be a simpler solution. And without the risks involved in forgetting after some time where weak references differ compared to the other.

violetmage:
@wwmm do you want to remove #2533 from milestones, and create a new milestone issue for this to track progress on each plugin that needs to be refactored with g_object_ref() and g_object_unref() ?

I think it would be easier for anyone who wants to help to contribute to the refactoring that way.

wwmm:
We can close this.

PR #2766, opened 2023-11-19
Add ability to disable inactivity timeout

violetmage:
For users who do not care about saving CPU cycles, and would rather their effects chain be always ready for audio, allow the inactivity timeout to be disabled by setting the "Inactivity Timeout" setting to 0

### DIFF
--- a/data/schemas/com.github.wwmm.easyeffects.gschema.xml
+++ b/data/schemas/com.github.wwmm.easyeffects.gschema.xml
@@ -47,7 +47,7 @@
             <range min="1" max="3600" />
             <default>10</default>
         </key>
-        <key name="inactivity-timer-enabled" type="b">
+        <key name="inactivity-timer-enable" type="b">
             <default>true</default>
         </key>
         <key name="meters-update-interval" type="i">
--- a/data/schemas/com.github.wwmm.easyeffects.gschema.xml
+++ b/data/schemas/com.github.wwmm.easyeffects.gschema.xml
@@ -44,7 +44,7 @@
             <default>true</default>
         </key>
         <key name="inactivity-timeout" type="i">
-            <range min="1" max="3600" />
+            <range min="0" max="3600" />
             <default>10</default>
         </key>
         <key name="meters-update-interval" type="i">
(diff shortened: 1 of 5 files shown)


violetmage:
Currently testing https://github.com/wwmm/easyeffects/pull/2766/commits/7b3f73533870b57d48705c97ce3a9aefbbfbfc47, seeing a log of repeated debug messages and not totally sure why. I am observing that there seems to be a bunch of messages that are emitted all together when playback stops, then after a short pause, another group of repeated messages. Wondering if anyone knows what's going on here.

wwmm:
I think that is counter intuitive that setting the value to `zero` disables inactivity timeout. This parameter controls how fast EE unlins its pipeline. I would expect zero to make it instantly unlink instead of waiting some time to unlink.

It is probably better to add a new switch that disables the functionality or a more complicated approach where we use the same trick done in some of the LSP plugins control to show `Inf` when the value is set to maximum. But I will think it will be more clear to users to add a new switch. It does not have to be in a new line in the preferences window. We can add it to the same line the spinbutton already is.

violetmage:
When I get back to my computer, I will definitely add a switch as you've said.

Also, do you have any idea why the debug message is repeated so many times?

wwmm:
>Also, do you have any idea why the debug message is repeated so many times?

Because PipeWire emits the signal attached to `on_link_changed` frequently for many different reasons.  I've ignored some of the cases we are not interested to through those returns that are there. But even so there are still many we have to evaluate to decide whether or not to change the pipeline links. 

I think the only reason why we do not see many prints now is because we only do them when the pipeline is linked or unlinked and siltently ignore the other cases where it has to preserve its current state. But the check you added will be called when `apps_want_to_play() == false`  and `inactivity_timeout == 0` whenever PipeWire calls that callback.

wwmm (commented):
The new gtkswitch code seems fine. I think we can move on from the draft status.

violetmage:
Ended up moving the switch out from the same item / row contianer thing, as I think the `activatable-widget` property needs to be there, and I'm not sure if you can put that together with another activatable widget.

Also need to investigate how difficult it would be to make the spinbutton become sensitive / un-sensitive in accordance with the switch

wwmm:
>as I think the activatable-widget property needs to be there, and I'm not sure if you can put that together with another activatable widget.

I think it can.

>Also need to investigate how difficult it would be to make the spinbutton become sensitive / un-sensitive in accordance with the switch

This can probably be done just with some gtk xml code without the need to add more C/C++ code. I will do this tomorrow after merging it.

Issue #2732, opened 2023-11-08
Output stops with any effect added


fshubthunfischbaum:
### EasyEffects Version

7.1.1

### What package are you using?

Flatpak (Flathub)

### Distribution

Debian 12

### Describe the bug

Soundoutput stops completely after adding any effect to the players. 

### Expected Behavior

Soundoutput should continue to play with the applied effects.

### Debug Log

Commit: f4d4132

<details>
<summary> Log </summary>

```
(process:2): easyeffects-DEBUG: 00:12:53.641: 	easyeffects.cpp:35	easyeffects version: 7.1.1
(process:2): easyeffects-DEBUG: 00:12:53.641: 	easyeffects.cpp:45	locale directory: /app/share/locale
(easyeffects:2): easyeffects-DEBUG: 00:12:53.647: 	presets_manager.cpp:162	user presets directory already exists: /home/uwe/.var/app/com.github.wwmm.easyeffects/config/easyeffects/
(easyeffects:2): easyeffects-DEBUG: 00:12:53.647: 	presets_manager.cpp:162	user presets directory already exists: /home/uwe/.var/app/com.github.wwmm.easyeffects/config/easyeffects/input
(easyeffects:2): easyeffects-DEBUG: 00:12:53.647: 	presets_manager.cpp:162	user presets directory already exists: /home/uwe/.var/app/com.github.wwmm.easyeffects/config/easyeffects/output
(easyeffects:2): easyeffects-DEBUG: 00:12:53.647: 	presets_manager.cpp:162	user presets directory already exists: /home/uwe/.var/app/com.github.wwmm.easyeffects/config/easyeffects/autoload/input
(easyeffects:2): easyeffects-DEBUG: 00:12:53.647: 	presets_manager.cpp:162	user presets directory already exists: /home/uwe/.var/app/com.github.wwmm.easyeffects/config/easyeffects/autoload/output
(easyeffects:2): easyeffects-DEBUG: 00:12:53.762: 	pipe_manager.cpp:1429	compiled with PipeWire: 0.3.79
(easyeffects:2): easyeffects-DEBUG: 00:12:53.762: 	pipe_manager.cpp:1430	linked to PipeWire: 0.3.81
(easyeffects:2): easyeffects-DEBUG: 00:12:53.765: 	pipe_manager.cpp:1400	core version: 0.3.65
[...]

PR #2779, opened 2023-11-28
Fix compatabilty with old pipewire servers

violetmage:
Many recent reports of easyeffects breaking when used with an older pipewire server.

[This comment](https://github.com/wwmm/easyeffects/issues/2732#issuecomment-1817535159) seems to suggest that reverting the runtime may help.

### DIFF
--- a/util/flatpak/com.github.wwmm.easyeffects.Devel.json
+++ b/util/flatpak/com.github.wwmm.easyeffects.Devel.json
@@ -1,7 +1,7 @@
 {
     "id": "com.github.wwmm.easyeffects.Devel",
     "runtime": "org.gnome.Platform",
-    "runtime-version": "45",
+    "runtime-version": "44",
     "sdk": "org.gnome.Sdk",
     "command": "easyeffects",
     "finish-args": [
--- a/util/flatpak/com.github.wwmm.easyeffects.Devel.json
+++ b/util/flatpak/com.github.wwmm.easyeffects.Devel.json
@@ -1,7 +1,7 @@
 {
     "id": "com.github.wwmm.easyeffects.Devel",
     "runtime": "org.gnome.Platform",
-    "runtime-version": "44",
+    "runtime-version": "45",
     "sdk": "org.gnome.Sdk",
     "command": "easyeffects",
     "finish-args": [
--- a/util/flatpak/easyeffects-modules.json
+++ b/util/flatpak/easyeffects-modules.json
@@ -2,6 +2,27 @@
     "name": "easyeffects-modules",
     "modules": [
         {
+            "name": "pipewire",
+            "buildsystem": "meson",
+            "config-opts": [
+                "-Dgstreamer=disabled",
+                "-Dman=disabled",
+                "-Dsystemd=disabled",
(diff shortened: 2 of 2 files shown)


vchernin:
I would suggest trying to build with an older PipeWire version instead. It is not ideal to hold back the runtime version when likely only pipewire needs to be reverted.

violetmage:
Logically, I agree that bundling pipewire makes the most sense, but it seems like a fair bit of work. Additionally, I have to ask what the benefit is of moving to gnome 45, as afaict the gnome 44 runtime is not deprecated yet. I don't think easyeffects really gains anything yet moving from 44 to 45 right away, as it's not a super intense application or anything.

Also, once all the major distrobutions get to pipewire 0.3.7x and up, this should all just go away, and we can build with gnome 45.

vchernin:
>  I don't think easyeffects really gains anything yet moving from 44 to 45 right away, as it's not a super intense application or anything.

We get a newer version of gtk, but more noticeably newer compiler versions. Easyeffects tends to use fairly modern C++ features, which are simply not available in older compilers.

You should be able to reuse (i.e. copy paste) the removed module from this commit https://github.com/wwmm/easyeffects/commit/5dd8884afcf99bc5bf5400050252f276a13d5f0d and it back to `easyeffects-modules.json`. The main difference is the tag should be 0.3.69 (and the commit [from here](https://gitlab.freedesktop.org/pipewire/pipewire/-/commit/cd8be0ba3b27542253f7744b699c2ede159e2d7c)). 0.3.69 is what [freedesktop sdk 22.08](https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/blob/freedesktop-sdk-22.08.19/elements/components/pipewire-base.bst?ref_type=tags) uses (which gnome 44 uses), so it should restore the gnome 44 runtime behaviour.

violetmage:
Thanks for the tip. I'll start on that right now.

violetmage:
Tested and working on RHEL 9 w/ pipewire 0.3.67, it works like it should on gnome 45 with the bundled pipewire when it didn't work before.

vchernin (approved):

wwmm:
I wonder how a downgrade to the PipeWire version used to build the flatpak package is fixing the bug people with older pipewire in there systems are having. If it were a downgrade in EasyEffects I could understand. It feels like the kind of thing that should be reported PipeWire's developers.

vchernin:
> I wonder how a downgrade to the PipeWire version used to build the flatpak package is fixing the bug people with older pipewire in there systems are having. If it were a downgrade in EasyEffects I could understand. It feels like the kind of thing that should be reported PipeWire's developers. 

If you look at the last commit for [PipeWire in freedesktop-sdk 22.08](https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/commit/55513a130feed8e5ed78e7ef832297158ecd0bb4), you can see PipeWire was frozen due to newer versions having an ABI break. I wouldn't be surprised if this hasn't happened before given in the past building the flatpak with an older pipewire version did improve compatibility. Hopefully now that pipewire reached 1.0.0 things will break less.

Issue #1808, opened 2022-09-07
Feature request: Add support for GraphicEQ importing


Cpt-Dingus:
Hey there, APO setting importing already exists but it sadly does not include importing GraphicEQ settings. Would it be possible to add it?

Sample: GraphicEQ: 5.53608 0.8; 13.2 8.2; 26.7 -6.2; 62.7 -7; 247.2 8.2; 1326.8 -5.8; 3219.8 5.4

Digitalone1:
Hello @Cpt-Dingus. Can you link its format documentation?

Cpt-Dingus:
[Here you go](https://sourceforge.net/p/equalizerapo/wiki/Configuration%20reference/#graphiceq-since-version-10)

Digitalone1:
It's doable with regular expressions: https://regex101.com/r/EQY8BZ/1

But isn't specified which filter is applied?

Cpt-Dingus:
The filter mentioned is just a basic EQ, not aware of any other possible filters to be applied (talking about graphiceq that is)

Digitalone1:
So it's just PK, MODAL or PEQ in APO format? A bell filter? 

https://github.com/wwmm/easyeffects/blob/master/src/equalizer_ui.cpp#L37

It's missing the quality also (which I suppose we will set to our default value).

Cpt-Dingus:
Not sure about the exact type, the default type & quality should do the job, no?

Cpt-Dingus:
Dug around a bit, I found quite a few references to the filter type being PEQ.

Digitalone1:
> Not sure about the exact type, the default type & quality should do the job, no?

I suppose they do.

> Dug around a bit, I found quite a few references to the filter type being PEQ.

Okay, it's bell filter in LSP.

wwmm:
>Okay, it's bell filter in LSP.

Yes. It will have to be done with bell. From a technical point of view the correct thing to be done would be to use a graphical equalizer with the same number of bands and algorithm instead of using a parametric equalizer. Graphic equalizers are not necessarily a parametric equalizer with bell filters at each center frequency. But besides the fact we do not have a graphic equalizer in EasyEffects for each number of bands and implementation algorithm something slightly different may be happening under the hoods when a graphical equalizer is written. It is not viable for us at this moment to deal with graphical equalizers. For each number of bands we would need a different one.

wwmm:
>It's missing the quality also (which I suppose we will set to our default value).

[...]

PR #2820, opened 2023-12-21
Export apo preset

iciclejj:
Adding support for exporting equalizer presets to EqualizerAPO preset files, to complement the already in-place import functionality. For now does not support split channels (added popup when attempting).

I have an export_geq_preset ready as well that simply treats the GEQ filters as Bell filters, similar to how import_geq_preset is currently implemented, but not sure if I should add it, considering they aren't actually implemented as Bell/Peak filters in APO (as discussed in #1808 [here](https://github.com/wwmm/easyeffects/issues/1808#issuecomment-1732255085)).

I personally feel like putting the import/export buttons in their current location becomes a bit cluttered as more options are added, but I don't want to start messing with the ui layout, so just adding them logically beneath the current import buttons for now.

### DIFF
--- a/data/ui/equalizer.ui
+++ b/data/ui/equalizer.ui
@@ -357,6 +357,32 @@
                                                         </child>
                                                     </object>
                                                 </child>
+
+                                                <child>
+                                                    <object class="GtkLabel">
+                                                        <property name="label" translatable="yes">Export Preset</property>
+                                                        <property name="valign">center</property>
+                                                    </object>
+                                                </child>
+
+                                                <child>
+                                                    <object class="GtkBox">
+                                                        <property name="orientation">vertical</property>
+                                                        <property name="valign">center</property>
+
+                                                        <style>
+                                                            <class name="linked" />
+                                                        </style>
+
+                                                        <child>
+                                                            <object class="GtkButton" id="export_apo">
+                                                                <property name="label">APO</property>
+                                                                <property name="valign">center</property>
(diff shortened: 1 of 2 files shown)


wwmm:
>I have an export_geq_preset ready as well that simply treats the GEQ filters as Bell filters, similar to how import_geq_preset is currently implemented, but not sure if I should add it, considering they aren't actually implemented as Bell/Peak filters in APO (as discussed in https://github.com/wwmm/easyeffects/issues/1808 https://github.com/wwmm/easyeffects/issues/1808#issuecomment-1732255085).

Considering how different the graphical equalizer implementation is when compared to EasyEffects parametric eq approach it is probably not going to be of much help to people using APO.

wwmm:
Thanks for the contribution! :-)

iciclejj:
> Thanks for the contribution! :-)

Happy to contribute, thanks for the merge :)

Quick question: Is there a standard approach for error handling in our codebase? I noticed we're simply passing `nullptr` to the `GError **` parameters in some gtk functions in `equalizer_ui.cpp`, so I just followed that pattern.

wwmm:
>Is there a standard approach for error handling in our codebase?

Nothing besides using our own `util::warning` or `util::debug` when there is an advantage in doing so. In theory we should also handle the error argument in some gtk calls but in gtk's own source code you do not see this being done often. So far I did not see the need to handle most of them. It would hurt doing so but it would be tedious to add all this code.

Issue #1211, opened 2021-10-15
Don't change volume with scrolling


Elendil211:
**EasyEffects Version**
6.1.3

**What package are you using?**
Arch Official (easyeffects)

**Distribution**
Arch Linux

**Describe the bug**
When scrolling through the list of different players/recorders (by using the mouse wheel), and you end with your mouse cursor on the volume bar of one player/recorder, the next mouse wheel movement does not result in scrolling the list, but instead in the volume being changed.

**Expected Behavior**
Scrolling should not affect the volume. Scrolling should just scroll through the list.

**Solution**
Make the volume bar insensible to scrolling.

wwmm:
>Perhaps there is a way to disable mouse wheel scrolling for those sliders

Not that I remember. But even if there is a way to do that most of the time I change sliders positions through the mouse scroll. So I do not think that disabling it is the best solution.

saivert:
You would have a variable that is set to 1 when a scrolling event starts on the container widget and will be reset to 0 when any mouse movement occurs. This would avoid the issue where you scroll, then stop for a bit and continue scrolling and now it adjusts volume on something instead of continuing scrolling.

jannuary:
```
noëlle: is there away for a gtkscale to ignore the scrollwheel?
Alexander Mikhaylenko: noëlle: add your own event controller (probably with a capture phase) and stop events there```

wwmm:
>add your own event controller (probably with a capture phase) and stop events there```

I tried to add an event controller but the scale still got scroll events anyway. Maybe the controller has to be added in a very specific and non obvious way.

wwmm:
I will try again to add a controller. Maybe I did not test all the possible flags the last time I tried.

wwmm:
Adding a controller created as `gtk_event_controller_scroll_new(GTK_EVENT_CONTROLLER_SCROLL_NONE);`. Does nothing. The scale still gets scroll events. Same thing with `GTK_EVENT_CONTROLLER_SCROLL_HORIZONTAL`. Last time I looked at the `GtkScale` sources it seemed to have the scroll events hardcoded. Adding an event controller only seems to allow additional event handling. It does not override the ones that are hardcoded.

wwmm:
>probably with a capture phase

I wonder what he meant by `capture phase`.

wwmm:
Oh! Maybe I have to actually set a callback ignoring the events from the controller. I will try that. I was just trying to set the controller.

[...]

PR #2868, opened 2024-01-26
Change the spin button to a slider in the app info ui

toppa102:
As i understand it was a slider before but for some reason was changed to a spin button? Anyway in my opinion a slider is just faster and easier to use. 

Here's what this PR proposes:
20240126_16h04m03s_grim

### DIFF
--- a/data/ui/app_info.ui
+++ b/data/ui/app_info.ui
@@ -252,12 +252,10 @@
                                 </child>
 
                                 <child>
-                                    <object class="GtkSpinButton" id="volume">
+                                    <object class="GtkScale" id="volume">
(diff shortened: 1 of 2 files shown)


wwmm:
>As i understand it was a slider before but for some reason was changed to a spin button? 

Yes there was a reason to remove the slider #1211. It forces the mouse scroll event to go through it. What in turns does not allow the scroll to make the list go up/down. At the time there was no way to make the gtkscale to ignore the mouse scroll event. It is hardcoded in its sources. And as far as I know the situation has not changed.

Honestly I do not think there is really the need to bring the gtkscale back and the old problem together with it.

PR #2930, opened 2024-02-24
Equalizer APO: fix regex

Digitalone1:
The regex for on/off state only needs to check the initial part of the line. No need of the entire line or positive lookahead.

It uses the icase flag, so we need to force the string in uppercase, otherwise `Off` != `OFF`.

No need to remove duplicate whitespaces like in "parse_apo_filters".

@wwmm @iciclejj please, test it. No issues on my side.

### DIFF
--- a/include/app_info.hpp
+++ b/include/app_info.hpp
@@ -1,5 +1,5 @@
 /*
- *  Copyright © 2017-2023 Wellington Wallace
+ *  Copyright © 2017-2024 Wellington Wallace
  *
  *  This file is part of Easy Effects.
  *
--- a/include/application.hpp
+++ b/include/application.hpp
@@ -1,5 +1,5 @@
 /*
- *  Copyright © 2017-2023 Wellington Wallace
+ *  Copyright © 2017-2024 Wellington Wallace
  *
  *  This file is part of Easy Effects.
  *
--- a/include/application_ui.hpp
+++ b/include/application_ui.hpp
@@ -1,5 +1,5 @@
 /*
- *  Copyright © 2017-2023 Wellington Wallace
+ *  Copyright © 2017-2024 Wellington Wallace
  *
  *  This file is part of Easy Effects.
  *
--- a/include/apps_box.hpp
+++ b/include/apps_box.hpp
@@ -1,5 +1,5 @@
 /*
- *  Copyright © 2017-2023 Wellington Wallace
+ *  Copyright © 2017-2024 Wellington Wallace
  *
  *  This file is part of Easy Effects.
  *
--- a/include/autogain.hpp
+++ b/include/autogain.hpp
@@ -1,5 +1,5 @@
 /*
- *  Copyright © 2017-2023 Wellington Wallace
+ *  Copyright © 2017-2024 Wellington Wallace
  *
  *  This file is part of Easy Effects.
  *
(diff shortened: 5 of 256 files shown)


wwmm:
>please, test it. No issues on my side.

As far as I can see the PR did not introcude any issue. But now that I am testing I ǘe notice a problem that may have been happening for a while. The bands gain value are not being applied. The other settings seems to have take effect but the gains remain at zero while loading some presets from https://github.com/grpan/Oratory1990-APO-Presets/tree/main/presets. 

I wonder if it is related to locale settings. I have my system LANG set to English but `LC_NUMERIC=pt_BR.UTF-8`.

wwmm:
>I wonder if it is related to locale settings. I have my system LANG set to English but LC_NUMERIC=pt_BR.UTF-8.

Probably not. Otherwise the quality values would probably be wrong too...

Digitalone1:
@wwmm  I make the tests setting the English language and the band gain is imported correctly. Can you share which preset is triggering the issue?

wwmm:
> @wwmm I make the tests setting the English language and the band gain is imported correctly. Can you share which preset is triggering the issue?

I've chosen at random https://github.com/grpan/Oratory1990-APO-Presets/blob/main/presets/1More%20Quad%20Driver.txt and https://github.com/grpan/Oratory1990-APO-Presets/blob/main/presets/64%20Audio%20N8.txt

Digitalone1:
Oh, maybe I understood, the presets there use the comma as decimal separator. I'm using other APO presets with the point as decimal separator.

wwmm:
> Oh, maybe I understood, the presets there use the comma as decimal separator. I'm using other APO presets with the point as decimal separator.

Oh... This is probably the reason...

wwmm:
I've never read APO presets specifications. They can be written with different kind of separators? That seems really bad...

Digitalone1:
> > Oh, maybe I understood, the presets there use the comma as decimal separator. I'm using other APO presets with the point as decimal separator.
> 
> Oh... This is probably the reason...

Handling these formats is a mess. But I remember from APO documentation that the point is used us decimal separator. We also handle the comma as thousands separator for frequencies. This would require a more complex regex.

wwmm:
> > > Oh, maybe I understood, the presets there use the comma as decimal separator. I'm using other APO presets with the point as decimal separator.
> > 
> > 
> > Oh... This is probably the reason...
> 
> Handling these formats is a mess. But I remember from APO documentation that the point is used as decimal separator. We also handle the comma as thousands separator for frequencies. This would require a more complex regex.

Considering how popular Oratory presets are I am surprised no one opened an issue about this yet... Or maybe EE users that need APO presets are getting them from AutoEQ where maybe this mess does not happen.

Digitalone1:
> I've never read APO presets specifications. Can they be written with different kind of separators? That seems really bad...

Here they are: https://sourceforge.net/p/equalizerapo/wiki/Configuration%20reference/

No comma there. And I will stick to it honestly.

wwmm:
>Or maybe EE users that need APO presets are getting them from AutoEQ where maybe this mess does not happen.

I looked at some AutoEQ presets and they indeed use point as decimal separator.

wwmm:
Well. I am fine with things as they are.

Digitalone1:
I made a PR to fix the issue on their side.

https://github.com/grpan/Oratory1990-APO-Presets/pull/3

PR #2932, opened 2024-02-24
Fix preset name in message dialog

Digitalone1:
This PR fix the following issue in the message dialog when a corrupted preset is loaded:

Schermata del 2024-02-24 17-43-53

This is supposed to show the preset name which is missing because we try to convert the string using `get_translated()` util, but we get an empty string because the plugin name has a trailing `#` (added when the multiple instances were introduced).

This has been fixed retrieving the base name from `get_base_name()`, before passing it to `get_translated()`. Now fhe preset name is shown, but unfortunately the string remains in English without being translated in the local language. I will open a separate issue for this.

Besides, since `missing_plugin_box()` ui helper uses the same `get_translated()` util, I changed the parameter name to `base_name` to better understand at a first sight that the function needs the base name, not the one with trailing `#`.

### DIFF
--- a/include/ui_helpers.hpp
+++ b/include/ui_helpers.hpp
@@ -45,7 +45,7 @@ void show_autohiding_toast(AdwToastOverlay* toast_overlay,
                            const uint& timeout = 5U,
                            const AdwToastPriority& priority = ADW_TOAST_PRIORITY_HIGH);
 
-auto missing_plugin_box(const std::string& name, const std::string& package) -> GtkWidget*;
+auto missing_plugin_box(const std::string& base_name, const std::string& package) -> GtkWidget*;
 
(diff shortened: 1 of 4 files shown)


wwmm:
> but unfortunately the string remains in English without being translated in the local language. I will open a separate issue for this.

>preset_load_error.emit(_("Preset Not Loaded Correctly"), plugin_translated + _("One or More Parameters Have a Wrong Format"));

I think that in the past I tried to do string concatenation inside a `_()` when one of the strings was only defined at runtime and the translation also failed. I have the feeling that `_()` only accepts string literals.

Digitalone1:
> I think that in the past I tried to do string concatenation inside a `_()` when one of the strings was only defined at runtime and the translation also failed. I have the feeling that `_()` only accepts string literals.

See #2934

PR #2965, opened 2024-03-09
Disable many clang-tidy rules

vchernin:
This isn't ideal, but these are failing and leaving them failing slows down ci. Ideally some of these should be fixed

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -53,6 +53,14 @@ Checks: 'bugprone-*,cert-*,clang-analyzer-*,concurrency-*,cppcoreguidelines-*,hi
   -performance-inefficient-vector-operation,
   -readability-else-after-return,
   -bugprone-exception-escape,
+  -bugprone-empty-catch,
+  -bugprone-switch-missing-default-case,
+  -cppcoreguidelines-misleading-capture-default-by-value,
+  -cert-env33-c,
+  -misc-include-cleaner,
+  -modernize-type-traits,
+  -performance-avoid-endl,
+  -readability-simplify-boolean-expr,
   '
 # todo, clang-analyzer-core.NullDereference is also suppressed in the code itself with NOLINT, consider fixing that too
 


wwmm:
> Ideally some of these should be fixed

Now that Arch repositories finally have clang 17 I will be able to slowly fixing them more comfortably. Let's keep this PR open for now.

wwmm:
Once of the clang's suggestions to fix 
>No header providing "std::string_literals::operator""s" is directly included (fix available)

Somehow breaks compilation in our CI image https://github.com/wwmm/easyeffects/actions/runs/8211463593/job/22460430610. But it works on my computer when compiling the code on gcc. weird

vchernin:
> But it works on my computer when compiling the code on gcc.

It might be because clang in CI uses `libc++` not `libstdc++`.

vchernin:
It might be good to disable the `misc-include-cleaner` rule as it is causing a lot of repeated warnings, and I think fixing it fully would require changing a lot of includes. Frankly I am not entirely sure why it is finding so much to complain about.

wwmm:
> It might be good to disable the `misc-include-cleaner` rule (as done in this PR) as it is causing a lot of repeated warnings, and I think fixing it fully would require changing a lot of includes. Frankly I am not entirely sure why it is finding so much to complain about.

I have already implemented it in many files. So now I will go all the way :smile: . Most of the repeated warnings left are related to out `_ui` files. And that is because all of them have the same patterns that trigger this warning. As vscode offers a fast way to apply the include headers suggestions it is no going to take long to apply them to all our `_ui` files.

That being said many of the inlcude suggestions feels dumb. The only reason I am implementing them is because I think that the new style will help to move to c++ modules when the compilers become mature enough for we to use it.

But not all is bad. Some include headers were really questionable although everything was working. For example at some point removing a header not sued by `presets_manager.hpp` caused a compilation failure in `equalizer_ui.cpp`. This level of coupling does not make sense even if the compilation works. So there is some value in all of this.

wwmm:
The number of warnings has been greatly reduced. And some of the ones left only happen in the CI image:

- The warning related to pipe_manager and preferences_general does not happen on my computer. The header is already there.
- The one about `#include <execution>` also seems a false positive. It does not happen on my computer and if the header is removed as the log from the CI image suggests the compilation fails because of the missing header.

Of the warnings that I also see on my computer I have no proper solution for https://github.com/wwmm/easyeffects/pull/2965#issuecomment-1986697993. Although `bits/basic_string.h` fixes the warning on my computer it breaks compilation in the CI image. And the header has a comment saying it is a private header not meant to be directly included. So even if there is a way to make libc++ to work with it it is not a good idea.

wwmm:
I see now that I have to do some tests running `ninja clang-tidy`. So far I have been looking only at vscode output but there are things that are shown running `ninja clang-tidy` that are not shown by vscode diagnostic. Curious...

vchernin:
> I see now that I have to do some tests running `ninja clang-tidy`. So far I have been looking only at vscode output but there are things that are shown running `ninja clang-tidy` that are not shown by vscode diagnostic. Curious...

Since clang tidy is so slow I would not be surprised if vscode skips some rules or overzealously caches output.

wwmm:
>Since clang tidy is so slow I would not be surprised if vscode skips some rules or overzealously caches output.

It is really weird. There are warnings visible in the output of `ninja clang-tidy` that are not shown inside vscode. But some of these warnings are false positive. And to make things even stranger I do not think I have seen them in the CI image output.

wwmm:
In any case the warning

> No header providing "std::string_literals::operator""s" is directly included (fix available)

is by far the one that happens the most now.

vchernin:
> And to make things even stranger I do not think I have seen them in the CI image output.

Since arch now has semi recent clang available, we can probably switch the clang tidy CI job to arch and switch it to `libstdc++`. It should make it somewhat more consistent with your machine. The alpine job could be kept to check for `libc++` compatibility.

vchernin:
Should this be closed? I can only see a `clang-analyzer-core.CallAndMessage` error and a few `misc-include-cleaner` errors causing failures in master.

wwmm:
> Should this be closed? I can only see a `clang-analyzer-core.CallAndMessage` error and a few `misc-include-cleaner` errors causing failures in master.

I think it can be. I still do not have a solution for the warnings that are left. But at the same time it does not feel right to disable the check just because of them.

vchernin:
> I think it can be. I still do not have a solution for the warnings that are left. But at the same time it does not feel right to disable the check just because of them.

It may be sensible to add `NOLINTNEXTLINE` in more places like we are already doing for the glib macro. Personally I think scattering `NOLINTNEXTLINE` in the codebase is preferable over having the clang tidy ci job always failing, if it normally passes it is easier to spot regressions. https://github.com/wwmm/easyeffects/blob/71536c67b46ba0122fdeb210a22d414ccd0f857d/src/deesser_ui.cpp#L80-L81

PR #3065, opened 2024-04-12
Backward compatibility for deprecated presets

Digitalone1:
I don't know if there's a better way to do this. Anyway, I tested the old json key and it's working.

@wwmm Please test with your old/new presets.


wwmm:
> I don't know if there's a better way to do this. Anyway, I tested the old json key and it's working.
> 
> @wwmm Please test with your old/new presets.

It is fine. I do not see the need to change the code.

PR #3088, opened 2024-04-22
Flatpak presets

vchernin:
This should be fairly complete. I also did some fixes for the flathub build being broken due to rnnoise updates.

A couple of questions:
- ~~should the `irs` and `rnnoise` subfolders also be added to the flatpak packages for forward compatibility? Right now it only checks `output` and `input`. I mostly wrote this before I realized irs and rnnoise were actively supported now, but it should be fairly trivial to add.~~ it was easy to fix so I did it anyways.
- the `com.github.wwmm.easyeffects` vs `io.github.wwmm.easyeffects` issue (see the updated docs) is a bit of a nuisance and might confuse packagers. Flathub has a fairly nice mechanism to rename packages which could do now to avoid pain later on, and then there is no ambiguity what the names of the flatpak packages are.

### DIFF
--- a/COMMUNITY_PRESETS_GUIDELINES.md
+++ b/COMMUNITY_PRESETS_GUIDELINES.md
@@ -47,6 +47,101 @@ Noise Reduction effect.
 - The folders above mentioned should contain a subdirectory with the name
 associated to the installed package. In example:
 - - `XDG_DATA_DIR/easyeffects/output/package-name`
+- The name of the package should be something descriptive e.g. `easyeffects-presets-LoudnessEqualizer`. If your package contains a group of presets without a particular "theme", naming a package `easyeffects-USERNAME-presets` is reasonable.
+
+### Flatpak
+
+Flatpak extension packages do not use `$XDG_DATA_DIRS`, and instead place and search for files under a different directory (detailed in examples below). Besides this detail, flatpak and distribution packages should behave the same in regard to community presets.
+
+In order to package a flatpak on flathub you do the following general steps:
+
+1. Create AppStream metainfo file (should be stored in your preset repository as it is not flatpak-specific).
+2. Create flatpak json file
+3. Submit to flathub, described following [flathub's guide](https://docs.flathub.org/docs/for-app-authors/submission/).
+
+#### Step by step instructions
+
+1. Decide on a name. It should be e.g. `io.github.wwmm.easyeffects.Presets.PRESET_PACKAGE_NAME`. The name must be consistent across the files used for flatpak otherwise the package will not build/work properly.
+
+> [!NOTE]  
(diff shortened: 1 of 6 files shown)


COMMUNITY_PRESETS_GUIDELINES.md
```diff
@@ -47,6 +47,101 @@ Noise Reduction effect.
 - The folders above mentioned should contain a subdirectory with the name
 associated to the installed package. In example:
 - - `XDG_DATA_DIR/easyeffects/output/package-name`
+- The name of the package should be something descriptive e.g. `easyeffects-presets-LoudnessEqualizer`. If your package contains a group of presets without a particular "theme", naming a package `easyeffects-USERNAME-presets` is reasonable.
```
vchernin, line 50: Is this a reasonable change? Later on this doc gives examples like `wwmm-presets`. Putting the username in the package name makes sense to avoid collisions, but the username is also not very useful information as it doesn't explain what the package provides. 

At the very least this should doc be consistent, so if this lined isn't preferred it should be removed.

wwmm:
>the com.github.wwmm.easyeffects vs io.github.wwmm.easyeffects issue (see the updated docs) is a bit of a nuisance and might confuse packagers. Flathub has a fairly nice mechanism to rename packages which could do now to avoid pain later on, and then there is no ambiguity what the names of the flatpak packages are.

I am confused. What is the actual impact of this? Is there the need to change something in the code to make both sides consistent with each other? Or only the location the presets are saved are affected?

vchernin:
> I am confused. What is the actual impact of this? Is there the need to change something in the code to make both sides consistent with each other? Or only the location the presets are saved are affected?

There is no direct impact from this, as it stands this setup works and users will not notice. The location isn't really affected either.

The flatpak package name is `com.github.wwmm.easyeffects`, and that name is also used in the repository by the icons, metainfo file, etc. It is well established consistent name. However, the "extension point" for flatpak extension packages is `io.github.wwmm.eayeffects` which means the extensions need to use that as their package name. As it stands it works correctly since everything is carefully using the right name. But if a packager puts the name in the wrong place things won't work and it is an easy typo to make. This is mostly only relevant for flatpak since that is where these reverse dns names are visibly used.

It is one of those things that someone is going to read and wonder why on earth we are using 2 very similar names when we could just use one, and it is not entirely obvious which is correct unless you have correct documentation on hand and know how flatpak extensions work. Users installing easyeffects flatpak via the command line might also wonder if `io.github.wwmm.easyeffects.Preset.Foo` is some different software from `com.github.wwmm.easyeffects` altogether. 

The only way to "fix" this would be rename every instance of `com.github.wwmm.easyeffects`, in any other scenario we will have a mix. I would be ok with renaming the flathub package as it should be mostly seamless for users https://docs.flathub.org/docs/for-app-authors/maintenance#end-of-life-rebase.

COMMUNITY_PRESETS_GUIDELINES.md
Digitalone1: It's OK. No problem.

wwmm:
Ok. I think I understand. For now let's keep things as they are here in EE repo. I am not sure if I will get used to the "io" alternative because I never think about that site...

PR #3091, opened 2024-04-22
Add navigation box in the collapsed sidebar

Digitalone1:
We added the ability to reduce the sidebar hiding the plugins list, but at the moment this feature is somewhat limited because it forces the user to reenable the plugins list to show other effects in the pipeline.

Considering that collapsing the sidebar is useful only if the user has few plugins, maybe they want the ability to move between them easily without having the plugins list always visible. So I added a navigation box to move up and down through the pipeline, only visible when the sidebar is collapsed.

This widget does not affect the existing configurations since it's disabled by default, just like the collapsed sidebar.

### DIFF
--- a/data/ui/plugins_box.ui
+++ b/data/ui/plugins_box.ui
@@ -52,7 +52,6 @@
                     <object class="GtkBox" id="navigation_box">
                         <property name="halign">center</property>
                         <property name="valign">center</property>
-                        <property name="margin-top">6</property>
                         <property name="spacing">6</property>
                         <property name="orientation">vertical</property>
                         <child>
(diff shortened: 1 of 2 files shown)


wwmm:
It is far from being a priority but while working with the kirigami widget that acts as a collapsible panel I've noticed that they switch the icon of the button that hides/show the panel depending on the panel state. When the panel is hidden they use the same icon we have now to suggest a movent to the right that will reveal the panel. And when the panel is visible they use a very similar icon where the ">" sign is inverted "<" to indicate that the panel will move to the left and be hidden.

PR #3103, opened 2024-04-24
Make sure to initialize irs and rnnoise user directories

vchernin:
Otherwise we will try copying to a non existent directory in a fresh install.

### DIFF
--- a/src/presets_manager.cpp
+++ b/src/presets_manager.cpp
@@ -107,6 +107,8 @@ PresetsManager::PresetsManager()
 
   create_user_directory(user_input_dir);
   create_user_directory(user_output_dir);
+  create_user_directory(user_irs_dir);
+  create_user_directory(user_rnnoise_dir);
   create_user_directory(autoload_input_dir);
   create_user_directory(autoload_output_dir);
 


Digitalone1:
Maybe that was already done in Convolver/RNNoise. I don't remember, please check it out.

vchernin:
> Maybe that was already done in Convolver/RNNoise. I don't remember, please check it out.

From a brief inspection I do not see the directories being created anywhere. In any case it seems reasonable to create all these directories together as done in this PR.

wwmm:
> Maybe that was already done in Convolver/RNNoise. I don't remember, please check it out.

Yes:

https://github.com/wwmm/easyeffects/blob/7263b305c6478cc41f06a221d2f18164619a721a/src/convolver_ui.cpp#L659
https://github.com/wwmm/easyeffects/blob/7263b305c6478cc41f06a221d2f18164619a721a/src/rnnoise_ui.cpp#L452

We should probably keep just one of these places.

wwmm:
> > Maybe that was already done in Convolver/RNNoise. I don't remember, please check it out.
> 
> From a brief inspection I do not see the directories being created anywhere. In any case it seems reasonable to create all these directories together as done in this PR.

It makes sense. The presets manager class always runs. The plugins only if the user puts them in the pipeline.

Digitalone1:
> We should probably keep just one of these places.

Since the RP has been merged, you can delete those in Convolver and RNNoise UI then.

wwmm:
> > We should probably keep just one of these places.
> 
> Since the PR has been merged, we should remove those in Convolver and RNNoise UI then.

Done

PR #3111, opened 2024-04-26
Improve community presets management

Digitalone1:
This is a big load of improvements for community presets management.

Since we used the kernel/model-name to search the files, we could fall in some misunderstandings when multiple packages had the same filename. In example `irs/pack-1/device.irs` and `irs/pack-2/device.irs` were always picking one file, no matter the preset chosen.

In order to avoid this behavior, I had to introduce a mechanism to signal the selection of a community package name. In practice, when a community preset is loaded, we save the community package name to gsettings. Otherwise, when a local preset is loaded, the previous community package name is reset. With this key, we can select a file from the right path and the search is also restricted to the paths belonging to chosen package, not everyone installed. 

Now we are also aware if we have loaded a preset from the local path or from a community package. But that required to introduce the pipeline type (in/out) and global settings to all plugins instances.

@wwmm one thing to notice is how the community package is retrieved after the import/try button is clicked. The only way I found it to work is setting the package name label object on the gtk button object. See [here](https://github.com/Digitalone1/easyeffects/blob/master/src/presets_menu.cpp#L297-L299). I tried to pass only the label string, but I kept getting errors on the compiler. Don't know if there's a better way to do it.

Another fix I noticed is the use of `g_settings_get_string` in various parts of the code which it seems to require to free the string, but we didn't. I switched them to the relative `util::gsettings_get_string` which frees it internally.

This should be heavily tested. I did it on my system and it seems good at the moment. I hope there's no edge case where race conditions could show hidden bugs.

@vchernin Please test you too on Flatpak with the newly created extensions. Thanks.

### DIFF
--- a/data/schemas/com.github.wwmm.easyeffects.gschema.xml
+++ b/data/schemas/com.github.wwmm.easyeffects.gschema.xml
@@ -10,11 +10,17 @@
         <key name="use-dark-theme" type="b">
             <default>false</default>
         </key>
-        <key name="last-used-input-preset" type="s">
-            <default l10n="messages">"Presets"</default>
+        <key name="last-loaded-input-preset" type="s">
+            <default>""</default>
         </key>
-        <key name="last-used-output-preset" type="s">
-            <default l10n="messages">"Presets"</default>
+        <key name="last-loaded-output-preset" type="s">
+            <default>""</default>
+        </key>
+        <key name="last-loaded-input-community-package" type="s">
+            <default>""</default>
+        </key>
+        <key name="last-loaded-output-community-package" type="s">
+            <default>""</default>
         </key>
         <key name="window-maximized" type="b">
             <default>false</default>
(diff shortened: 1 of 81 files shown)


wwmm:
>But that required to introduce the pipeline type (in/out) and global settings to all plugins instances.

This is fine. The only place where it feels strange to do this is in the the class that manages them like `StreamOutputEffects`. Differently from the plugins they can not be of other types.

>one thing to notice is how the community package is retrieved after the import/try button is clicked. The only way I found it to work is setting the package name label object on the gtk button object. See [here](https://github.com/Digitalone1/easyeffects/blob/master/src/presets_menu.cpp#L297-L299). I tried to pass only the label string, but I kept getting errors on the compiler. Don't know if there's a better way to do it.

I do not think there is a better way to deal with this inside the listview factory callbacks. The limitations of the C language hit hard in this place.

>This should be heavily tested. I did it on my system and it seems good at the moment. I hope there's no edge case where race conditions could show hidden bugs.

Just looking at the code nothing caught my attention. I will merge it to make it easier to test both on native as well as on Flatpak packages.

Digitalone1:
> This is fine. The only place where it feels strange to do this is in the the class that manages them like `StreamOutputEffects`. Differently from the plugins they can not be of other types.

Yes, you're right.

Issue #3193, opened 2024-06-19
app crashes when we uploads APO or reset the eq 


Ayush583-Aswal:
### EasyEffects Version

7.1.6

### What package are you using?

Flatpak (Flathub)

### Distribution

Ubuntu 24.04 LTS

### Describe the bug

when i upload APO to an Eq the app crashes and force quit/wait option comes. This is also the case for , when we are resetting the eq filters.
this error was coming in terminal: (easyeffects:2): GLib-GIO-WARNING **: 21:46:40.879: g_settings_set_value: value for key 'band9-frequency' in schema 'com.github.wwmm.easyeffects.equalizer.channel' is outside of valid range

 where band9 filter was something like this : 'Filter 9: Off PK Fc 0 Hz Gain 0 dB Q 0'

### Expected Behavior

the app should be able to switch the APO filters and reset the filters without getting force quit alert. like we can change the eq settings easily in wavelet(on android) without getting any error and it should function smoothly

### Debug Log

<details><summary>Debug Log</summary><br/>

```
Paste your log here
```
</details>

wwmm:
Attach here on github the preset you tried to load. The last time I tested the APO import it was fine. But i is not a feature I have a reason to use. So maybe there was a regression somewhere.

Digitalone1:
> where band9 filter was something like this : 'Filter 9: Off PK Fc 0 Hz Gain 0 dB Q 0'

Indeed 0 Hz frequency is not valid. But I'd like the app won't crash on out of bounds values. I didn't know gsettinsg does not have an internal check and it crashes directly...

wwmm:
> > where band9 filter was something like this : 'Filter 9: Off PK Fc 0 Hz Gain 0 dB Q 0'
> 
> Indeed 0 Hz frequency is not valid (10 is minimum). But I'd like the app won't crash on out of bounds values. I didn't know gsettinsg does not have an internal check and it crashes directly...

Hum... I wonder if the crash was really inside gsettings. In this bug report there is only a warning coming from it.

Digitalone1:
I think it's in gsettinsg, but it does not provide an easy way to check the key range.

We have an unused util for that, but https://docs.gtk.org/gio/method.SettingsSchemaKey.get_range.html says it's better to not use it in normal programs.

Then there's
https://docs.gtk.org/gio/method.SettingsSchemaKey.range_check.html but dealing with GVariant is very annoying.

[...]

PR #3195, opened 2024-06-20
Improve APO/GraphicEQ import/export

Digitalone1:
* Added gsettings key range check for double/float values inside APO/GEQ import. I don't like this, the functions are now very complex because of this implementation, but it's the only way to make a check on valid values. Besides I suspect native gsettings is only emitting a warning on out of bounds values, but on Flatpak it may crash directly because the settings engine is different. If this is the case, those key checks will fix a big issue on that platform.  

* Since `band_width` affect the spectrum of the selected band depending on the filter type chosen, I reset it at the default on APO/GEQ preset import. Maybe this value should be properly calculated on every filter, but we don't know how to do it. So at the moment the best thing to do is to reset it to the default, avoiding to change the result depending on which previous value the user set.
* Reset `band_width` also inside `on_calculate_frequencies`. This seems a good thing to do since `band_width` affects the band spectrum, but maybe I could be wrong here. @wwmm What do you think?

* Reset `input-gain` on GEQ import. APO import always sets the preamp (0 db if not specified). But if we import an APO preset with a specific preamp and then import a GEQ preset, we retain the previous input gain, which I think it's not what we expect. GEQ does not have a preamp control, so I assume it's always 0 db by default.

* Improved APO preset parsing. The best thing to do is to follow this algorithm: https://github.com/lsp-plugins/lsp-plugins-para-equalizer/blob/master/src/main/ui/para_equalizer.cpp#L1502 - Which sets a specific quality or frequency on certain filters, no matter which is the value inside the preset file.

* Added the Band-Pass filter in the APO preset parsing. This is not present in the LSP algorithm, but we may add it anyway. @wwmm Any suggestion here?

* `EasyEffectsToApoFilter` and `ApoToEasyEffectsFilter` converted to maps. There's no real reason to use them as unordered maps. Besides the EasyEffectsToApoFilter was specifying the same LSP EQ filter multiple times which is useless. It makes sense to traslate LSP filters only to APO filters that have center frequencies.

* In APO export, write the gain value only for filters that need it. See APO config documentation:  https://sourceforge.net/p/equalizerapo/wiki/Configuration%20reference/#filter - The gain is mandatory only for Peak, Low-Shelf and High-Shelf filters.

* Use `tags::equalizer` namespace on the entire files so we don't have to specify it multiple times.

* Added number label on top of each bands inside Equalizer UI. This may seem unnecessary, but I think it's very useful. When I was testing the import feature, comparing APO presets bands with the results in the UI was not so easy because sometimes I had to count the bands. Indeed if the user has a big number of bands, looking at one in the center it's not easy for them to know which is the number of the chosen band. With a label on top, it's more straightforward.

And at the last I'd like to point out that I'm a bit skeptical on the use of `band-mute` in APO import. At the moment the APO OFF value sets the `band-mute`, but is it really the same? OFF means the band is not processed, so we still hear it, right? Which is different from the muting which is just emitting silence, I guess? Maybe like band gain to `-infinite`. Maybe it should be better to set `band-type` on Off when APO OFF is specified. The problem seems that APO does not have a concept of "muting the band". @wwmm What do you think?

### DIFF
--- a/data/ui/equalizer_band.ui
+++ b/data/ui/equalizer_band.ui
@@ -7,6 +7,18 @@
         <property name="margin-bottom">6</property>
         <property name="spacing">6</property>
         <property name="orientation">vertical</property>
+
+        <child>
+            <object class="GtkLabel" id="band_number_label">
+                <property name="width-chars">8</property>
+                <property name="label">N</property>
+
+                <style>
+                    <class name="dim-label" />
+                </style>
+            </object>
+        </child>
+
         <child>
             <object class="GtkMenuButton">
                 <property name="margin-bottom">6</property>
--- a/src/equalizer_band_box.cpp
+++ b/src/equalizer_band_box.cpp
@@ -34,6 +34,8 @@
 
 namespace ui::equalizer_band_box {
 
+using namespace tags::equalizer;
+
 struct Data {
  public:
   ~Data() { util::debug("data struct destroyed"); }
@@ -56,6 +58,8 @@ struct _EqualizerBandBox {
 
   GtkPopover* popover_menu;
 
+  GtkLabel* band_number_label;
+
   GSettings *settings, *app_settings;
 
   Data* data;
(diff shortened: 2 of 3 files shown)


wwmm:
>Reset band_width also inside on_calculate_frequencies. This seems a good thing to do since band_width affects the band spectrum, but maybe I could be wrong here. @wwmm What do you think?

It has been so long since I thought about this LSP `band_width` parameter that I do not remember anymore exactly what it is doing. But as far as the `on_calculate_frequencies` code goes I think it is fine to reset it. It would only be a problem if the bad)width parameter conflicted with the `quality` value calculated inside `on_calculate_frequencies`. But I think this was not the cause.

>The problem seems that APO does not have a concept of "muting the band". @wwmm What do you think?

Hum... I think that we probably should set them to `OFF`.

Digitalone1:
@vchernin can you please check this PR importing the APO preset [here](https://github.com/wwmm/easyeffects/issues/3193#issuecomment-2179277321) on Flatpak and see if something weird happens? Thanks.

Issue #3224, opened 2024-06-28
Drops and errors on system load


Massimo-B:
### EasyEffects Version

7.1.3

### What package are you using?

Gentoo

### Distribution

Gentoo 23.0

### Describe the bug

On higher system loads I have drops and errors in pw-top when running easyeffects. This also happens when CPU is not fully used (about 50%), but with Chrome and a MS Teams session with video. Load AVG is about 20 with some btrfs workers in the back. The issue is usually solved when killing easyeffects.
I also disabled the spectrum now as it also did a lot of load.

### Expected Behavior

_No response_

### Debug Log

```
S   ID  QUANT   RATE    WAIT    BUSY   W/Q   B/Q  ERR FORMAT           NAME                                                                                                                                                                                                                                                                                                                                                               
S   29      0	   0    ---     ---   ---   ---     0                  Dummy-Driver
S   30      0	   0    ---     ---   ---   ---     0                  Freewheel-Driver
R   75   1024  48000  31,9us   0,5us  0,00  0,00    2    S16LE 1 48000 alsa_input.usb-Microsoft_Microsoft___LifeCam_HD-3000-02.mono-fallback
R   92      1     25  19,4us   6,1us  0,00  0,00    6       F32LE 1 25  + PulseAudio Volume Control
R   76   1024  48000  73,5us 105,6us  0,00  0,00  550    S32LE 2 48000 alsa_output.pci-0000_00_1b.0.analog-stereo
R  124      1     25  28,9us  28,9us  0,00  0,00    8       F32LE 1 25  + PulseAudio Volume Control
R   77   1024  48000  65,8us   0,9us  0,00  0,00    2    S32LE 2 48000 alsa_input.pci-0000_00_1b.0.analog-stereo
R  128      1     25  34,3us  22,2us  0,00  0,00   10       F32LE 1 25  + PulseAudio Volume Control
I  269      0	   0   0,0us   0,0us  ???   ???     0                  ee_test_signals
R   74   1024  48000  15,2ms   1,6us  0,71  0,00  169    S16LE 2 48000 alsa_input.usb-Samson_Technologies_Samson_C01U-00.analog-stereo
R   35      0	   0   8,3us  12,7us  0,00  0,00    5     F32P 2 48000  + easyeffects_sink
R   36      0	   0  33,9us  13,1us  0,00  0,00  202     F32P 2 48000  + easyeffects_source
[...]

PR #3242, opened 2024-07-09
spectrum: use gtk tick callback to trigger spectrum redraws

tleb:
Related issue: #3224. Related previous PR: #3231.

I have written all informations inside commit messages. The last one is the bulk of the series.

 - We start by some cleanup: we switch all `std::vector` in Spectrum to being arrays. This avoids `.resize()` calls inside the constructor. The goal is simpler code, it is not an optimization.
 - We then switch `PluginBase::bypass` to become atomic. This allows removing the `data_mutex` locking in `Spectrum`.
 - We switch `fftw_ready` to become atomic as well.
 - We use `std::memmove()` to shift previous samples. This improves the average time spent inside `process()`.
 - We finish by switching the main patch. It changes the way spectrum chart redraw is being scheduled.

After this patch series, `Spectrum::process()` does not appear anymore in a `perf` recording, and we get a 60FPS (or more depending on display) spectrum chart.

---

Some stats on my workstation. 256 quantum, 48kHz sample rate, ran 45 seconds, no other load.

```
           before      after
calls        8912       8923
avg (µs)       65.0       34.2
stdev          16.9        7.6
min             0.0        7.0
5th perc       39.0       20.0
25th perc      54.0       29.0
median         69.0       35.0
75th perc      74.0       40.0
95th perc      82.0       43.0
max           151.0      127.0
```

---

I use this small patch to dump chart redraw interval timings:

<details>
```diff
diff --git a/src/chart.cpp b/src/chart.cpp
index 8e9d53b13..91f609905 100644
--- a/src/chart.cpp
+++ b/src/chart.cpp
@@ -249,6 +249,8 @@ void set_x_data(Chart* self, const std::vector<double>& x) {
   });
 }

+static gint64 time_since_last_drawing;
+
 void set_y_data(Chart* self, const std::vector<double>& y) {
   if (!GTK_IS_WIDGET(self) || y.empty()) {
     return;
@@ -280,6 +282,10 @@ void set_y_data(Chart* self, const std::vector<double>& y) {
   }

   gtk_widget_queue_draw(GTK_WIDGET(self));
+
+  gint64 time = g_get_monotonic_time();
+  printf("%.0f ms\n", (double)(time - time_since_last_drawing) / (1e3));
+  time_since_last_drawing = time;
 }

 void on_pointer_motion(GtkEventControllerMotion* controller, double xpos, double ypos, Chart* self) {
```
</details>

### DIFF
--- a/include/plugin_base.hpp
+++ b/include/plugin_base.hpp
@@ -21,6 +21,7 @@
 
 #include <gio/gio.h>
 #include <glib.h>
+#include <atomic>
 #include <pipewire/filter.h>
 #include <sigc++/signal.h>
 #include <spa/utils/hook.h>
@@ -91,7 +92,8 @@ class PluginBase {
 
   bool package_installed = true;
 
-  bool bypass = false;
+  std::atomic<bool> bypass = {false};
+  static_assert(std::atomic<bool>::is_always_lock_free);
 
   bool connected_to_pw = false;
 
--- a/include/spectrum.hpp
+++ b/include/spectrum.hpp
@@ -19,6 +19,7 @@
 
 #pragma once
 
+#include <atomic>
 #include <fftw3.h>
 #include <sigc++/signal.h>
 #include <sys/types.h>
@@ -51,7 +52,7 @@ class Spectrum : public PluginBase {
 
   auto get_latency_seconds() -> float override;
 
-  sigc::signal<void(uint, uint, double*)> power;  // rate, nbands, magnitudes
+  std::tuple<uint, uint, double*> compute_magnitudes();  // rate, nbands, magnitudes
 
  private:
   std::atomic<bool> fftw_ready = false;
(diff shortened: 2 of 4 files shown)


wwmm:
Thanks! For some reason I never thought about using std::atomic. So far no problems on my computer.

wwmm:
I feel like I will forget in a few days what is being done by those bitwise operations but it is not a big concern :smile: . I rarely touch bitwise operations. It may be a good thing to add some comments explaining how the code around `compare_exchange_weak` works.

tleb:
I thought about extracting that *mess* into a separate abstraction and comment it (excessively), but my C++ skills (or lack of) meant I wasn't confident enough to create an abstraction and reason about what happens to data (copying, references, templating, move semantic, operators to implement, all that crap).

I'm saying that because I wouldn't be surprised that the double-buffering thing could be used elsewhere. Whenever you want to extract a lot of data from realtime into GUI thread. "A lot of data" is everything bigger than what can be done atomically, ie often `2*sizeof(void*)`.

wwmm:
>I thought about extracting that mess into a separate abstraction and comment it (excessively), but my C++ skills (or lack of) meant I wasn't confident enough to create an abstraction and reason about what happens to data (copying, references, templating, move semantic, operators to implement, all that crap).

It is fine even if it is a superficial explanation :-).

>About atomics: it might mean we could remove a bunch of mutexes from plugins. I haven't looked much yet. I might go for a hunt!

I think we do not have many. But the few we have probably can be replaced by an atomic.

I have started to work on a Qt port in a secondary branch. It will be a long journey. But I can already say that any work on optimizations heavily dependent on gtk or gsettings calls should be avoided. The files unrelated to the ui or the database probably will be reused without major changes. So it is fine to invest energy in the classes managing the filters.

PR #3257, opened 2024-07-20
spectrum: explanation comment for double-buffering mechanism

tleb:
Follow up to #3242. The double buffering strategy is some lock-free code that is hard to understand, so insert a comment brick explaining what it does (with some references).

Feel free to edit if you want to add/remove details.


wwmm:
Thanks! :-)

Issue #3312, opened 2024-08-11
[request] ability to delay the spectrogram/meters by ... frames / milliseconds


magicgoose:
Would help to be able to force spectrogram/meters to be in sync with what's physically being played at the end of the pipeline,
especially when using a Bluetooth output device.

If this can be saved per device (just like effects) then that's extra convenient.

Also (not 100% sure), some bluetooth devices report their latency, might be good to use that as the starting value.

wwmm:
>Would help to be able to force spectrogram/meters to be in sync with what's physically being played at the end of the pipeline

The spectrum is at the end of the pipeline. We process the buffers as soon as PipeWire sends them. The thing is that the speed at which those buffers come also depends on the target server latency and whatever else is affecting PipeWire's decisions. And to make things worse a client like EasyEffects does not have the necessary information to synchronize with what the players are doing.  Maybe PipeWire has this information. But not us. So it is unlikely we can do something about it.

magicgoose:
Thanks for analysis, understandable. Still a fully manual knob might possibly make sense?
something like passing the snapshots of (spectrum fft info) through a queue of N items, and the user just has to type N in.

wwmm:
> Thanks for analysis, understandable. Still a fully manual knob might possibly make sense? something like passing the snapshots of (spectrum fft info) through a queue of N items, and the user just has to type N in.

Maybe an average over a time window. The correct thing to do would probably be to calculate the fft once in a larger window. But this would have performance side effects.

violetmage:
I think magicgoose is more concerned with the synchronization of when a sample of the spectrum is displayed and when the sound reaches their ears.

It seems to me that a simple "delay line" type of knob would be sufficient. A FIFO from the end of the pipeline into the spectrum with a user configurable length or something. This would probably require changes to the pipeline manager, as the audio path would need to branch at the end of the pipeline; one branch going into the delay line towards the spectrum, and the other directly out the system audio device.

wwmm:
>It seems to me that a simple "delay line" type of knob would be sufficient.

I think it is not so simple. But if it is this can be tested now. We have a delay plugin that can be put in the pipeline. And the stereo tools can also add delay.

But based on what I see on my computer I do not think that we need delay. When listening to a song with easy to notice bass I can see that the spectrum jumps exactly when the bass kicks.

violetmage:
[...]

PR #3340, opened 2024-08-30
spectrum compensating delay (av sync)

violetmage:
Adds ability to delay the data going to the spectrum display, so that in case of an audio device with long latency, the visual feedback can remain synchronized with what is audible.

### DIFF
--- a/include/spectrum.hpp
+++ b/include/spectrum.hpp
@@ -70,6 +70,7 @@ class Spectrum : public PluginBase {
   std::array<float, n_bands> right_delay_input_array;
   std::span<float> left_delay_input;
   std::span<float> right_delay_input;
+  
   std::array<float, n_bands> left_delayed_array;
   std::array<float, n_bands> right_delayed_array;
   std::span<float> left_delayed;
--- a/include/spectrum.hpp
+++ b/include/spectrum.hpp
@@ -66,8 +66,10 @@ class Spectrum : public PluginBase {
   std::array<float, n_bands> real_input;
   std::array<double, n_bands / 2U + 1U> output;
 
-  std::array<float, n_bands> left_delayed;
-  std::array<float, n_bands> right_delayed;
+  std::array<float, n_bands> left_delayed_array;
+  std::array<float, n_bands> right_delayed_array;
+  std::span<float> left_delayed;
+  std::span<float> right_delayed;
 
   std::array<float, n_bands> latest_samples_mono;
 
--- a/include/spectrum.hpp
+++ b/include/spectrum.hpp
@@ -66,6 +66,10 @@ class Spectrum : public PluginBase {
   std::array<float, n_bands> real_input;
   std::array<double, n_bands / 2U + 1U> output;
 
+  std::array<float, n_bands> left_delay_input_array;
+  std::array<float, n_bands> right_delay_input_array;
+  std::span<float> left_delay_input;
+  std::span<float> right_delay_input;
   std::array<float, n_bands> left_delayed_array;
   std::array<float, n_bands> right_delayed_array;
   std::span<float> left_delayed;
--- a/include/spectrum.hpp
+++ b/include/spectrum.hpp
@@ -66,10 +66,8 @@ class Spectrum : public PluginBase {
   std::array<float, n_bands> real_input;
   std::array<double, n_bands / 2U + 1U> output;
 
-  std::array<float, n_bands> left_delayed_array;
-  std::array<float, n_bands> right_delayed_array;
-  std::span<float> left_delayed(left_delayed_array);
-  std::span<float> right_delayed(right_delayed_array);
+  std::array<float, n_bands> left_delayed;
+  std::array<float, n_bands> right_delayed;
 
   std::array<float, n_bands> latest_samples_mono;
 
--- a/include/spectrum.hpp
+++ b/include/spectrum.hpp
@@ -66,6 +66,11 @@ class Spectrum : public PluginBase {
   std::array<float, n_bands> real_input;
   std::array<double, n_bands / 2U + 1U> output;
 
+  std::array<float, n_bands> left_delayed_array;
+  std::array<float, n_bands> right_delayed_array;
+  std::span<float> left_delayed(left_delayed_array);
+  std::span<float> right_delayed(right_delayed_array);
+
   std::array<float, n_bands> latest_samples_mono;
 
   std::array<float, n_bands> hann_window;
--- a/src/spectrum.cpp
+++ b/src/spectrum.cpp
@@ -143,19 +143,19 @@ void Spectrum::process(std::span<float>& left_in,
 
   if ( n_samples < n_bands) {
     // Drop the oldest quantum.
-    std::memmove(left_delay_input[0], left_delay_input[n_samples], (n_bands - n_samples) * sizeof(float));
-    std::memmove(right_delay_input[0], right_delay_input[n_samples], (n_bands - n_samples) * sizeof(float));
+    std::memmove(&left_delay_input_array[0], left_delay_input_array[n_samples], (n_bands - n_samples) * sizeof(float));
+    std::memmove(&right_delay_input_array[0], right_delay_input_array[n_samples], (n_bands - n_samples) * sizeof(float));
 
     // Copy the new quantum.
     for (size_t n = 0; n < n_samples; n++) {
-      left_delay_input[n_bands - n_samples + n] = left_in[n];
-      right_delay_input[n_bands - n_samples + n] = right_in[n];
+      left_delay_input_array[n_bands - n_samples + n] = left_in[n];
+      right_delay_input_array[n_bands - n_samples + n] = right_in[n];
     }
   } else {
     // Copy the latest n_bands samples.
     for (size_t n = 0; n < n_bands; n++) {
-      left_delay_input[n] = left_in[n_samples - n_bands + n];
-      right_delay_input[n] = right_in[n_samples - n_bands + n];
+      left_delay_input_array[n] = left_in[n_samples - n_bands + n];
+      right_delay_input_array[n] = right_in[n_samples - n_bands + n];
     }
   }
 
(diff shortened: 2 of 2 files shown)


Digitalone1:
If you're not doing it, please use `clang-format` on new changes. If you're using VS Code, just install [clang](https://archlinux.org/packages/extra/x86_64/clang/) package and [clang-format](https://github.com/xaverh/vscode-clang-format) extension.

wwmm:
@violetmage this implementation can be simpler. There is no need to create so many arrays. Two `std::vector<float>` are enough. Imagine they are named `left_tmp` and `right_tmp`. After resizing them in the setup function you can do something like this when connecting the ports
```
lv2_wrapper->connect_data_ports(left_in, right_in, left_tmp, right_tmp);
```

And when initializing the mono array
```
latest_samples_mono[n] = 0.5F * (left_tmp[n] + right_tmp[n]);
```

In other words you can still use the `std::span` that provide the input data. THe only difference is that the delay plugin will output to temporary vectors and the mono sample will be created from them instead of the input `std::span`.

wwmm:
>After resizing them in the setup function 

Something like 
```
left_tmp.resize(n_samples);
right_tmp.resize(n_samples);
```
With this approach the FFT code can stay untouched. The difference is that it will process delayed samples.

violetmage:
The reason I didn't do it that way is because I hadn't realized until now that setup() gets called whenever the rate or quantum changed, so I had incorrectly assumed that messing with the size of the buffers wouldn't work. I'll fix that now.

Do you have any advice on how to go about detecting the reported output device latency, or where the spectrum could get that number from in the current code?

wwmm:
>Do you have any advice on how to go about detecting the reported output device latency, or where the spectrum could get that number from in the current code?

That is actually a good question. I realized now that I have no idea :smile: . In order to pass the plugin extra latency to PipeWire we call [pw_filter_update_params](https://github.com/wwmm/easyeffects/blob/f0bb548985003480699f547b1e6f0ccb28b44242/src/plugin_base.cpp#L167C3-L167C26). But the PipeWire's header that defines it does not have something like "get_params" https://gitlab.freedesktop.org/pipewire/pipewire/-/blob/master/src/pipewire/filter.h. I hope we do not have to iterate over each the nodes ports to get this information. It would make it more annoying...

wwmm:
If I remember well the latency offset video players use to do the playback compensation flows from the soundcard ports up to the video player node ports. So in the worse case scenario where this has to be extracted from the ports we would probably have to read the soundcard node input ports.

violetmage:
Playing around with pw-dump, it seems to me that device nodes don't have a `node.max-latency` value, instead they have the `latency.internal.rate` and `latency.internal.ns` properties.

AFAICT, it's up to the driver to decide what to do with the `latency.internal.whatever` variables, and any setting seems accepted by pipewire; both values can be null, just one, or both might be set to a value.

It also seems that there's a more consistently formatted "Latency" object in the "Props" section of a node's "params". The values in the "Latency" object seem to be the automatically calculated ones that get passed along the graph. I have no idea how this translates into libpipewire code atm, but I think that reading from this "Latency" object is the right approach. It appears to me that the "Latency" object is present on every connected node, regardless of if it's an ALSA device or virtual sink, so I assume that this is an abstraction intended for this purpose.

Any advice here would be appreciated.

Digitalone1:
If those values are strings, we have an util to convert to floats/double.

violetmage:
That's good, but I'm still not sure about what functions to call to get any latency values from pipewire.

violetmage:
If I understand correctly, to get info about a node there would need to be something like this:

https://github.com/wwmm/easyeffects/blob/fee51c5e86ebf462d7758bc9b18ac76b6097bc4e/src/pipe_manager.cpp#L529-L543

Except instead of checking for SPA_PARAM_Format or whatever, it would look for a SPA_PARAM_Latency object, and go from there?

wwmm:
I've found the PipeWire's issue where the lack of latency compensation when EE was running was discussed in the past https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/3888. The latency value we need is the one set to the ports and not the one set to the nodes.

wwmm:
According to what one of PW developers said

>The latency values are supposed to propagate through filters etc. How it works is that Pipewire sets latency value of an output port, to the latency value of the input port it is connected to. The filter whose latency was changed, is then supposed to update the value in its input port(s), and it propagates from there to the next filters and finally to applications using it.

It is enough to read the properties of the output ports of the last filter in the chain. In our case the `output level meter` filter.

wwmm:
> If I understand correctly, to get info about a node there would need to be something like this:
> 
> https://github.com/wwmm/easyeffects/blob/fee51c5e86ebf462d7758bc9b18ac76b6097bc4e/src/pipe_manager.cpp#L529-L543
> 
> Except instead of checking for SPA_PARAM_Format or whatever, it would look for a SPA_PARAM_Latency object, and go from there?

We will have to do something like this for the ports. I do not remember if we already have this implemented. Probably not.

violetmage:
I'm still a bit lost on how to go about getting the latency of a specific node. I've not been able to really get anywhere by reading the pipewire docs.

I'm pretty sure that getting the output latency should be pipe_manager's responsibility, but I'm not sure of where in pipe_manager the latency check should go, or of the best way to go about making that information available to easyeffects plugins in the pipeline.

wwmm:
> I'm still a bit lost on how to go about getting the latency of a specific node. I've not been able to really get anywhere by reading the pipewire docs.
> 
> I'm pretty sure that getting the output latency should be pipe_manager's responsibility, but I'm not sure of where in pipe_manager the latency check should go, or of the best way to go about making that information available to easyeffects plugins in the pipeline.

As the word latency may have slightly different meanings depending on where it is being used things may get a little confusing. But as we want to do the sae that video players do what we need is the latency offset that may have been set by the user or PipeWire. And according to what PipeWire's developers said in that issue they set this information to the port objects and not to the nodes.

As we never had a reason to extract this kind of information from ports it will be necessary to add more callbacks on our side. The first thing will probably be to define a `on_port_info` similar to https://github.com/wwmm/easyeffects/blob/3085129a1bf993e331098ae3a85dde59dc105159/src/pipe_manager.cpp#L271. And then inside of it do something like https://github.com/wwmm/easyeffects/blob/3085129a1bf993e331098ae3a85dde59dc105159/src/pipe_manager.cpp#L531.

I did not have time to look at PipeWire's sources but I imagine that some kind of `on_port_event_param` similar to  https://github.com/wwmm/easyeffects/blob/3085129a1bf993e331098ae3a85dde59dc105159/src/pipe_manager.cpp#L605 will also be necessary. If everything goes well the next step would be to figure out how to actually read the latency offset being set to the output ports of our `output level meter`.

The last step would be to savethat information somewhere. Probably in a new public member of the `pipe_manager` class.

wwmm:
One of the annoyances of having to deal with the port objects is that we have to iterate over the node list to find out which of them has an id equal to the `node.id` set in the port.

wwmm:
It may be easier at this moment to create a configuration option in our preferences window. Probably something like `Latency Offset` with units in milliseconds `ms` and the value controlled through a spinbutton. The range `0 ms -> 1000 ms` is more or less what Pavucontrol uses if I remember well.

violetmage:
It's better than nothing. I'll start working on adding a preferences option.

violetmage:
This PR is a basic implementation of #3312

violetmage:
Also, skimming through the fftw3 docs again, I found this:
[https://fftw.org/fftw3_doc/Wisdom-Import.html](https://fftw.org/fftw3_doc/Wisdom-Import.html)

Specifically, the `fftw_import_system_wisdom()` function seems interesting for easyeffects, as it could be combined with the fftw planner flag `FFTW_WISDOM_ONLY`, in order to load an optimal plan without incurring the long compute time on startup. Then, a simple null check could allow falling back to `FFTW_ESTIMATE`. Something like this:
```
if ( fftw_import_system_wisdom() ) {
  plan = fftwf_plan_dft_r2c_1d(static_cast<int>(n_bands), real_input.data(), complex_output, FFTW_WISDOM_ONLY);
}

if (plan == NULL) {
  plan = fftwf_plan_dft_r2c_1d(static_cast<int>(n_bands), real_input.data(), complex_output, FFTW_ESTIMATE);
}
```

wwmm:
> Also, skimming through the fftw3 docs again, I found this: https://fftw.org/fftw3_doc/Wisdom-Import.html
> 
> Specifically, the `fftw_import_system_wisdom()` function seems interesting for easyeffects, as it could be combined with the fftw planner flag `FFTW_WISDOM_ONLY`, which tells fftw to load an optimal plan without implicitly falling back to a mode with a long computation time on startup if no wisdom is found. Then, a simple null check could allow falling back to `FFTW_ESTIMATE`. Something like this:
> 
> ```
> if ( fftw_import_system_wisdom() ) {
>   plan = fftwf_plan_dft_r2c_1d(static_cast<int>(n_bands), real_input.data(), complex_output, FFTW_WISDOM_ONLY | FFTW_PATIENT);
> }
> 
> if (plan == NULL) {
>   plan = fftwf_plan_dft_r2c_1d(static_cast<int>(n_bands), real_input.data(), complex_output, FFTW_ESTIMATE);
> }
> ```

FFTW documentations talks about `/etc/fftw/wisdom` as the file providing the system wisdom. But at least on my system this does not exist. So unless the user figure out how to create the system file with the fftw wisdom this is not going to do anything.

Based on what I remember drawing the spectrum in the screen is more costly than calculating the FFT. The problem was doing it in the realtime thread. But this isn't the case anymore.

violetmage:
Fair enough, just thought it was worth mentioning.

Anyways I've tested the dev flatpak of this PR, and it seems to all work well. Not sure if you want to merge it as is, or keep the PR open until the auto detection is implemented.

wwmm:
> Fair enough, just thought it was worth mentioning.
> 
> Anyways I've tested the dev flatpak of this PR, and it seems to all work well. Not sure if you want to merge it as is, or keep the PR open until the auto detection is implemented.

I do not think we have to wait. The automatic detection can be done in the future.

wwmm:
Does the current code handle the case where LSP is not installed? As LSP is an optional plugin there has to be spectrum without it.

wwmm:
> Does the current code handle the case where LSP is not installed? As LSP is an optional plugin there has to be spectrum without it.

It seems it does. I will merge this PR.

wwmm:
Hum... I am doing some tests with a latency test video https://www.youtube.com/watch?v=TjAa0wOe5k4 and it does not seem that the delay control is delaying the spectrum.

violetmage:
Just retested using that (very useful) video, the dev flatpak built from bebe78d seems to work fine for me.

violetmage:
Same test with the dev flatpak built from 42c41a7 also checks out for me.

wwmm:
I've finally understood what is going on. It is working. The "issue" was that I decided to set the maximum possible delay of 1000 ms. As the interval of the beats in the video is also of 1 second it feels like there is no delay. But it is just hard to perceive with the eyes. Ass soon as I selected half a second things behaved as expected :smile: .

8BallBomBom:
This seems to have broke the Spectrum on my end 🫨

wwmm:
> This seems to have broke the Spectrum on my end 🫨

How?

8BallBomBom:
Not entirely sure as i haven't looked deeper 🤔 
But i'll investigate.

PR #3494, opened 2024-11-11
QML: prefer identity over equality

Digitalone1:
@wwmm You used equality on `Qt.Key_PageUp`, but not on `Qt.Key_PageDown`. I suppose you wanted to do identity for both (I tested both keys). Besides it's quite safe to use identity on array lenghts and indexes.


wwmm:
>I suppose you wanted to do identity for both (I tested both keys). Besides it's quite safe to use identity on array lenghts and indexes.

Thanks! As Javascript can be a little tricky with this I always try to use the comparison operator that does not do implicit type casting.

PR #3507, opened 2024-11-16
Implement the Maximizer

Digitalone1:
@wwmm I copied what you did in various files and finally I managed to implement the Maximizer. I chose this plugins because it's simple, but it has not been easy to complete it. Many times I had a core dump for typos or because I forgot to write different lines.

Please make a double check and test it on your system. It works for me.

1. As I understood, now all plugins have two `process` methods and the one with `probes` should be used only when there are sidechains, right? 
2. Does the `Reduction` level produced by the Maximizer is in db? I'm pretty sure it is, but we don't report the unit in the libadwaita UI.
3. Should the native UI be visible for the Maximizer?

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -25,6 +25,7 @@ target_sources(easyeffects PRIVATE
     local_server.cpp
     lv2_wrapper.cpp
     main.cpp
+    maximizer.cpp
     output_level.cpp
     plugin_base.cpp
     pw_manager.cpp
@@ -39,7 +40,7 @@ target_sources(easyeffects PRIVATE
     resources.qrc
 )
 
-target_include_directories(easyeffects PRIVATE 
+target_include_directories(easyeffects PRIVATE
     ${LIBPIPEWIRE_INCLUDE_DIRS}
     ${LIBLILV_INCLUDE_DIRS}
     ${LIBEBUR128_INCLUDE_DIRS}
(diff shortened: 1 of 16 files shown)


wwmm:
>As I understood, now all plugins have two process methods and the one with probes should be used only when there are sidechains, right?

For some reason now the compiler complains if we do not override both of them. I wonder why...  

>Does the Reduction level produced by the Maximizer is in db? I'm pretty sure it is, but we don't report the unit in the libadwaita UI.

Yes. According to the file `/usr/lib/lv2/ZaMaximX2.lv2/ZaMaximX2_dsp.ttl` it is in decibel.

>Should the native UI be visible for the Maximizer?

Yes. I've updated the repository with this. It is curious that the maximizer plugin still has the `ceiling` control. The ttl file does not have it anymore.

wwmm:
>but it has not been easy to complete it. Many times I had a core dump for typos or because I forgot to write different lines.

I am not sure if there is a way to reduce the amount of lines we need to write. In a way this is a problem we have always had. Adding new plugins was a pain before and still is. In my opinion a bit less painful. But still painful...

What I have been doing to minimize the chances of segmentation faults is doing the work on the c++ classes before touching what is needed on QML. C++ code starts first and has to live longer that the QML sections that need them.

Digitalone1:
> Yes. I've updated the repository with this. It is curious that the maximizer native window still has the `ceiling` control. The ttl file does not have it anymore.

I didn't try the native window button. You could implement it for the Maximizer (in reality I didn't remember if you already did for other plugins, can't check it right now...)

PR #3509, opened 2024-11-17
Optimize sed regex

Digitalone1:
I tested them on https://sed.js.org/

I couldn't test them on the schema files, please do it @wwmm, but I don't think there will be issues.

Basically it's always better to avoid `.*` because it consumes all the line (or worse, the entire string file) and then the engine has to go back (backtracking) to check the possible matches.

I don't know if sed supports possessive quantifiers (.*+) that avoid backtracking, but the new regexes should reduce it and improve performance anyway.

### DIFF
--- a/util/convert_schema_to_kcfg.sh
+++ b/util/convert_schema_to_kcfg.sh
@@ -13,7 +13,7 @@ cd "$UTIL_DIR"
 
 # s/regexp/replacement/
 cat $SCHEMA |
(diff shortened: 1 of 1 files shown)


wwmm:
Thanks! There was only a small regression  I Was able to fix. The line that parses the range was failing because it did not take into account negative numbers. I've replaced `([0-9.]*)` by `([-+]?[0-9.]*)` and it seems fine now.

Digitalone1:
Nice. Now that I think about it, in the first one, replace `<key\sname` with `<key\s+name`. It could not work with multiple whitespaces.

PR #3547, opened 2024-12-09
Improve user input cleaning and fix translations

Digitalone1:
Qml has it's own regex syntax, but maybe it's better if we stick to the js one which is more intuitive. In both perl and js regex engine, `${100}` means "match exactly 100 times", but we want between 1 and 100. Using `{1,100}` is not confusing, even if `{100}` works the same way in Qml.

Besides I wanted to exclude strings with whitespaces only, but it's not possible with the single regex validator in Qml. Asking ChatGTP I got the following which works in js: https://regex101.com/r/ohg7BD/1

Using it in Qml instead, the user can still add a string with only multiple whitespaces (i.e `              `). So I modified the check on the trimmed version of the string.

Made the same for the preset manager, plus an additional replace string method to remove the preset `.json` final extension, if specified (since we automatically add it, it's not wanted because we end up writing a file with double extension).

There are also other issues, but I will list them in #2960

### DIFF
--- a/src/contents/ui/Autogain.qml
+++ b/src/contents/ui/Autogain.qml
@@ -156,7 +156,7 @@ Kirigami.ScrollablePage {
                         id: momentary
 
                         label: i18n("Momentary")
-                        unit: i18n("LUFS")
+                        unit: "LUFS"
                         from: Common.minimumDecibelLevel
                         to: 10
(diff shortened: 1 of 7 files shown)


Digitalone1:
Made also improvements to translations.

PR #3557, opened 2024-12-16
Convert string to double using std::stod

### LINKED ISSUE #3532: Can't change some parameters in Effects
### EasyEffects Version

7.1.9

NixOS

### Distribution

NixOS Unstable (25.05)

### Describe the bug

Can´t change Voice Activity Probability for some reason.

Also happens on Gate Sidechain Filters Frequencies. I've been using the native window as a fallback.

[Screencast_20241203_112210.webm](https://github.com/user-attachments/assets/62ee220d-7701-4f72-8bce-32d7096c2850)

### Expected Behavior

Value should be accepted when in range. If not in range or not accepted for another reason, it should be prompted by an error somewhere.

### Debug Log

```
Paste your log here
```
</details>

### Additional Information

_No response_

zaps166:
Closes #3532

### DIFF
--- a/src/ui_helpers.cpp
+++ b/src/ui_helpers.cpp
@@ -161,20 +161,18 @@ auto parse_spinbutton_input(GtkSpinButton* button, gdouble* new_value, const boo
 
   gtk_spin_button_get_range(button, &min, &max);
 
-  std::istringstream str(gtk_editable_get_text(GTK_EDITABLE(button)));
+  std::string s = gtk_editable_get_text(GTK_EDITABLE(button));
 
-  if (!lower_bound) {
-    auto s = str.str();
+  // Workaround to fix locale issue in #3532; see PR #3557.
+  s.erase(std::remove(s.begin(), s.end(), ' '), s.end());
 
-    util::str_trim_start(s);
+  if (!lower_bound && s.starts_with(_("-inf"))) {
+    *new_value = util::minimum_db_d_level;
 
-    if (s.starts_with(_("-inf"))) {
-      *new_value = util::minimum_db_d_level;
-
-      return TRUE;
-    }
+    return TRUE;
   }
 
+  std::istringstream str(s);
   str.imbue(ui::get_user_locale());
 
   auto v = 0.0;


wwmm:
I think that this kind of workaround should be applied by the users in an environmental variable and not in EasyEffects code. It doesn't really fix the problem. It just hides it while creating another. For example in my system I have its language set to English while having `LC_NUMERIC=pt_BR.UTF-8`. Numbers follow the Brazilian localization just fine. But if I set `LC_NUMERIC=C` then all the numbers are forced to  the English locale. Personally I do not mind having numbers in English. But if we do that then the new issues opened by users will be "EasyEffects does not follow my locale anymore".

Considering that not everybody is affected something must have changed in a system library related to EasyEffects. For example in the past PipeWire was the one messing with locale when it shouldn't. After I reported the issue to them they fixed. Maybe something else decide to do something similar. Or gtk4 has a new bug. But I would everybody to be hit by it in this case.

zaps166:
Understand, ok.

> Or gtk4 has a new bug.

The problem is I don't know GTK4 at all. Only did some very small things in GTK3. I've never seen such issue in Qt :sweat_smile: Do you use GTK4 spin boxes or it's from external library? Notice that not all are affected, only some of them with integers, hmm.

zaps166:
@wwmm Another attempt to fix it (no workaround this time). I've checked it with `C` locale and `pl_PL.utf8` locale. Could you look at this?

wwmm:
> Another attempt to fix it (no workaround this time). I've checked it with `C` locale and `pl_PL.utf8` locale. Could you look at this?

At least on my computer is as fine as it was before these changes. So I have no reason to not accept the patch. But before merging this it would be good to also have @Digitalone1 feedback. It has been a long time but I think we had a reason in the past to save the user locale for later reapplication. Maybe something has changed in the system libraries and `std:stod` is the way to go now.

Digitalone1:
@wwmm In the past we had issues with user locale and saving the locale helped us handling the various localization formats.

This is because the strings are shown in the user locale and we need to parse them using the proper locale to return them as correct double/float. 

I don't remember at the moment why string stream was preferred over `stod`, but it worked for long time. Now it does not work for some users, but we do not know the exact reason.

As far as I know, both string stream and `stod` are locale dependent, but stream stream was preferred in the past.

Maybe recently something is changed, but I'm afraid this would fix the issue for some users and broke it for others. We don't even know what changed and maybe it could be a bug that should be fixed by system libraries, not us.

Digitalone1:
> Maybe something has changed in the system libraries and `std:stod` is the way to go now.

If it's the way to go, it would be better since we won't have to pass the locale every time. We could also remove the locale variable and the util functions to set/get it. 

But I have the suspect it's not so easy and there should be some other issue lying hidden somewhere. Locale has always been a pain to work with.

I hope this issue won't be present in the Qt port.

zaps166:
> Maybe something has changed in the system libraries and std:stod is the way to go now.

`std::stod` requires locales to be set globally (e.g. `std::locale::global(std::locale(""));`), but looks like it's working in EasyEffects.

---

Tests with `glibc 2.40+r16+gaa533d58ff-2`.

---

Test code in EasyEffects:
```c++
std::cerr << std::stod("55 Hz")   << std::endl;
std::cerr << std::stod("55.3 Hz") << std::endl;
std::cerr << std::stod("55,4 Hz") << std::endl;
```

Locale: `pl_PL.utf8`:
```
55
55
55.4
```

Locale: `C`:
```
55
55.3
55
```

Locale: `pt_BR.UTF-8`:
```
55
55
55.4
```

---

Test code in EasyEffects:
```c++
auto str_to_d = [](const char *text) {
  std::istringstream str(text);
  str.imbue(ui::get_user_locale());

  double v = 0.0;
  if (str >> v)
    return v;
  return std::numeric_limits<double>::quiet_NaN();
};
std::cerr << str_to_d("55 Hz")   << std::endl;
std::cerr << str_to_d("55.3 Hz") << std::endl;
std::cerr << str_to_d("55,4 Hz") << std::endl;
```

Locale: `pl_PL.utf8`:
```
nan
55
55.4
```

Locale: `C`:
```
55
55.3
55
```

Locale: `pt_BR.UTF-8`:
```
55
nan
55.4
```

Digitalone1:
> `std::stod` requires locales to be set globally (e.g. `std::locale::global(std::locale(""));`), but looks like it's working in EasyEffects.

Maybe this is the reason we preferred stringstream since we didn't want to set the locale, but only to get it from the one that the user set in their system.

Anyway, this seems a clear system library bug to me because there's so reason a `55 Hz` integer would return `nan`.

Digitalone1:
So I tried to reproduce on my system generating a Polish locale, but it's working. Tested on Arch upstream package.

https://github.com/user-attachments/assets/d9419caa-357d-4132-8ed0-ed164cdcdbf4

Where it's not working, it could be likely a broken system locale.

At least on Arch, uncomment the locale you want in `/etc/locale.conf`, then regenerate them with `sudo locale-gen`. Restart Easy Effects and test it.

Digitalone1:
New test:

https://github.com/user-attachments/assets/c28a93aa-a3b2-4a84-9995-11e58fc4053c

I think it'a a bug in stringstream. Using `stod` is a workaround. I'm pretty sure this has been introduced recently, otherwise this issue should have been already appear two years ago.

zaps166:
> I think it'a a bug in stringstream. Using stod is a workaround. I'm pretty sure this has been introduced recently, otherwise this issue should have been already appear two years ago.

I checked with libstdc++ and libc++ - same results with PL locale.

I noticed it can't convert integer value with locales like PL when it has any text suffix after space like ` Hz`, so `55 Hz` - FAIL (EasyEffects case), but `55Hz` - OK.

zaps166:
> As I said before, it seems to affect only some locales, not all of them.

Yes, which is very strange behavior.

> It could be recent. I suppose you already tested this in the past with your locale, otherwise you already reported it years/months ago.

That's interesting I never noticed it long time ago (noticed few months ago). I haven't used integer spin boxes or it's really a recent bug.

> The issue should be reported to C++ devs anyway, but I don't know where.

Both libstdc++ and libc++ has this issue now. Maybe bug is in both or in glibc which is the base of entire system? :thinking: 

---

Alternative "workaround" for some locales:

```diff
diff --git a/src/ui_helpers.cpp b/src/ui_helpers.cpp
index da0adb000..90ef33ea5 100644
--- a/src/ui_helpers.cpp
+++ b/src/ui_helpers.cpp
@@ -161,13 +161,10 @@ auto parse_spinbutton_input(GtkSpinButton* button, gdouble* new_value, const boo
 
   gtk_spin_button_get_range(button, &min, &max);
 
-  std::istringstream str(gtk_editable_get_text(GTK_EDITABLE(button)));
+  std::string s = gtk_editable_get_text(GTK_EDITABLE(button));
+  util::str_trim_start(s);
 
   if (!lower_bound) {
-    auto s = str.str();
-
-    util::str_trim_start(s);
-
     if (s.starts_with(_("-inf"))) {
       *new_value = util::minimum_db_d_level;
 
@@ -175,6 +172,9 @@ auto parse_spinbutton_input(GtkSpinButton* button, gdouble* new_value, const boo
     }
   }
 
+  s.erase(std::remove(s.begin(), s.end(), ' '), s.end());
+
+  std::istringstream str(s);
   str.imbue(ui::get_user_locale());
 
   auto v = 0.0;
```

zaps166:
@Digitalone1 

You're right, tested stringstream on few-years-old (not updated) Arch Linux on my very old laptop - it's working correctly there :smile:

Digitalone1:
@zaps166 Yes, using  `s.erase(s::remove(s.begin(), s.end(), ' '), s.end());` is a safer workaround since it discards all the whitespaces.

Indeed I think the bug with the Polish locale (and other ones affected) is that the trailing whitespace is not discarded and this causes the wrong convertion.

The `>>` operator extracts the characters until it founds a whitespace. For example:

```c++
std::string input = "41 3.14 false hello world";
std::istringstream stream(input);

int n;
double f;
bool b;

// The following
stream >> n >> f >> std::boolalpha >> b;
// will fill the variables like this: n = 41; f = 3.14; b = false
```

But the behavior of this operator is also affected by the locale. Since with the Polish locale `50` is converted correctly but `50  ` with trailing whitespaces fails, I think some locales are not discarding the trailing whitespaces.

So erasing all the whitespaces should be a good workaround without using `stod`. Besides, with this trick we don't even have the need to call the `str_trim_start` util to check for `-inf`.

@zaps166 Please, make those changes when you have time.

zaps166:
@Digitalone1 Done!

Digitalone1:
> @Digitalone1 Done!

@zaps166 The `trim_start` is not needed because the `erase` does it already.

Please, replace the following

```c++
  std::string s = gtk_editable_get_text(GTK_EDITABLE(button));

  // Workaround to fix locale issue in #3532; see PR #3557.
  s.erase(std::remove(s.begin(), s.end(), " "), s.end());

  if (!lower_bound && s.starts_with(_("-inf"))) {
    *new_value = util::minimum_db_d_level;

    return TRUE;
  }

  std::istringstream str(s);
  str.imbue(ui::get_user_locale());
```

zaps166:
@Digitalone1

`util::str_trim_start` removes also ``\n\r\t\v\f`` - are you sure it's not need to remove them before checking for `-inf`?

Digitalone1:
> @Digitalone1
> 
> `util::str_trim_start` removes also `\n\r\t\v\f` - are you sure it's not need to remove them before checking for `-inf`?

That was a generic function like javascript `trim` method of the string class that removes all the whitespaces from the start and end of the text, but here we need only to remove spaces. In theory tabs and new lines can't even be inserted since  their keys are supposed to confirm the text written inside the spinbutton. Regarding vertical tab and form feed, I don't even know how to input them, but we can ignore them.

Digitalone1:
@zaps166 If you paste a new line from the text editor, it's not a big issue rejecting the input. If plain `-inf` works as expected and you tested the spinbuttons with your locale, it's OK for me to be merged.

Digitalone1:
Does someone have an account to report the issue on [gcc bugzilla](https://gcc.gnu.org/bugzilla/)? New account creation is restricted.

zaps166:
Also Clang's libc++ has the same problem, so both: libstdc++ and libc++ have bug.

Very old libstdc++ works correctly (tested).

PR #3560, opened 2024-12-17
Add active_preset and active_presets arguments.

N0tACyb0rg:
Adds the `active_preset <input|output>` and `active_presets`, which return either the active presets for input and output, or returns the preset for a specific category.


wwmm:
Thanks! I will add a note to our Qt port roadmap so I do not forget to add this there too.

PR #3562, opened 2024-12-18
UI improvements

Digitalone1:
I added decimals to input/output spin buttons of all effects.

But I noticed we have lots of dB controls with various decimals and stepSize. In order to make them similar, I set 2 decimals on all of them. But since the up/down key changes them slowly, I set `0.1` as stepSize. So the user have more precision, but still a faster way to change them through up/down keys.

Is it OK for you? Should I make something similar also for ms controls since we have differences on decimals/stepSize also for these ones.

I added also a top margin for rows/cards that were too close to the widget above.

### DIFF
--- a/src/contents/ui/Autogain.qml
+++ b/src/contents/ui/Autogain.qml
@@ -160,7 +160,7 @@ Kirigami.ScrollablePage {
                         from: Common.minimumDecibelLevel
                         to: 10
                         value: 0
-                        decimals: 0
+                        decimals: 1
 
                         anchors {
                             left: parent.left
@@ -177,7 +177,7 @@ Kirigami.ScrollablePage {
                         from: Common.minimumDecibelLevel
                         to: 10
                         value: 0
-                        decimals: 0
+                        decimals: 1
 
                         anchors {
                             left: parent.left
@@ -194,7 +194,7 @@ Kirigami.ScrollablePage {
                         from: Common.minimumDecibelLevel
                         to: 10
                         value: 0
-                        decimals: 0
+                        decimals: 1
 
                         anchors {
                             left: parent.left
@@ -211,7 +211,7 @@ Kirigami.ScrollablePage {
                         from: Common.minimumDecibelLevel
                         to: 10
                         value: 0
-                        decimals: 0
+                        decimals: 1
 
                         anchors {
                             left: parent.left
@@ -228,7 +228,7 @@ Kirigami.ScrollablePage {
                         from: 0
                         to: 50
                         value: 0
-                        decimals: 0
+                        decimals: 1
 
                         anchors {
                             left: parent.left
@@ -245,7 +245,7 @@ Kirigami.ScrollablePage {
                         from: Common.minimumDecibelLevel
                         to: 10
                         value: 0
-                        decimals: 0
+                        decimals: 1
 
                         anchors {
                             left: parent.left
--- a/src/contents/ui/Autogain.qml
+++ b/src/contents/ui/Autogain.qml
@@ -79,8 +79,8 @@ Kirigami.ScrollablePage {
                         from: pluginDB.getMinValue("target")
                         to: pluginDB.getMaxValue("target")
                         value: pluginDB.target
-                        decimals: 0
-                        stepSize: 1
+                        decimals: 2
+                        stepSize: 0.1
                         unit: "dB"
                         onValueModified: (v) => {
                             pluginDB.target = v;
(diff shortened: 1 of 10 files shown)


wwmm:
I have to go to work now so only tonight I will be able to take a closer look at the changes. While porting the code I tried to keep the step size equal to the ones used in the native window. Sometimes things get confusing when our step size is smaller. 

One day I was using the mouse scroll over one of our spinbuttons and nothing was happening in the native window. So I thought the code had some kind of problem. But what was actually happening was that our step size was smaller. So it was taking some time until the value changed in the native window.

I think that in the plugins that do not have a native window it is fine to select a step size and use it in all of them. But if I was confused by the smaller step size while using the native window users probably will be too. So it may be better to match what the native window is doing.

Digitalone1:
And what about the decimals? At least can we have all at 2 or shoud we follow the native window either?

Digitalone1:
On another read, maybe I understood what you said. If our spinbutton has a stepSize of `0.01` and the same control in the native LV2 window has a stepSize of `0.1`, changing the stepSize in EE does not immediately change the displayed value in the native window.

That's because if we have `0.00` in EE and start to change it, we will only see `0.1` in native window as soon as we reach `0.10` or more in EE. Basically, we don't see values from `0.01` to `0.09` in the native window.

But in this PR I set `0.1` stepSize for all dB values (to change them faster) which is bigger than the previous one, so I suppose there are no big issues. At the moment I can't test it, but I wait for you to do it when you have time.

wwmm:
Yes. It is like you have said. I think that having a larger step size in our custom ui is fine for decibel values.  But for milliseconds and other units it is probably better to use the same precision as the native window.

wwmm:
> But since the up/down key changes them slowly, I set 0.1 as stepSize. So the user have more precision, but still a faster way to change them through up/down keys.

I've got home. And now I remember that that I had added a a callback to the `Keys.onPressed` event that uses the `page up/down` keys to make larger steps. For example consider the `Lookahead` parameter in the compressor. The mouse scroll and the up/down keys will have a step size of `0.001`. But the page up/down keys will jump by 10 times the step size used normally. What seems to be the faster way that you wanted.  It seems that it isn't really necessary to force the standard step to be larger. Or am I missing something?

I added the page up/down controls exactly because I also thought that the default step size was too small for fast changes in some cases.

Digitalone1:
Oh, I didn't know about that. So I will restore the previous stepSize. But for precision, is it OK for you to keep all dB controls on 2 decimals?

wwmm:
>Oh, I didn't know about that. So I will restore the previous stepSize.

I have the feeling that the same may happen to the users. But I can't see a way to make it obvious that the page up/down has a different behavior. Maybe once the shortcuts window is back we could put there a information about this. 

>But for precision, is it OK for you to keep all dB controls on 2 decimals?

It should be fine. Depending on what the user sets in the native window the value may be rounded by the custom window. But it is not like we fully follow the native window behavior anyway. I've noticed that LSP changes the step size depending on the range on the current value. As nice as this is replicating on our side will require more work than I am willing to put on this.

Digitalone1:
@wwmm So I restored the step size in all effects except Autogain (that has no native UI).

I took `0.1` for Autogain. Is it OK for you? Or you want `0.01`? 

Anyway, later I tested the spinbuttons and noticed a strange behavior.

When changing the value on a spinbutton with 2 decimals and step size `0.1`, the modification is not precise. For example in Autogain, if I have `-70.00` and start to change it, the values won't modify in steps of `0.1`, but maybe sometime it adds/remove `0.11`.

Can you test it on your branch before the merge? Try to add 2 decimals on Autogain dB controls and change by 0.1 step size. Sometimes it was even changing by `1.0`, so weird!

I investigated this issue and it's likely that Javascript (and QML) does not have a specific type for integers and floats (at least inside functions), so the Number class is a generic one to bundle both and them and it causes some rounding problems.

Besides, if you don't already know it, take a look at this one: https://patrickkarsh.medium.com/why-math-is-hard-in-javascript-floating-point-precision-in-javascript-41706aa7a89d

TLDR: Not always the following is true in Javascript: `0.1 + 0.2 === 0.3`

So the only way I found to get the expected step sizes at the moment is to use some tricks like `toFixed` when converting from the plugin value to the internal integer representation of the spinbutton. `toFixed` returns a string, so it has to make the implicit casting which is not ideal, but at least we have the expected precision now.

I also added the followings:
- Clamp values generated from page up/down since it shows an error when the value under/overflows the control bounds.
- Added a nullish coalescing on the regex for converting the text to plugin value since I got the error "no 0 property on null object" when I confirm an empty string into the spinbutton. Now we only get "invalid format", which is more appropriate.
- Use `Math.floor` for the `-inf` value which is more appropriate.
- Show one decimal on all Autogain levels (before only the Gain was showing the decimal precision).

wwmm:
>When changing the value on a spinbutton with 2 decimals and step size 0.1, the modification is not precise. For example in Autogain, if I have -70.00 and start to change it, the values won't modify in steps of 0.1, but maybe sometime it adds/remove 0.11.

I've noticed this when I added the page up/down shortcut. I imagined that some kind of rounding issue was happening somewhere but I did not have the time to investigate it. As far as I can see everything seems fine now after your patch.

PR #3583, opened 2024-12-23
Implement the Crossfeed

Digitalone1:
I tried to implement the Crossfeed, plus some other minor changes. 

Unfortunately this has various conflicts, so I will resolve them.

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -141,7 +141,7 @@ jobs:
         depends="lv2"
         makedepends="
         qt6-qtbase-dev
-        qt6-qtcharts-dev
+        qt6-qtgraphs-dev
         kiconthemes-dev
         kirigami-dev
         ki18n-dev
@@ -231,7 +231,7 @@ jobs:
         pipewire-devel
         pkgconf
         qt6-qtbase-devel
-        qt6-qtcharts-devel
+        qt6-qtgraphs-devel
         rnnoise-devel
         soundtouch-devel
         speexdsp-devel
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -119,138 +119,139 @@ jobs:
         if-no-files-found: error
 
 
-  alpine-linux:
-    name: Alpine Linux (stdc++)
-    runs-on: ubuntu-22.04
-    container:
-      image: alpine
+  # alpine-linux:
+  #   name: Alpine Linux (stdc++)
+  #   runs-on: ubuntu-22.04
+  #   container:
+  #     image: alpine
 
-    steps:
-    - name: Checkout repository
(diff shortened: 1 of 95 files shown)


Digitalone1:
Since I had to rebase and add the fix for QtChart header, the brance were not aligned, but now I have resolved the conflicts.

@wwmm Please take a look at the Croosfeed since I'm not really sure the implementation is the best. I used the connection since we have to call the mutex to apply the values.

To apply the presets inside the UI, I made a QML function. But I did not implemented the Crossfeed plugin preset. Maybe I will do it the next time (AFAIK, only the Autogain has the plugin preset).

wwmm:
>Please take a look at the Croosfeed since I'm not really sure the implementation is the best. I used the connections since we have to call the mutex to apply the values.

The code is fine. The connections are exactly the way I would have done them. They are needed in this case because if I remember well the internal state of the library is reset when we change them. So we have to use the mutex.

I have updated the repository using pkgconfig to find libbs2b. As the library provides it it is better than a manual search.

You will notice I've made a very subtle change in the way the crossfeed presets are handled in the QML file. THere is something about QML bindings that is quite tricky. The assignment
```
fcut.value = 700;
```
breaks the binding being done in the spinbutton through `value: pluginDB.fcut`. The standard behavior in QML is making a new binding on assignments and forgetting the ones done before. That is why in some places I have done things similar to
```
Qt.binding(function() {
                        return DB.Manager.siePluginsDB;
                    })
```
. But in the crossover case the easy solution is writing the value to the database instead of writing to the widget value property.

Digitalone1:
I thought to write them to the plugin db, but I wasn't sure the binding was bidirectional, so I chose the widget,

Is it okay to remove the duplicate headers? I saw lots of them in cpp files that was already included in the hpp one...

wwmm:
>Is it okay to remove the duplicate headers? I saw lots of them in cpp files that was already included in the hpp one...

I have been using the clang feature that helps to deal with unused headers https://clangd.llvm.org/guides/include-cleaner. But it does seem to be a little aggressive when suggesting what to include. For consistency I have been adding its suggestions as they are harmless anyone. Maybe there is a way to keep just the good parts of what it does. I do not know...

>I thought to write them to the plugin db, but I wasn't sure the binding was bidirectional, so I chose the widget,

QML does not have bidirectional bindings. initially I thought this would be annoying. But as writing the callbacks is straightforward. I do not mind it.

What happens in this case is that the update done to the database propagates to the widget. In a bidirectional binding a change in the widget value property would be sent to the database. But this is not the case in QML. What happens is that the previous binding done to the value property is lost and a new one takes its place.

Digitalone1:
I knew that a library included in the header it's not needed in the cpp file (indeed it was compiling fine). But maybe I'm wrong...

PR #3603, opened 2024-12-28
Update schema converter

Digitalone1:
With the string lists, the schema converter is a little bit easier, but it still needs to manually insert the properties and the labels (and reorder the tags if needed).

I improved the commands changing the delimiter where `/` is used, so there's no need to escape them. Since we use 4 spaces as indentation, I replaced the tabs since they are rendered differently in various editors (in Gnome Text they are reproduced by 2 spaced which does not fit good with the rest). @wwmm Please, test it.

Finally I added also `undefined` and `null` to JS `isEmpty`.

### DIFF
--- a/src/contents/ui/Common.js
+++ b/src/contents/ui/Common.js
@@ -7,6 +7,10 @@ function isEmpty(v) {
             return v.length === 0;
         case "number":
             return isNaN(v);
+        case "undefined":
+            return true;
+        case "object":
+            return v === null;
         default:
             return false;
   }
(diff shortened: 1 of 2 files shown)


wwmm:
>With the string lists, the schema converter is a little bit easier, but it still needs to manually insert the properties and the labels (and reorder the tags if needed).

It is not a big problem. AS it is it already helps a lot :+1: . Another limitation is that the integer variable has a string set as default. But we already had to edit the default anyway.

Digitalone1:
> Another limitation is that the integer variable has a string set as default.

In my tests that default is retained. You should only set the properties (with `Labels` suffix) and the labels.

PR #3606, opened 2025-01-01
Add active-preset and active-presets command line options (QT Port)

N0tACyb0rg:
Adds the stated command line options to get the active preset for a specific type, or get the active presets for both types of presets.

My only issue before I merge this is how to remove the informational text that gets outputted before and after my command line options, because it ends up looking like this:
```
util.cpp:75	directory already exists: /home/caleb/.config/easyeffects/db
Input: None
Output: Laptop
db_manager.cpp:98	Saving settings...
```
(Also I'd like to know how to exit properly instead of `std::exit(0);` lol)

### DIFF
--- a/src/command_line_parser.cpp
+++ b/src/command_line_parser.cpp
@@ -18,12 +18,15 @@
  */
 
 #include "command_line_parser.hpp"
+#include "easyeffects_db.h"
+#include "util.hpp"
 #include <qcommandlineparser.h>
 #include <qobject.h>
 #include <qtmetamacros.h>
 #include <KLocalizedString>
 #include <QApplication>
 #include <memory>
+#include <iostream>
 
 CommandLineParser::CommandLineParser(QObject* parent)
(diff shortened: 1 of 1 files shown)


wwmm:
>(Also I'd like to know how to exit properly instead of std::exit(0); lol)

Is there really the need to even call a function to exit? Assuming another instance is already running the one created to get the output will exit on its own naturally.

wwmm:
> Assuming another instance is already running the one created to get the output will exit on its own naturally.

Or at least it will once the port is done. I have to take a look at this part of the code again to remember its current state.

wwmm:
Something that may work is calling `QApplication::quit();`. It is probably better than std::exit(0). I will merge and do that.

wwmm:
>My only issue before I merge this is how to remove the informational text that gets outputted before and after my command line options, because it ends up looking like this:

I will be better once I make Qt to only print warnings when a debug environmental variable is set. I  did not have time to investigate how to do this yet. We definitely do not want EasyEffects to print soe many debug messages by default.

PR #3644, opened 2025-01-25
Update copyright notice

Digitalone1:
As the title says.

### DIFF
--- a/include/lv2_wrapper.hpp
+++ b/include/lv2_wrapper.hpp
@@ -1,5 +1,5 @@
 /*
- *  Copyright © 2017-2024 Wellington Wallace
+ *  Copyright © 2017-2025 Wellington Wallace
  *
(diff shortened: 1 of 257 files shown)


include/lv2_wrapper.hpp
```diff
@@ -284,9 +284,15 @@ class Lv2Wrapper {
   std::array<std::pair<size_t, uint>, 64> control_ports_cache;
 
   struct {
-    struct { uint left, right; } in;
```
Safari77, line 287: Not a copyright year.

Digitalone1: Yes, but we use the "format on save" option of clang-format vscode extension, so whenever we save a c++ file (or we execute the "replace all" in "search in project"), the syntax of the file is reformatted.

PR #3667, opened 2025-02-07
Fix ninja argument order

mattfbacon:
Flags after positional arguments are technically invalid and broken for samurai (ninja reimplementation in C).


wwmm:
Thanks!

PR #3732, opened 2025-03-31
Update symbolic icon design

ramdomPTM:
Updated the design of the symbolic icon.

Before:
image

After:
image

The file was optimized using svgcleaner.


wwmm:
Thanks!

PR #3740, opened 2025-04-05
Implement separate pitch parameters: cents, semitones, and octaves

mikroffarad:
This pull request introduces changes to the Pitch effect plugin to allow users to adjust the pitch using three separate parameters: **cents**, **semitones**, and **octaves**. Previously, only a single semitones parameter was available.

**Changes include:**
- **Code modifications:**  
    - Added new parameters `cents` and `octaves` to the `Pitch` class.
    - Created a new function `update_pitch()` that calculates the overall pitch shift using the formula:  
    ```
    total_semitones = semitones + (octaves * 12) + (cents / 100)
    ```
    and applies it using SoundTouch.
    - Updated signal handlers to react to changes for the new parameters.
    - Modified `init_soundtouch()` to call `update_pitch()` to set the initial pitch.
- **GSettings Schema:**  
    - Updated the GSettings schema to include new keys for `cents` and `octaves`.

- **User Interface:**  
    - Added new GtkSpinButton widgets for `cents` and `octaves` in the UI template.
     - Updated the UI bindings to include the new keys. 

**Notes:**  
- This change does not alter the underlying SoundTouch library usage; it only modifies how parameters are computed and exposed to the user.

Interface before (since v7.0.5):
image

Interface I propose (like it was in v7.0.4), only `cents`, `semitones` and `octaves` are working: 
image

Please review and let me know if any further adjustments are required.

### DIFF
--- a/data/schemas/com.github.wwmm.easyeffects.pitch.gschema.xml
+++ b/data/schemas/com.github.wwmm.easyeffects.pitch.gschema.xml
@@ -12,24 +12,18 @@
             <range min="-36" max="36" />
             <default>0</default>
         </key>
-        <key name="semitones" type="d">
-            <range min="-12.0" max="12.0" />
-            <default>0.0</default>
-        </key>
-        <!-- New keys -->
         <key name="cents" type="d">
-            <default>0.0</default>
-            <summary>Pitch shift in cents</summary>
-            <description>Amount to shift pitch in cents.</description>
+            <range min="-100" max="100" />
+            <default>0</default>
         </key>
-
+        <key name="semitones" type="d">
+            <range min="-12" max="12" />
+            <default>0</default>
+        </key>
         <key name="octaves" type="d">
-            <default>0.0</default>
-            <summary>Pitch shift in octaves</summary>
-            <description>Amount to shift pitch in octaves.</description>
+            <range min="-3" max="3" />
+            <default>0</default>
         </key>
-        <!-- End of new keys -->
-
         <key name="quick-seek" type="b">
             <default>false</default>
         </key>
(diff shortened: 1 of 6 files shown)


wwmm:
I do not see any problem in providing controls for octaves and cents. But I think it is better to keep the current window layout and just adding these controls in the same card that the semitones controls already are. The reasons are:

1. We were already moving plugins window style to the one currently used by the pitch plugin window. Going back to the old layout will not be productive
2. I have already ported the pitch plugin to the Qt branch and the layout used there is also based on two cards. One of the reasons is that the layout we used in the past (the one from your second image) does not place nice with small horizontal space at all. Handling this with cards is a lot easier.

Considering that the Qt branch will eventually replace our master branch adding these controls only here may not be the most effective thing to do. If I knew you were working on this I would have suggested to implementing this there. But as it is already done :-) my suggestion is to keep the current window style and just adding the controls there so we can merge this PR.

Digitalone1:
If the new parameters are integer, we should use `type="i"` in gsettings schema and `gsettings_get_int` to retrieve the value in the code.

But, I wonder if this is really necessary. If you're an advanced user and know the formula to calculate the semitones, you can just type the value in the current spinbutton. For most users is more straightforward to have a single parameter.

wwmm:
>But, I wonder if this is really necessary. If you're an advanced user and know the formula to calculate the semitones, you can just type the value in the current spinbutton. For most users is more straightforward to have a single parameter.

From a technical point of view it definitely isn't necessary. I see this as a convenience. And it might help people that are used to the rubberband library where all the 3 controls are usually provided.

wwmm:
I've updated our Qt branch with the same changes.

PR #3835, opened 2025-06-03
Improve Convolver UI

Digitalone1:
1. Avoid showing `nan` labels in the Graph when the impulse is unset. A small function has been added to check if the chart is a non-empty list. Unfortunately `nan` labels are still visible when Spectrum and LogScale buttons are toggled. I don't know how to avoid that, but at least the labels are not shown when the effect is loaded on an empty/unset impulse.
2. Show a message to the user when the impulse is unset rather than `""`.
3. Sanitize the combined output impulse filename like it has been done for new preset names.

Unfortunately I had issues in the last months with Arch Code OSS (I couldn't update the extensions anymore for an unknown issue), so I had to install Microsoft VSCode from AUR to keep using the latest extensions on other projects I need. I have to say it's not bad, even if it's partially closed source with telemetry. But the good news is that I can install official QT and QML extensions which I find better than the older ones. This will show some minor changes in the syntax format.

The extensions also shows lots of warnings for QML files:

Screenshot From 2025-06-03 16-52-26

### DIFF
--- a/src/contents/ui/Convolver.qml
+++ b/src/contents/ui/Convolver.qml
@@ -122,6 +122,10 @@ Kirigami.ScrollablePage {
                 id: combinedImpulseName
 
                 label: "Output Impulse Name"
+
+                validator: RegularExpressionValidator {
+                    regularExpression: /^[^\\/]{1,100}$/ //strings without `/` or `\` (max 100 chars)
+                }
             }
 
             Controls.ProgressBar {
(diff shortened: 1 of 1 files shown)


wwmm:
>But the good news is that I can install official QT and QML extensions which I find better than the older ones. This will show some minor changes in the syntax format.

Interesting. I think I saw somewhere in Arch's wiki a way to force the opensource vscode to use Microsoft market place. I will try to do that later.

wwmm:
In the end it was not necessary to install the hacks for the proprietary market place. Qt developers have finally added the extension to the open market. Now I have to figure out how to fix so many false positives. It is almost like it does not know where the system libraries are...

wwmm:
After adding `set(QT_QML_GENERATE_QMLLS_INI ON)` to the cmake file it now finds the packages from the Qt library. But the for the ones from Kirigami Addons it may be needed to add the additional paths to the workspace configuration.

wwmm:
>But the for the ones from Kirigami Addons it may be needed to add the additional paths to the workspace configuration.

Actually the ones from Kirigami are being found. The ones it can't are created by EasyEffects on runtime like `ee.database` or anything else starting with `ee.`. For these we will have to find a way to ignore the warning.

wwmm:
The unused import feature is useful but we have to be careful. There was one case where it said the import was not needed and the app crashed without it https://github.com/wwmm/easyeffects/blob/ac1fcecc22f23cef1e34e6ff1927263cafd7db91/src/contents/ui/PreferencesSheet.qml#L3. So testing is needed before actually removing the import.

Issue #3834, opened 2025-06-02
Implementation of XDG Global Shortcuts


Digitalone1:
Ticket to implement [XDG Global Shortcuts](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.GlobalShortcuts.html).

Digitalone1:
In the last days I tried to do this, but failed. I was curious to add this so the user has the opportunity to so something when the app UI is not visible while the service is running in the background. I thought to a shortcut to toggle the global bypass.

First I tried to follow some ChatGPT guidelines, but it was not working.

Then I tried to follow what it's done in the following two projects:
- https://git.dec05eba.com/gpu-screen-recorder-gtk/tree/src/global_shortcuts.c
- https://github.com/ghostty-org/ghostty/pull/7083

I think this is the right approach since they are working (at least for Ghostty they say it's working; I don't know for sure about GTK screen recorder).

The problem is that when I test it, I don't get the Gnome popup to bind the shortcut (which I saw months ago for Chromium that was asking me to bind a shortcut for an extension in the browser).

I fixed all the returning errors, but D-Bus calls get done right, but I'm surely missing something because I don't get popup to associate to bind the shortcut (and, as a consequence, I can't test the `Activated` signal).

Since the Qt port is in progress, this should be done there, but I am more confident on the GTK side, so I tried on the main branch. The code is here (just the latest commit): https://github.com/Digitalone1/easyeffects/tree/bypass_shortcut

@wwmm Since it's not working, I didn't want to make a PR, but would you like to take a look? Do you want me to make a PR anyway to test on your end?

wwmm:
>Since the Qt port is in progress, this should be done there, but I am more confident on the GTK side, so I tried on the main branch.

It makes sense. I do not see any problems in backporting the whole thing to the gtk branch once we figure out how to make it work. But I will probably do the first attempts on the Qt branch. 

I imagine that this will also require some way for the user to customize these shortcuts. it is one thing to have a fixed shortcut that is only used when the window is active. For global shortcuts conflicts with other apps have to be managed somehow.

>I fixed all the returning errors, but D-Bus calls get done right, but I'm surely missing something because I don't get popup to associate to bind the shortcut (and, as a consequence, I can't test the Activated signal).

Hum... Usually dbus features require a dbus file installed in the system. Or does this approach works entirely one the fly when the app starts running? I will try to take a look at this this week if possible.

[...]

PR #3838, opened 2025-06-04
GlobalShortcuts: fix handle tokens management

Digitalone1:
Related to #3834 

As it's stated [here](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Request.html), "the token that the caller provides should be unique and not guessable. To avoid clashes with calls made from unrelated libraries, it is a good idea to use a per-library prefix combined with a random number".

So, as far as I understood, the session handle should be a single token for the whole window/service session, but every D-Bus request should have a different (not guessable) token.

I added an util for generating random strings (appended to app name). Maybe it's too much since the recommendation says "random number", but it can be useful in the future, who knows...

@wwmm Since I can't test the `Activate` signal from command line under Gnome, please do it on your system.

**Edit:** I forgot it. Given that we want to add the toggle bypass global shortcut, I suppose we also want to add an internal one? Maybe with `CTRL+B`? As I made in my test: https://github.com/Digitalone1/easyeffects/blob/6f26e2af921ae62bbf59b9a843fd0bd31a5ade49/src/application.cpp#L814-L826

@wwmm What do you think?

### DIFF
--- a/src/global_shortcuts.cpp
+++ b/src/global_shortcuts.cpp
@@ -31,7 +31,6 @@
 #include <QDBusMetaType>
 #include <QDBusReply>
 #include <utility>
-#include "util.hpp"
 
 // Based on https://github.com/SourceReviver/qt_wayland_globalshortcut_via_portal/blob/main/wayland_shortcut.cpp
 
@@ -40,8 +39,8 @@ GlobalShortcuts::GlobalShortcuts(QObject* parent) : QObject(parent) {
   qDBusRegisterMetaType<QList<QPair<QString, QVariantMap>>>();
 
   QMap<QString, QVariant> options;
-  options["handle_token"] = handle_token;
-  options["session_handle_token"] = handle_token;
+  options["handle_token"] = QString("easyeffects%1").arg(util::random_string(32));
+  options["session_handle_token"] = session_handle_token;
 
   QList<QVariant> args_create_session;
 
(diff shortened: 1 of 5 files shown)


wwmm:
>What do you think?

It is fine to use this way to generate the unique token. Definitely an overkill. But it is fine. I just did another commit disabling the instantiation of the global shortcuts class. As things are right now we can't have this by default. Forcing KDE system settings to open whenever EasyEffects starts is not good. This feature will have to be stay in our `Experimental Features` settings for a while.

>I suppose we also want to add an internal one? Maybe with CTRL+B?

It is a good idea. The whole thing can be done entirely on QML as you can see here https://github.com/wwmm/easyeffects/blob/5cbc100c91ccd327ad709b6cd181ff8c719bed65/src/contents/ui/main.qml#L66. For the global shortcut it may be better to use the `Meta` keys in the combination. We will definitely have conflicts with other apps shortcuts if we use the common `Ctrl+` combinations. And even this way I think that providing a way for the user to change the global shortcuts keys is unavoidable. So I would only bother with selecting a combination after a shortcuts configuration window is implemented.

wwmm:
>Since I can't test the Activate signal from command line under Gnome, please do it on your system.

The activate signal is called without problems.

Digitalone1:
> I just did another commit disabling the instantiation of the global shortcuts class. As things are right now we can't have this by default. Forcing KDE system settings to open whenever EasyEffects starts is not good. This feature will have to be stay in our `Experimental Features` settings for a while.

Oh, I forgot this, you already said it in the issue. 

It seems like this one: https://discourse.gnome.org/t/issue-with-global-shortcuts-setup-popup-in-gnome-48/27946

It was fixed by Chromium. Maybe we should avoid to make the bind request if the shortcut is already bound (even if it's in the Experiment Feature).

wwmm:
>Maybe we should avoid to make the bind request if the shortcut is already bound (even if it's in the Experiment Feature).

That code example I've found shows how to list shortcuts. I do not know how far that search goes. If it is only the app shortcut or everything in the system. But we probably will have to use it and search for our own shortcuts before deciding if we have to make the binding or not.

wwmm:
Just in case it helps you to figure out things on your system this is what EasyEffects prints when the shortcut is activated

>Got GlobalShortcuts Activated Signal -> "/org/freedesktop/portal/desktop/session/1_93/easyeffectsaGBxpgDJjwF5V9fhfrctFCwaqffesPIZ" "toggle_global_bypass" 0 QMap()

Digitalone1:
> Just in case it helps you to figure out things on your system this is what EasyEffects prints when the shortcut is activated
> 
> > Got GlobalShortcuts Activated Signal -> "/org/freedesktop/portal/desktop/session/1_93/easyeffectsaGBxpgDJjwF5V9fhfrctFCwaqffesPIZ" "toggle_global_bypass" 0 QMap()

Okay, if I install the Qt branch using a custom AUR PKGBUILD, how can I look the debug messages since I can't launch it from the command line?

wwmm:
>Okay, if I install the Qt branch using a custom AUR PKGBUILD, how can I look the debug messages since I can't launch it from the command line?

As the qt branch is right now it always prints debug messages. I did not have time to hide this behind an environment variable like we have in the gtk branch. This means that your `journalctl` output will have everything easyeffects will print. Just run `sudo journalctl -fb` in a terminal to see things in realtime.

PR #3840, opened 2025-06-05
Improve Global Shortcuts handling and add internal Bypass toggling

Digitalone1:
1. Global shortcuts are now defined inside an array, so in the future it should be easy to add other shortcuts alongside description and title.
2. Added the experimental option to enable the Global Shortcuts (needs restart to apply).
3. Added the CTRL+B internal shortcut for Global Bypass toggling.

Unfortunately I can't test this further since I should install the dev build like a normal application which is not suitable to my workflow.

For the problem with BindShortcuts spawning the system settings every time, if it's enough to make it only one time and the shortcut remains correctly registered, you may add an internal bool settings to track the binding.

- The first time the user enables the experimental feature (and restart the app/service), the flag is false and the BindShortcuts call is executed.
- If everything goes right, the flag is changed to true and the next time the BindShortcuts call is skipped.
- If the user changes something and the Bind is lost, they can disable and reenable the Global Shortcuts experimental feature (upon disabling, the flag is turned off/false, and the next time the BindShortcuts call is executed).

This is only an idea and, if it works, it seems more straightforward than looking the Shortcut List. I can't test it, but you may give it a try.

### DIFF
--- a/src/contents/kcfg/easyeffects_db.kcfg
+++ b/src/contents/kcfg/easyeffects_db.kcfg
@@ -36,6 +36,10 @@
             <label>Last used output community package.</label>
             <default code="true">""</default>
         </entry>
+        <entry name="xdgGlobalShortcuts" type="Bool">
+            <label>Enable support for XDG Global Shortcuts</label>
+            <default>false</default>
+        </entry>
         <entry name="showNativePluginUi" type="Bool">
             <label>Show Plugin's Native Window</label>
             <default>false</default>
--- a/src/contents/kcfg/easyeffects_db.kcfg
+++ b/src/contents/kcfg/easyeffects_db.kcfg
@@ -129,7 +129,7 @@
     </group>
     <group name="EffectsPipelines">
         <entry name="bypass" type="Bool">
-            <label>Puts the Effects Pipelines in Passthrough Mode.</label>
+            <label>Global Bypass: Puts the Effects Pipelines in Passthrough Mode.</label>
             <default>false</default>
         </entry>
         <entry name="processAllOutputs" type="Bool">
--- a/src/contents/ui/PreferencesSheet.qml
+++ b/src/contents/ui/PreferencesSheet.qml
@@ -320,6 +320,18 @@ Kirigami.OverlaySheet {
         ColumnLayout {
             FormCard.FormCard {
                 EeSwitch {
+                    id: xdgGlobalShortcuts
+
+                    label: i18n("Global Shortcuts")
+                    subtitle: i18n("Enables support for XDG Global Shortcuts (Easy Effects service needs to be restarted)")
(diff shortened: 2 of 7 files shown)


wwmm:
>Added the experimental option to enable the Global Shortcuts (needs restart to apply).

I've found a way to make this work without the need to restart EasyEffects. It seems fine so far. 

>For the problem with BindShortcuts spawning the system settings every time, if it's enough to make it only one time and the shortcut remains correctly registered, you may add an internal bool settings to track the binding.

This is probably the easiest thing to do. I implemented it in my last commit. I've also added some initial code to change the initialization a little based on the desktop running. But based on the comments on the other issue gnome and hyprland seem to always want the binding call. So for now KDE is the only one  with special handling. It seems fine so far.

wwmm:
In the future the array `ee_global_shortcuts_array` should probably go to the database xml. The reason is that when the user becomes able to customize the global shortcuts the `preferred_trigger` information will need to be updated. The easiest way to handle this is probably moving the whole thing to the database.

wwmm:
>In the future the array ee_global_shortcuts_array should probably go to the database xml. 

I mean... We do not have to necessarily make an array. This may not be practical. I am thinking more about have the information somehow in the database.

Digitalone1:
> In the future the array `ee_global_shortcuts_array` should probably go to the database xml. The reason is that when the user becomes able to customize the global shortcuts the `preferred_trigger` information will need to be updated. The easiest way to handle this is probably moving the whole thing to the database.

Description and shortcut ID are constants, should we store them in the database too? Since only the preferred trigger is variable and could be set by the user, we could store only that one for every global shortcuts.

To update the key changed by the user, we should listen to Shortcut Changed signals, right?

Digitalone1:
Shortcuts Changed signals returns the shortcuts list, so maybe it's better to make map (or a QMap) because we should search the database key from the shortcut ID. With the array we should iterate the structures, while with the map we can access the database key by the shortcut ID.

wwmm:
>Description and shortcut ID are constants, should we store them in the database too? Since only the preferred trigger is variable and could be set by the user, we could store only that one for every global shortcuts.

From a technical point of view they do not need to be in the same place. I am just not sure if it will feel strange or not to have the id and the description in one place while the preferred trigger that has to be used with them is in a totally different location.

In any case something else that came to my mind now is that it will probably better to have these preferred triggers in an independent database xml file just for them.

>To update the key changed by the user, we should listen to Shortcut Changed signals, right?

I think it will depend on how the window that will provide this configuration will be implemented. I did not look at Qt or Kirigami docs yet to see if they provide any kind of facility for this.

Digitalone1:
Okay, in the meanwhile I updated the TODO list for Qt branch reporting these things.

PR #3844, opened 2025-06-08
Various UI improvements

Digitalone1:
As the title says.

I noticed in the Gate level meters that the padding is not natural on labels as the other widgets. It seems that `fillwidth: true` let the text collapse on the container border. Using Qt style center alignment in its place is better. Besides using the `uniformCellWidths` is not a good idea because translations can have different lengths showing an unnecessary empty space on a single side of the container.

@wwmm If you work on the port of the next plugins, can you please prioritize Multiband Gate and Equalizer? If these two are implemented, I could definitively switch to the Qt branch installing it in place of the current one. This could help me in discovering bugs.

Porting these two plugins, do you think it's ready to be used on a daily basis? Does presets autoload work?

### DIFF
--- a/src/contents/ui/Autogain.qml
+++ b/src/contents/ui/Autogain.qml
@@ -17,7 +17,7 @@ Kirigami.ScrollablePage {
 
     function updateMeters() {
         if (!pluginBackend)
-            return ;
+            return;
 
         momentary.value = pluginBackend.getMomentaryLevel();
         shortterm.value = pluginBackend.getShorttermLevel();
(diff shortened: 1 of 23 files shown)


wwmm:
>If you work on the port of the next plugins, can you please prioritize Multiband Gate and Equalizer?

My original plan was `deepfilter -> equalizer -> multiband compressor -> multiband gate`. These are the only ones that were not ported yet. But I can do the gate before the compressor. No problems.

>Porting these two plugins, do you think it's ready to be used on a daily basis? Does presets autoload work?

I think it is. The autoloading works. But its profiles have to be recreated because the Qt branch uses a different PipeWire node field for it. And for some reason PipeWire names the hardware profile in a slightly different way there.

PR #3866, opened 2025-06-27
Preliminary work for multiband plugins

Digitalone1:
In order to speed up the implementation of the last plugins, I made a preliminary work for the multiband plugins.

The commits implement:

1. multiband plugins database
2. a draft of the multiband plugins process (needs to add the controls for the bands)
3. a draft of the multiband plugins preset (needs to add the presets for the bands)

And the QML UI should be implemented from scratch for both mb gate and mb compressor. I don't have a clue on how to shape them, so it's better to leave that to you.

I hope this helps. Search for `TODO` comments in the projects. Feel free to merge it when it's more comfortable to you (maybe after the full equalizer implementation).

I'll check for conflicts in the next days, but there shouldn't be any I think.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -71,6 +71,7 @@ target_sources(easyeffects PRIVATE
     main.cpp
     maximizer.cpp
     maximizer_preset.cpp
+    multiband_compressor.cpp
     output_level.cpp
     pitch.cpp
     pitch_preset.cpp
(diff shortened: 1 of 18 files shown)


wwmm:
Thanks!  I think it is safe to merge this now.

Now that each plugin port is next to be finished I have been trying to do something that should have been done a long time ago https://github.com/wwmm/easyeffects/blob/eeqt/util/adding_new_plugins.txt. It is too easy to forget the steps required to add new plugins. We should document it somewhere.

PR #3867, opened 2025-06-27
Replace gain scale widgets with spinButtons

nbanman:
First ever PR on any project, so forgive me if I put the cart before the horse on anything, like if I'm supposed to get an Issue added first.

**The issue:**
Using scale widgets for gain adjustment is dangerous, in my experience. The gain range is from -30dB to +30dB, a 60dB swing. I personally use an EQ for my headphones based on Oratory1990's Harman adjustment for Sennheiser HD650s, which calls for a -9.3dB gain. 

But inadvertently clicking on the bar at any point instantly changes the gain to that level. More crucially, clicking on the label to the right of the scale (still part of the scale widget) instantly raises the gain the maximum +30dB setting. I clicked on the label hoping that it was editable, because using my mouse to edit the gain made it hard to enter -9.3dB exactly. It kept skipping over. When I clicked, I was hit with a +40dB blast of distorted noise.

**Solution**
For this reason, a spin button is better. Much finer control, the ability to enter in a value using keyboard values, and no risk of accidentally damaging hearing/equipment.

Two files are changed, fairly self-explanatory if you look at the changes below. I used AI to help with the .ui file, as I don't have experience with Gtk layout. Hopefully the changes are not too clumsy.

### DIFF
--- a/data/ui/equalizer.ui
+++ b/data/ui/equalizer.ui
@@ -390,11 +390,9 @@
                             <object class="GtkBox">
                                 <property name="hexpand">1</property>
                                 <property name="vexpand">0</property>
-                                <property name="homogeneous">1</property>
-                                <property name="spacing">6</property>
+                                <property name="spacing">12</property>
                                 <child>
                                     <object class="GtkBox">
-                                        <property name="hexpand">1</property>
                                         <property name="vexpand">0</property>
                                         <property name="spacing">6</property>
                                         <child>
@@ -406,8 +404,8 @@
                                         </child>
                                         <child>
                                             <object class="GtkSpinButton" id="input_gain">
-                                                <property name="hexpand">1</property>
                                                 <property name="valign">center</property>
+                                                <property name="width-chars">11</property>
                                                 <property name="adjustment">
                                                     <object class="GtkAdjustment">
                                                         <property name="lower">-36</property>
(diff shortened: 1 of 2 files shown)


wwmm:
What you are proposing is exactly what has already been done to the Qt branch
image
. And considering that we have been trying to keep the gtk4 branch in some kind of "maintenance mode" until it is replaced I think we should keep it as it is and just wait until the Qt branch is ready.

wwmm:
>What you are proposing is exactly what has already been done to the Qt branch

This happened there for totally different reasons. In the Qt branch I am trying to make the window more friendly to tile managers and similar situations where the available width becomes quite small. And as scales are only really good when there is plenty of horizontal or vertical space using a spinbutton was the obvious choice.

nbanman:
Totally understandable. I didn't even know there was a Qt branch! The nice thing about FOSS is that I can still use my version until the Qt version is released!

I do think this is a safety issue, but you're in a much better position than me to evaluate whether it's a common problem or just something that happened to me. 

In an ideal world, a spinButton isn't the best solution either, because my experience typing in values with Equalizer APO is that you can delete a decimal point and suddenly get blasted (luckily I've only done this with specific bands rather than the gain). The safest would be a slider that you can **only** drag, with no jumps, with plus/minus buttons on the side for 0.1dB adjustments. That being said that doesn't address the tiling manager layout issues you mentioned.

Amazing, amazing work, by the way, on this project.

Issue #3882, opened 2025-07-05
Exporting EQ preset with "Resonance" band crashes


amphyvi:
### EasyEffects Version

7.2.3

### What package are you using?

Flatpak (Flathub)

### Distribution

Ubuntu 25.04

### Describe the bug

Hi, I'm getting some strong crystalline sharpness somewhere in the 12kHz range with EasyEffects enabled. So, I figured I'd just EQ it down, especially since I can set a band to "Resonance".

But, if I press the "APO" button under "Export Preset" and attempt to save it with any bands with its type set to "Resonance", the following happens all at once:
- Easy Effects crashes.
- All equalizer effects halt.
- An empty file is generated with the name I specified when exporting.

Feels as if Easy Effects generates the file successfully, but can't figure out how to export with a band set to "Resonance". This is what the file looks like:
```Preamp: -6 db
Filter 1: ON LSC Fc 20 Hz Gain 5.7 dB Q 0.6666667
Filter 2: ON PK Fc 96 Hz Gain 2.5 dB Q 1.41
Filter 3: ON PK Fc 234 Hz Gain -1.7 dB Q 1.27
Filter 4: ON PK Fc 1148 Hz Gain -2.4 dB Q 1.15
Filter 5: ON PK Fc 1697 Hz Gain 1.4 dB Q 1.22
Filter 6: ON PK Fc 2348 Hz Gain 2.5 dB Q 3.3
Filter 7: ON PK Fc 5815 Hz Gain 0.6 dB Q 2.92
Filter 8: ON PK Fc 6000 Hz Gain 0.5 dB Q 0.56
Filter 9: ON PK Fc 7679 Hz Gain 1.4 dB Q 3.04
Filter 10: ON PK Fc 9514 Hz Gain 1.7 dB Q 3.52
Filter 11: ON PK Fc 11516 Hz Gain 0.8 dB Q 1.2
Filter 12: ON PK Fc 12000 Hz Gain -2.85 dB Q 4.36
Filter 13: ON HSC Fc 19828 Hz Gain -5.8 dB Q 0.6666667
```
[...]

PR #3885, opened 2025-07-07
APO export: handle unsupported LSP EQ filters

Digitalone1:
Related to #3882 

Can we make this as the last release before the Qt switch?


wwmm:
>Can we make this as the last release before the Qt switch?

We can try to make it the last gtk4 release. Let's hope nothing major happens before the switch.

PR #3886, opened 2025-07-07
Add JSDoc + small fixes

Digitalone1:
Added the JSDoc syntax to Javascript files to keep them documented.

In the past I made the drafts for the APO/GraphicEQ import, but for some reasons I didn't make the ones for the export. I will make them in the next weeks if I have time.

Regarding the import, I suppose it's just a matter of passing the APO file content to the Javascript function from the QML code. Then the Javascript function returns the imported data and we have to update the widgets from the QML code, right?

So the difference from Libadwaita is that we imported the data from the C++ code and updated the gsettings keys, while now we import it from the QML code and we update the UI. Am I correct?

Regarding the export instead, can we write a file from the QML code?

### DIFF
--- a/src/contents/ui/Common.js
+++ b/src/contents/ui/Common.js
@@ -1,6 +1,13 @@
+/** @type {number} */
 var minimumLinearLevel = 0.00001;
+/** @type {number} */
 var minimumDecibelLevel = -100.0;
 
+/**
+ * Check if a variable of any type is empty.
+ * @param {*} v 
+ * @returns {boolean}
+ */
 function isEmpty(v) {
     switch (typeof v) {
         case "string":
@@ -16,6 +23,12 @@ function isEmpty(v) {
     }
 }
 
+/**
+ * Check if two given arrays are equal.
+ * @param {Array} a 
+ * @param {Array} b 
+ * @returns {boolean}
+ */
 function equalArrays(a, b) {
     if (a.length !== b.length) {
         return false;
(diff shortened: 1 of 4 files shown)


wwmm:
> Regarding the import, I suppose it's just a matter of passing the APO file content to the Javascript function from the QML code. Then the Javascript function returns the imported data and we have to update the widgets from the QML code, right?

It could be done this way. It seems totally doable. But considering that we already have a lot of code implemented in the file [equalizer_ui.cpp](https://github.com/wwmm/easyeffects/blob/master/src/equalizer_ui.cpp) it may be easier to just turn some of those functions into `Q_INVOKABLE` methods that could be called from the QML side. For example the function https://github.com/wwmm/easyeffects/blob/e21e2313b976457af91471469c396401ea9a1e40/src/equalizer_ui.cpp#L372 could become a Q_INVOKABLE. We just have to adapt its arguments a little.

 I did something similar to the convolver. The file dialog stuff can go entirely to qml https://github.com/wwmm/easyeffects/blob/048dbd4ede50c1649754d959768a3b45060102a7/src/contents/ui/ConvolverImpulseSheet.qml#L31. Inside it we can call a Q_INVOKABLE https://github.com/wwmm/easyeffects/blob/048dbd4ede50c1649754d959768a3b45060102a7/src/contents/ui/ConvolverImpulseSheet.qml#L38 that is defined this way https://github.com/wwmm/easyeffects/blob/048dbd4ede50c1649754d959768a3b45060102a7/src/presets_manager.cpp#L847

Doing this we can more or less reuse all the code that the current equalizer has. What is needed is replacing the gsettings calls by the appropriate database calls we have now. In the particular case of the equalizer just the bands loops have become a little annoying to do because I have been avoiding to use generic methods like `setProperty("key_name", value)` and `property("key_name")`

wwmm:
>In the particular case of the equalizer just the bands loops have become a little annoying to do because I have been avoiding to use generic methods like setProperty("key_name", value) and property("key_name")

That being said when it was obvious that doing the job with macros would be ridiculously annoying https://github.com/wwmm/easyeffects/blob/bd7e7b382a502a2f9966ad845839519537368c3e/src/equalizer.cpp#L325 I just used the generic getters and setters and moved on with my life :smile:

wwmm:
Considering how big `equalizer.cpp` already is instead of copying to it the necessary functions from `equalizer_ui.cpp` it may be better to keep them in some kind of `equalizer_ui_helpers.cpp` or a similar name and including the corresponding header inside `equalizer.cpp`

wwmm:
> Considering how big `equalizer.cpp` already is instead of copying to it the necessary functions from `equalizer_ui.cpp` it may be better to keep them in some kind of `equalizer_ui_helpers.cpp` or a similar name and including the corresponding header inside `equalizer.cpp`

And then these functions would be wrapped by the Q_INVOKABLE methods defined in the `equalizer.cpp` class.

Digitalone1:
> It could be done this way. It seems totally doable. But considering that we already have a lot of code implemented in the file [equalizer_ui.cpp](https://github.com/wwmm/easyeffects/blob/master/src/equalizer_ui.cpp) it may be easier to just turn some of those functions into `Q_INVOKABLE` methods that could be called from the QML side. For example the function.

Yes, I agree. I don't like the fact that I spent so much time porting the code to Javascript, but I think it's still better lo use the existing C++ methods. 😅

If we use the QML/Javascript code, it's almost the same, but the C++ code is more useful because we would still be able to import/export the APO presets when the UI is not visible (maybe implementing a command line option).

So I think we can remove the EqualizerUI.js file.

PR #3934, opened 2025-07-23
Equalizer: small UI improvements

Digitalone1:
I added a minimum height to the gain slider since it's too little when the window height is highly reduced. But we need to attach a vertical scrollbar when the equalizer bands are longer than the container. I tried to add it, but failed.

Unfortunately, if you reduce the window width, the horizontal scrollbar overlaps the gain labels. Can we do something to avoid it?

Regarding the band gain, I added the decimals to the label and the `0.01` step size to the slider (as in Libadwaita), but I suppose we need also the page up/down control as we already have for the spinboxes, right? At the moment, pressing page up/down doesn't change the value.

Another thing. When I compiled the first time, I got the following warning:

```
easyeffects/src/convolver.cpp:645:44: warning: 
‘_LIBCPP_VERSION’ is not defined, evaluates to ‘0’ [-Wundef]
  645 | #if defined(ENABLE_LIBCPP_WORKAROUNDS) || (_LIBCPP_VERSION < 170000 || defined(_LIBCPP_HAS_NO_INCOMPLETE_PSTL))

```

### DIFF
--- a/src/contents/ui/EqualizerBand.qml
+++ b/src/contents/ui/EqualizerBand.qml
@@ -48,7 +48,7 @@ Controls.ItemDelegate {
                 if (f < 1000) {
                     return toLocaleLabel(f, 0, "Hz");
                 } else {
-                    toLocaleLabel(f * 0.001, 1, "kHz");
+                    return toLocaleLabel(f * 0.001, 1, "kHz");
                 }
             }
             enabled: false
--- a/src/contents/ui/EqualizerBand.qml
+++ b/src/contents/ui/EqualizerBand.qml
@@ -73,7 +73,7 @@ Controls.ItemDelegate {
             from: delegate.bandDB.getMinValue(bandName)
             to: delegate.bandDB.getMaxValue(bandName)
             value: delegate.bandDB[bandName]
-            stepSize: 1
+            stepSize: 0.01
             enabled: true
             onMoved: {
                 if (value != delegate.bandDB[bandName])
@@ -83,7 +83,7 @@ Controls.ItemDelegate {
 
         Controls.Label {
             Layout.alignment: Qt.AlignHCenter
-            text: Number(gainSlider.value).toLocaleString(Qt.locale(), 'f', 0)
+            text: Number(gainSlider.value).toLocaleString(Qt.locale(), 'f', 2)
             enabled: false
         }
     }
--- a/src/contents/ui/EqualizerBand.qml
+++ b/src/contents/ui/EqualizerBand.qml
@@ -1,6 +1,7 @@
 import QtQuick
 import QtQuick.Layouts
 import QtQuick.Controls as Controls
+import org.kde.kirigami as Kirigami
 
 Controls.ItemDelegate {
     id: delegate
@@ -67,6 +68,7 @@ Controls.ItemDelegate {
 
             Layout.alignment: Qt.AlignHCenter
             Layout.fillHeight: true
+            Layout.minimumHeight: Kirigami.Units.largeSpacing * 10
             orientation: Qt.Vertical
             from: delegate.bandDB.getMinValue(bandName)
             to: delegate.bandDB.getMaxValue(bandName)


wwmm:
>I added a minimum height to the gain slider since it's too little when the window height is highly reduced. But we need to attach a vertical scrollbar when the equalizer bands are longer than the container. I tried to add it, but failed.

There is already one added by the Kirigami widget there. But as this image shows
image
the minimum height may have to be applied to the card instead of the scales. I will do some tests now.

>Unfortunately, if you reduce the window width, the horizontal scrollbar overlaps the gain labels. Can we do something to avoid it?

I've noticed that. The crystalizer plugin needs the same fix. In the case of Presets Menu a custom vertical scrollbar fixed the problem https://github.com/wwmm/easyeffects/blob/e5f776d6427f4948e5ba49a4ff5b0185c039c290/src/contents/ui/PresetsLocalPage.qml#L188. Maybe something similar can be done here.

>but I suppose we need also the page up/down control as we already have for the spinboxes, right? 

Yes. As we do not have as many scales as we have spinbutton creating a custom scale widget may not be necessary. Handling the `Keys.onPressed` event locally should be enough.

>easyeffects/src/convolver.cpp:645:44: warning: 
‘_LIBCPP_VERSION’ is not defined, evaluates to ‘0’ [-Wundef]
  645 | #if defined(ENABLE_LIBCPP_WORKAROUNDS) || (_LIBCPP_VERSION < 170000 || defined(_LIBCPP_HAS_NO_INCOMPLETE_PSTL))

Hum... Now that I've executed `ninja clean` I see it too. Although it is a harmless warning caused by compiling with gcc instead of clang we should improve this workaround somehow. I imagine that the current master branch has the same warning.

wwmm:
I've moved the minimum height to the card layout
image

wwmm:
Inserting a custom scrollbar to the card `footer` is the way to go
image

wwmm:
>Inserting a custom scrollbar to the card footer is the way to go

But that removed the actions I've set to the card :smile: . I will try to move them to the header.

wwmm:
Fixed. Now I have to find a way to deal with the band menu being cut when there is not enough space to at the right side
image

wwmm:
>Now I have to find a way to deal with the band menu being cut when there is not enough space to at the right side

I think it may be better to use an overlaysheet like it was done to the shortcuts window. Now that I am testing the current band menu with reduced vertical and horizontal spaces I see its usability with the scrolls from the images above isn't the best.

Digitalone1:
It seems OK to me.

Screenshot From 2025-07-23 19-39-24

Only two things:

1. I'd like to have a bottom light gray border on the spectrum container (below the frequencies axis), the same we have on the top bar (where there is the Output/Input/PipeWire switcher). It seems like the spectrum is not properly "split" from the rest of the UI.
2. Is it normal you have the Wayland icon in the title bar? Shouldn't be the EasyEffects icon there?

wwmm:
>I'd like to have a bottom light gray border on the spectrum container (below the frequencies axis), the same we have on the top bar (where there is the Output/Input/PipeWire switcher). It seems like the spectrum is not properly "split" from the rest of the UI.

Although it looks fine when using a light theme for some reason on a dark theme it does feel like a separation is missing. We will probably have to add a border or something similar there.

>Is it normal you have the Wayland icon in the title bar? Shouldn't be the EasyEffects icon there?

No. It should be EasyEffects icon there. But for some reason KDE is not doing it. I wonder if the fact I am running from a local folder is the reason for that. Maybe KDE only gets the right icon if the app instance comes from a system installation.

wwmm:
I've added a separator to the spectrum header. It makes the visibility better. But there is a catch. Its effectiveness depends on the combination of colors used on the spectrum and on the app. For example dark theme on the app and on the spectrum is fine. I can see the separator. But if the app theme is dark and the user selects a light spectrum theme then the previous situation comes back.

Honestly this may be something that already happens in the gtk branch. I have never paid attention to what happens with different theme combinations. In any case at least the usual combinations seem fine.

Digitalone1:
I gave a look at it. It's okay to me, I don't see the need to use a different theme for the spectrum, but if a light chart is used on a dark theme, I think the split is just visible from the color. Anyway I like the new bottom border on the automatic theme.

PR #3941, opened 2025-07-25
Mark constants where possible

Digitalone1:
I started to do this prior to your last changes and I had to resolve a conflict. You can see from the changes that just the constants have been added.

Anyway I take this opportunity to highlight that there's a little issue on the GraphicEQ import. Fox example, if you have a 10 bands settings and import a 10+ bands GraphicEQ preset, the "band number" is not updated, so only the first 10 bands are loaded.

I don't remember how we do in the libadwaita branch and now I don't have the time to look at it, but I suppose the need to update the band number alongside all the other controls.

 You can test with the following custom preset which has more than 32 bands (and only the first 32 should be loaded).
[GraphicEq_32+bands.txt](https://github.com/user-attachments/files/21436800/GraphicEq_32%2Bbands.txt)


wwmm:
>Anyway I take this opportunity to highlight that there's a little issue on the GraphicEQ import. Fox example, if you have a 10 bands settings and import a 10+ bands GraphicEQ preset, the "band number" is not updated, so only the first 10 bands are loaded.

I did not have a file to test this one. So I did not see the problem. The import seems to be failing for some reason.

wwmm:
There was a mistake in the qml file. The wrong filedialog object was being used. It seems fine now.

Digitalone1:
> There was a mistake in the qml file. The wrong filedialog object was being used. It seems fine now.

I tested it. It's working.

PR #3942, opened 2025-07-28
Improve Kirigami inline messages

Digitalone1:
For the status messages, I used the same format as we use in other labels.

Then I wanted to do something to autohide the message after the preset was imported and with the help of ChatGPT I managed to add a Timer. It was pretty easy, but then I realized that it's better if we autohide only the positive statuses, while the failed ones remains visible so the user can have more time to read them.

To highlight successful and failed actions, I used the proper Kirigami message types. You can test them using a correct APO preset or an empty text file.

Do you like it? I made this only for the Equalizer, but I can replicate it for RNNoise if you want.

### DIFF
--- a/src/contents/ui/Equalizer.qml
+++ b/src/contents/ui/Equalizer.qml
@@ -55,9 +55,9 @@ Kirigami.ScrollablePage {
         nameFilters: [i18n("APO Presets") + " (*.txt)"]
         onAccepted: {
             if (pluginBackend.importApoPreset(apoFileDialog.selectedFiles) === true)
-                showStatus(i18n("Preset file imported!"));
+                showStatus(i18n("Preset File Imported."));
             else
-                showStatus(i18n("Failed to import the APO preset file!"));
+                showStatus(i18n("Failed to Import the APO Preset File."));
         }
     }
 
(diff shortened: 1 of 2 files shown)


wwmm:
>Do you like it? I made this only for the Equalizer, but I can replicate it for RNNoise if you want.

It seems good to me. I think that as time goes by we should do something similar to other plugins in the cases where having a feedback makes sense. Having that in mind I think it will be better to have the interval values as a key somewhere in the database so we do not have it hardcoded. At least for now I do not think it is necessary to expose a configuration for it in the preferences menu.

Digitalone1:
> Having that in mind I think it will be better to have the interval values as a key somewhere in the database so we do not have it hardcoded. At least for now I do not think it is necessary to expose a configuration for it in the preferences menu.

Which interval values do you mean?

wwmm:
> Which interval values do you mean?

the one from the timer 

>interval: 5000

Digitalone1:
> the one from the timer

Oh, now I get it. Where should I put it? In `src/contents/kcfg/easyeffects_db.kcfg`?

wwmm:
> Oh, now I get it. Where should I put it? In `src/contents/kcfg/easyeffects_db.kcfg`?

I think it is ok. Creating a new file just for this does not seem necessary.

PR #3943, opened 2025-07-29
Add systemd user service file for EasyEffects

m1st0:
Provides easyeffects.service to run EasyEffects as a user service. Includes instructions for manual setup in the user systemd directory.

### DIFF
--- a/debian/easyeffects/systemd/easyeffects.service
+++ b/debian/easyeffects/systemd/easyeffects.service
@@ -1,0 +1,19 @@
+# You may symlink this file in $HOME/.config/systemd/user/easyeffects.service
+#ln -s /path/to/easyeffects/debian/easyeffects/easyeffects.service ~/.config/systemd/user/easyeffects.service
+# Start it thereafter with
+#systemctl --user daemon-reload
+#systemctl --user enable --now easyeffects.service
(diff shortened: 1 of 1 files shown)


wwmm:
This should be a good example for people trying to use systemd to launch EasyEffects. Thanks!

m1st0:
Do you want me to update the wiki or the READMD.md further for a Debian/Ubuntu title or a systemd title?  The file is sort of hidden away in its location.  Unsure if you want me to look into the build issues with flatpack or archlinux which should not have any effect for a text file being present, so its likely something else there.

wwmm:
> Do you want me to update the wiki or the READMD.md further for a Debian/Ubuntu title or a systemd title? The file is sort of hidden away in its location. Unsure if you want me to look into the build issues with flatpack or archlinux which should not have any effect for a text file being present, so its likely something else there.

It is fine to add these instructions to our wiki. I think at least for now it is not worth the effort of going beyond that. It may even give people the wrong idea that we are integrating systemd in EasyEffects workflow. But in most cases an autostart desktop file is enough and some of our users probably are in distributions that do not use systemd. So just providing a good example for people that want to use it for some reason is enough.

PR #3949, opened 2025-08-05
Improve Kirigami inline messages

Digitalone1:
Implemented what we stated in #3942

I noticed that the inline message for the RNNoise plugin was showing the wrong style. That was because the returning value of `removeRNNoiseModel` was checked against a bool value, but the C++ function was a void type. I changed it.

I made the same for the Convolver, but I noticed something weird. When the inline message is shown, the overlaysheet automatically reduces its height:

Screenshot From 2025-08-05 14-25-40

Screenshot From 2025-08-05 14-26-00

Is this normal?

### DIFF
--- a/src/contents/ui/ConvolverImpulseSheet.qml
+++ b/src/contents/ui/ConvolverImpulseSheet.qml
@@ -13,8 +13,16 @@ Kirigami.OverlaySheet {
 
     required property var pluginDB
 
-    function showImpulseMenuStatus(label) {
+    function showImpulseMenuStatus(label, positive = true) {
         status.text = label;
+
+        if (positive) {
+            status.type = Kirigami.MessageType.Positive;
+            autoHideStatusTimer.start();
+        } else {
+            status.type = Kirigami.MessageType.Error;
+        }
+
         status.visible = true;
     }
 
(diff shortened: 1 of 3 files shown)


wwmm:
>Is this normal?

Although I expected a recalculation of the sheet height it seems to be changing too much. My guess is that the line https://github.com/wwmm/easyeffects/blob/c98c95283fcb40c188c4245430db826a85be0653/src/contents/ui/ConvolverImpulseSheet.qml#L34 has to be improved somehow.

wwmm:
I am doing some tests and I think I understood what is happening. As soon as `Kirigami.InlineMessage` is closed both its `height` and `implicitHeight` properties are set to zero. What triggers a recalculation of the sheet height. Hum...

wwmm:
One possible workaround for this is pretending that the footer height is equal to the header height and use that for the calculation instead `implicitHeight: control.parent.height - control.y - 2 * control.header.height`. I think it is fine. I will update the code with this approach.

Digitalone1:
> I am doing some tests and I think I understood what is happening. As soon as `Kirigami.InlineMessage` is closed both its `height` and `implicitHeight` properties are set to zero. What triggers a recalculation of the sheet height. Hum...

I think it's simply that you are counting the footer, but the footer height is changing when the inline message is shown? Indeed I removed both header and footer calculation and it works correctly.

wwmm:
>I think it's simply that you are counting the footer, but the footer height is changing when the inline message is shown? 

Yes. Usually a widget has at least its implicit height set to a non zero value even if hidden. For some reason this is not the case for the inline message.

PR #3950, opened 2025-08-05
Convolver impulse sheet: try to fix the height when inline messages are visible

Digitalone1:
As I said in #3949, the height doesn't change when you ignore the footer. I made a simpler calculation removing also the header.

If you accept it, you could even add a `Math.max(...)` to enforce a minimum height...


wwmm:
I think it is fine to do that in this case. At some point we have to decide what to do about those sheet heights. By default Kirigami tries to use the whole available height. The question is if we want a super tall menu even when there is only one or two entries in the list. In other words we have to decide what is the best visual style in these situations.

Digitalone1:
> I think it is fine to do that in this case. At some point we have to decide what to do about those sheet heights. By default Kirigami tries to use the whole available height. The question is if we want a super tall menu even when there is only one or two entries in the list. In other words we have to decide what is the best visual style in these situations.

You can make some tests with `Math.min` and `Math.Max` if the overlay becomes too long/small. There's also the `clamp` function in Javascript utils.

PR #3969, opened 2025-08-14
.

Obsessee:
.


wwmm:
It seems fine. So I merged it. But have in mind that as our Qt branch is getting close to the point where it can become the master branch I have been avoiding doing big changes to our current gtk4 code. Ideally it is better to implement new functionality like this one in the Qt branch.

PR #3982, opened 2025-08-18
Improve Common.toLocaleLabel

Digitalone1:
Improved the JSDoc for `toLocaleLabel()` common function. It actually returns a string and has additional optional parameters.

It's better to hint that the function expects a number or a string as the mandatory first parameter, but it can also fail to convert it to locale format, so I managed that case returning an empty string.

@wwmm At the moment the Multiband Gate fails in two situations to convert the value. The error log tells that it takes an undefined value. Please check it on your system.   

I saw also that every time we check the `NaN` value, we do it on numbers. In that case, it's slightly better to use the `Number.NaN()` static method. That's because it checks `NaN` directly, while global `isNaN()` tries to convert the value to number or, in the worst case, it makes an additional conversion to number type. See [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN#difference_between_number.isnan_and_global_isnan).

### DIFF
--- a/src/contents/ui/Common.js
+++ b/src/contents/ui/Common.js
@@ -13,7 +13,7 @@ function isEmpty(v) {
         case "string":
             return v.length === 0;
         case "number":
-            return isNaN(v);
+            return Number.isNaN(v);
         case "undefined":
             return true;
         case "object":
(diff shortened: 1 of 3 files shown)


wwmm:
> At the moment the Multiband Gate fails in two situations to convert the value. The error log tells that it takes an undefined value. Please check it on your system.

I've updated the Qt branch with a fix for this. Initializing the `list<real>` properties was the solution. Without that only when audio is being processed by the plugin those arrays are initialized.

PR #4018, opened 2025-08-29
Update QT Flatpak build

vchernin:
Seems to work ok. I could not get easyeffects to build with the older pipewire library 0.3.69, but it is probably best to finally use a newer version in flatpak with this migration.

I did see a warning:
```
CMake Warning:
  Manually-specified variables were not used by the project:

    DEVEL
    buildtype
```

Which I did not try investigating as I am not very familiar with cmake.

Also, the build succeeds even if `org.kde.kirigamiaddons.formcard` is not installed, but it will fail at runtime:

```
QQmlApplicationEngine failed to load component
qrc:/ui/main.qml:109:5: Type PreferencesSheet unavailable
qrc:/ui/PreferencesSheet.qml:6:1: module "org.kde.kirigamiaddons.formcard" is not installed
```

I added the dependency to fix this, but is there not some way to check this at build time?

### DIFF
--- a/util/flatpak/easyeffects-modules.json
+++ b/util/flatpak/easyeffects-modules.json
@@ -339,6 +339,24 @@
             "build-commands": [
                 "install -m 644 libdeep_filter_ladspa.so $FLATPAK_DEST/lib/ladspa/"
             ]
+        },
+        {
+            "name": "kirigami-addons",
+            "buildsystem": "cmake-ninja",
+            "builddir": true,
+            "sources": [
+                {
+                    "type": "archive",
+                    "url": "https://download.kde.org/stable/kirigami-addons/kirigami-addons-1.9.0.tar.xz",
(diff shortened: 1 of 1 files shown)


wwmm:
Thanks!

>I added the dependency to fix this, but is there not some way to check this at build time?

There is. I totally forgot to add `KirigamiAddons` to the findPackage command. I will fix this today.

>I did see a warning:

I am not sure what to do about the old `Devel` option. One of its uses where to append `.Devel` to the app id. But as the translations for desktop files is handled differently  in KDE apps I did not see yet how to keep that devel option working. I think I will just remove it for now. In the future we can take a look at this again.

wwmm:
And as I am attempting to handle the whole autostart logic with libportal the libportal compilation option also does not make sense anymore. I will remove it.

PR #4019, opened 2025-08-29
Implement a database check and make MB meters always visible

Digitalone1:
- I wanted to discover database keys which had defaults out of range. That's almost impossible to find them manually, so I made a script with the help of ChatGPT. I fixed the few that were wrong. The bot says also that they can make a YAML to integrate in GitHub Actions if you want.
- Multiband plugins meters are now always visible. I moved them into to footer below the stackviews.
- Having done that, I was seeing the locale conversion warnings for all the meters. I don't know why they were showing again even if you initialized the arrays. Anyway, in order to hide them, I had to add the nullish coalescing on meter values before passing to the locale conversion function.
- I have left the meters as they were before. If you don't like them, you can change (at the moment they are restricted if there are placed in 3 columns, otherwise they are expanded to fill all the space).

### DIFF
--- a/src/contents/ui/MultibandCompressor.qml
+++ b/src/contents/ui/MultibandCompressor.qml
@@ -543,196 +543,6 @@ Kirigami.ScrollablePage {
                     }
                 }
             }
-
-            Kirigami.CardsLayout {
-                maximumColumns: 3
-                uniformCellWidths: true
-
-                anchors {
-                    left: parent.left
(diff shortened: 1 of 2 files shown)


wwmm:
> The bot says also that they can make a YAML to integrate in GitHub Actions if you want.

I'm ok with it. It won't cause any harm to have one more test  in our github actions.

PR #4020, opened 2025-08-29
Fix CI builds

vchernin:
A bunch of fixes for CI.

- Replace the broken and commented out alpine and chimera linux ci builds with a fedora rawhide build for gcc and clang. 
Both alpine and chimera linux lack qt6-graphs which we now require. Hopefully they will be added soon, and then we can add chimera linux back to ci. 
I don't think alpine is very necessary when we have fedora rawhide, the packages in rawhide should be similarly up to date compared to alpine.

- One more significant feature is github actions now supports aarch64 runners, which are free for public repos. It was very easy to add it to the flatpak build. If there are issues with it we can remove it. At least from my flathub packaging experience the main troublemaker I can remember is rnnoise, which still has not tagged some fixes for their aarch64 build. The aarch64 build is a few minutes slower than x86_64 it seems, but with caching it should be reasonable.

- @wwmm one new thing you will probably like, github overhauled the way actions artifact works last year. https://github.com/actions/toolkit/tree/main/packages/artifact#v2---whats-new. One change I see now that I actually read the notes is we can access artifacts as soon as they are uploaded. This means we could have the bot comment in prs show the link for the Arch linux build as soon as it is done. It would not have to wait for the other jobs. I am not sure how easy the api it is, but I will definitely look at it when I have time.

- I improved the robustness of the new kcfg action. I also moved it to the existing CI file, which is more of an opinionated change but I think it is better. 

- I also did a few other random bits of cleanup like the runner ubuntu update and other action updates from the gtk4-legacy branch which should not have any impact.


wwmm:
>one new thing you will probably like, github overhauled the way actions artifact works last year. https://github.com/actions/toolkit/tree/main/packages/artifact#v2---whats-new. One change I see now that I actually read the notes is we can access artifacts as soon as they are uploaded. This means we could have the bot comment in prs show the link for the Arch linux build as soon as it is done. It would not ha

Interesting. I would be nice indeed.

PR #4021, opened 2025-08-29
Various UI improvements

Digitalone1:
Doing these changes I discovered some issues:

1. The app logo in the README is broken.
2. The PipeWire node state is not provided anymore. Indeed you can see from your screenshot that the final row in the app info card is starting with a middle dot. I can reproduce it also on my system. Don't know why this is happening, but I modified the string to not start with a leading middle dot when the state is missing.
3.  In presets menu, when an inline message is shown (i.e. when a preset is created/removed), the frame height changes. This is the same issue happening in the convolver menu. We fixed it not counting the footer/header in the implicit height: https://github.com/wwmm/easyeffects/pull/3950/commits/7293c9356a90b16ac034bef9a1fd7ad617efb5a2
4. I didn't fix the issue above because there's another problem to solve in the plugin menu: when the autoloading tab is visible, the frame footer collapses on the bottom of the main window. I don't know how to fix this one. Please @wwmm, take a look at both issues.
5. When I was testing the inline messages in the plugin menu, I could switch between good presets and corrupted ones (that contains errors in the JSON format). But I found out another issue: when loading an empty corrupted preset (a text file without characters), I'm not able to load another preset. The pipeline is stuck, this is what happens:
Screenshot From 2025-08-29 22-57-24

### DIFF
--- a/src/contents/ui/DelegateStreamsList.qml
+++ b/src/contents/ui/DelegateStreamsList.qml
@@ -74,7 +74,7 @@ Kirigami.AbstractCard {
                 RowLayout {
                     Controls.Label {
                         wrapMode: Text.WordWrap
-                        text: state + " · " + format + " · " + rate + " · " + nVolumeChannels + i18n(" channels") + " · " + latency
+                        text: (state === "" ? "" : `${state} · `) + `${format} · ${rate} · ${nVolumeChannels} ` + i18n("channels") + ` · ${latency}`
                         color: Kirigami.Theme.disabledTextColor
                     }
                 }
--- a/src/contents/ui/MenuAddPlugins.qml
+++ b/src/contents/ui/MenuAddPlugins.qml
@@ -125,7 +125,7 @@ Kirigami.OverlaySheet {
                             plugins.push(new_name);
                         }
                         streamDB.plugins = plugins;
-                        showMenuStatus(i18n("Added Plugin") + ": " + translatedName);
+                        showMenuStatus(i18n("Added Effect") + `: <strong>${translatedName}</strong>`);
                     }
                 }
             }
--- a/src/contents/ui/PreferencesSheet.qml
+++ b/src/contents/ui/PreferencesSheet.qml
@@ -395,8 +395,8 @@ Kirigami.OverlaySheet {
                 EeSwitch {
                     id: showNativePluginUi
 
-                    label: i18n("Native Plugin Window")
-                    subtitle: i18n("Allows The Native Plugin Window to be Shown/Hidden")
+                    label: i18n("Native Window of Effects")
+                    subtitle: i18n("Allows the Native User Interface of Effects to be Shown/Hidden")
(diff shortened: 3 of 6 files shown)


wwmm:
>The app logo in the README is broken.

I forgot that the icon is in a different folder now. I will fix it.

>The PipeWire node state is not provided anymore. Indeed you can see from your screenshot that the final row in the app info card is starting with a middle dot. I can reproduce it also on my system. Don't know why this is happening, but I modified the string to not start with a leading middle dot when the state is missing.

Really strange. There should be a non empty string there. I will investigate

>I don't know how to fix this one. Please @wwmm, take a look at both issues.

Ok. I will take a look at it.

>When I was testing the inline messages in the plugin menu, I could switch between good presets and corrupted ones (that contains errors in the JSON format). But I found out another issue: when loading an empty corrupted preset (a text file without characters), I'm not able to load another preset. The pipeline is stuck, this is what happens:

When you have time send me this file so I can see what happens on my computer.

Digitalone1:
> When you have time send me this file so I can see what happens on my computer.

It's just an empty text file.

wwmm:
>The PipeWire node state is not provided anymore. Indeed you can see from your screenshot that the final row in the app info card is starting with a middle dot. I can reproduce it also on my system. Don't know why this is happening, but I modified the string to not start with a leading middle dot when the state is missing.

Fixed. I guess we will have to take more seriously those qml linter warnings about `qualify the item with its parent id`. After I put an id to `Kirigami.AbstractCard` and used `root.state` instead of `state` the issue was gone. The reason is that the qml label has a property named state https://doc.qt.io/qt-6/qml-qtquick-controls-label-members.html. So it took that one instead...

Digitalone1:
Thanks @wwmm, the issues are fixed, but I discovered two other problems: 

- The inline message in the presets menu is partially covered by the footer.

Screenshot From 2025-08-30 11-19-02

- When an effect is removed from the pipeline, it's UI is still visible (normally it should update to the next/previous effect or show the "no effects" message).

Screenshot From 2025-08-30 10-34-29

wwmm:
>The inline message in the presets menu is partially covered by the footer.

I saw something similar in the past. The truth is that the current approach used to set those OverlaySheets height is not good yet. I have to figure out a more robust solution to this. 

>When an effect is removed from the pipeline, it's UI is still visible (normally it should update to the next/previous effect or show the "no effects" message).

I probably introduced this regression while trying to avoid exactly this problem when loading a preset file that has no plugin in its plugins array. I will take a look at this.

wwmm:
>When an effect is removed from the pipeline, it's UI is still visible (normally it should update to the next/previous effect or show the "no effects" message).

The main problem is fixed. But I still have to think about the best way to select the next/previous plugin. Right now the code resets to the first plugin in the list. But at least the plugins stack is correctly updated to its widgets instead of showing the widgets from the removed plugin.

PR #4024, opened 2025-08-30
Fix translations, UI and add Preferences subtitles

Digitalone1:
I added some subtitles for the Preferences. @wwmm take a look at them and see if they are correct. Unfortunately the one for the Inactivity Timeout was too long to be shown completely, so I exposed the `maximumLineCount` for the EeSwitch.


wwmm:
It seems fine. Something that we have to remember to do is running 
```
cd util
./extract-translation-messages.sh
```
when adding new text to the ui. @vchernin how hard would it be to update that "update translation" CI with this?

PR #4026, opened 2025-08-30
Add qml format and qml lint

vchernin:
Adds 2 new tools for the qml files. These tools are much faster and less clunky than clang tidy thankfully. The main irritating thing here is for some reach the arch qt qml package does not put the qmllint and qmlformat binaries on the usual search paths, so I wrote a simple wrapper script for both commands that tries to find the binary.

I originally learned about these tools from reading this article. Some of these tips may be worth reviewing if you have not seen this already. I think some of the currently disabled checks in the qmllint ini file are relevant here.
https://planet.kde.org/kdab-on-qt-2024-10-23-10-tips-to-make-your-qml-code-faster-and-more-maintainable/

Another interesting tool is [qmlpreview](https://doc.qt.io/qt-6/qtqml-tooling-qmlpreview.html) which I did not try out. It lets you live edit the qml like "hot reloading" in web development. I wonder how difficult it would be to get that working here.

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -322,3 +322,42 @@ jobs:
 
       - name: Run KCFG numeric-defaults checker
         run: ./util/check-kcfg-numeric-defaults.sh
+
+  check-qmlformat:
+    name: Check QML format
+    runs-on: ubuntu-24.04
+    container:
+      image: archlinux
+    steps:
+      # install deps first to get git for later on
+      - name: Install deps
+        run: pacman -Syu --noconfirm qt6-declarative git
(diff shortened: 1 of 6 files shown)


wwmm:
> I originally learned about these tools from reading this article. Some of these tips may be worth reviewing if you have not seen this already. I think some of the currently disabled checks in the qmllint ini file are relevant here.
> https://planet.kde.org/kdab-on-qt-2024-10-23-10-tips-to-make-your-qml-code-faster-and-more-maintainable/

I saw some of these recommendations somewhere. But I am not sure where it was.

PR #4030, opened 2025-08-30
Fix update translation templates ci

vchernin:
This deletes the old translation update script and changes ci to call the newer one. I believe this is enough to manage the translations on a normal basis? We don't need any automation/ci for `util/port_po_files.py` correct?


wwmm:
> I believe this is enough to manage the translations on a normal basis?

I think it is enough.

> We don't need any automation/ci for `util/port_po_files.py` correct?

No. I asked Google Gemini to write this script just because I did not want to go through the effort of manually copying each translation file to our current po folder structure. This process won't have to be done again.

PR #4095, opened 2025-09-19
Fix error handling in translation templates update script

vchernin:
Yet another thing I missed. This is the testing you forget when you copy commands from a yaml file into your terminal instead of just making a `.sh` file.

@wwmm do we want to keep the circleci job around? The tests keep failing here. It looks like it has not been updated in while.


wwmm:
>do we want to keep the circleci job around? The jobs keep failing. It looks like it has not been updated in while.

I forgot to remove it. There is no need for it anymore.

Issue #4096, opened 2025-09-19
Wrong color scheme in the bottom bar on Gnome


Digitalone1:
<img width="1748" height="608" alt="Image" src="https://github.com/user-attachments/assets/51cd9da6-5578-433c-8c6c-da12ae4ae9ea" />

Already reported this issue in another ticket, but it's good to track it in a new one.

Running in service mode, this does not happen when I open the window for the first time. But opening it further, it's always reproducible.

It happens on Gnome, not in KDE. It would be useful to know how it works in other desktop environments.

wwmm:
What happens if your system is using a light theme?

Digitalone1:
> What happens if your system is using a light theme?

Quite the same.

<img width="1814" height="614" alt="Image" src="https://github.com/user-attachments/assets/7342ea08-ab16-4a62-96a6-0bbc5c2ce38f" />

wwmm:
Ok. Just to register this is how the light theme is here on KDE

<img width="1177" height="470" alt="Image" src="https://github.com/user-attachments/assets/be0430d8-1c26-40c6-987b-5ff7be3e3b66" />

Digitalone1:
@wwmm It seems the same. Is the light font intended for the sample rate and the level meter? 🤔

wwmm:
> [@wwmm](https://github.com/wwmm) It seems the same. Is the light font intended for the sample rate and the level meter? 🤔

Oh... I thought you were talking about the darker tone that the bottom bar has in your image. Or at least to my eyes your bottom bar seems darker than your top bar.

Wasn't the font of the pipeline information always like that? Or am I mistaken about its looks? It is possible 😄

wwmm:
When I implemented the `Kirigami.Action` that draws that information I set `enabled: false` to make its looks a little distinct from the rest because I thought it was that way in the gtk version too. Maybe it was in the past and I forgot we changed it. I do not remember now.

wwmm:
> Oh... I thought you were talking about the darker tone that the bottom bar has in your image. Or at least to my eyes your bottom bar seems darker than your top bar.

And in the dark theme your bottom bar color is lighter than the one in your top bar. It does not happen here on my computer

<img width="1177" height="470" alt="Image" src="https://github.com/user-attachments/assets/efc2858c-66f5-419a-9f39-d53965e26908" />
So I thought that was the problem you were talking about 😄

wwmm:
[...]

PR #4097, opened 2025-09-20
Add processing debug and fix selectable values in bottom bar

Digitalone1:
I think adding a debug before the size check could help us.

For the bottom bar, see https://github.com/wwmm/easyeffects/issues/4096#issuecomment-3315242604

### DIFF
--- a/src/contents/ui/PageStreamsEffects.qml
+++ b/src/contents/ui/PageStreamsEffects.qml
@@ -581,17 +581,27 @@ Kirigami.Page {
                     Kirigami.Action {
                         id: actionRateValue
 
-                        text: "kHz"
+                        // Sample rate
+
+                        displayComponent: Controls.Label {
+                            text: actionRateValue.text
+                            textFormat: Text.RichText
+                        }
                     },
                     Kirigami.Action {
                         id: actionLatencyValue
 
-                        text: "0,0 ms"
+                        // Latency
+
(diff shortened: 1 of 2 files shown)


wwmm:
Now that I am thinking about this again as the warning are there may be false positives. They will be printed when the latency changes too. What is totally valid. What really matters is if `pw_filter_get_dsp_buffer` returned a null pointer or not.

Digitalone1:
> Now that I am thinking about this again as the warnings are there may be false positives. They will be printed when the latency changes too. What is totally valid. What really matters is if `pw_filter_get_dsp_buffer` returned a null pointer or not.

Okay, so please change it accordingly when you have time.

wwmm:
> Okay, so please change it accordingly when you have time.

Done. And to my surprise while switching presets I do see some warnings about null pointers coming from `pw_filter_get_dsp_buffer`. Somehow those null pointers had no side effect on my computer. So bizarre :smile: .

In any case it is possible this is just how things are and I forgot about it.

wwmm:
> Done. And to my surprise while switching presets I do see some warnings about null pointers coming from `pw_filter_get_dsp_buffer`. Somehow those null pointers had no side effect on my computer. So bizarre 😄 .

Forget it. It was actually one of the warnings from this PR I had forgot to remove while adding the others. So far it seems that the null pointer do not happen here.

PR #4100, opened 2025-09-21
Various UI improvements

Digitalone1:
Added bold styling to highlight values in bottom bar:

 Screenshot From 2025-09-21 11-56-07  .

 Screenshot From 2025-09-21 11-57-04  .

Rate is always highlighted while the latency only if greater than 0 ms and the level meter only if greater than -10 db. @wwmm Is this okay for you?

Than I noticed that values in app info card are not localized. I moved the format in QML and added also decimal precision to latency:

Screenshot From 2025-09-21 11-55-04

### DIFF
--- a/src/contents/ui/DelegateStreamsList.qml
+++ b/src/contents/ui/DelegateStreamsList.qml
@@ -22,8 +22,8 @@ Kirigami.AbstractCard {
     required property string mediaClass
     required property string state
     required property string format
-    required property string rate
-    required property string latency
+    required property real rate
+    required property real latency
 
     visible: {
         if (!isBlocklisted)
@@ -76,7 +76,12 @@ Kirigami.AbstractCard {
                 RowLayout {
                     Controls.Label {
                         wrapMode: Text.WordWrap
-                        text: `${root.state} · ` + `${root.format} · ${root.rate} · ${root.nVolumeChannels} ` + i18n("channels") + ` · ${root.latency}`
+                        text: {
+                            const rate = root.rate.toLocaleString(Qt.locale(), 'f', 1);
+                            const latency = root.latency.toLocaleString(Qt.locale(), 'f', 1);
+
+                            return `${root.state} · ` + `${root.format} · ${rate} kHz · ${root.nVolumeChannels} ` + i18n("channels") + ` · ${latency} ms`;
+                        }
                         color: Kirigami.Theme.disabledTextColor
                     }
                 }
--- a/src/contents/ui/PageStreamsEffects.qml
+++ b/src/contents/ui/PageStreamsEffects.qml
@@ -633,14 +633,22 @@ Kirigami.Page {
                             actionLevelSaturation.visible = true;
                         else if (actionLevelSaturation.visible !== false)
                             actionLevelSaturation.visible = false;
-                        const localeLeft = left.toLocaleString(Qt.locale(), 'f', 0).padStart(4, ' ');
-                        const localeRight = right.toLocaleString(Qt.locale(), 'f', 0).padStart(4, ' ');
-                        const latency = Number(pipelineInstance.getPipeLineLatency()).toLocaleString(Qt.locale(), 'f', 1);
+
+                        const localeLeft = left.toLocaleString(Qt.locale(), 'f', 0).padStart(3, ' ');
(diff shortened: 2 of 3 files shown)


wwmm:
>Is this okay for you?

In the dark theme it looks totally fine. But in the light theme it seems a little too strong
image
I like the idea. But I wonder if there is a "lighter bold" that could be used.

wwmm:
Or maybe the white background just doesn't play nice with this idea. In a dark background the white bold improves clarity. But for some reason the dark bold in a white background does not provide the same feeling.

Digitalone1:
> In the dark theme it looks totally fine. But in the light theme it seems a little too strong 
>
> I like the idea. But I wonder if there is a "lighter bold" that could be used.

Do you mean only the rate and the latency? I can leave it only for the global level meter (if > -10).

I don't know if there is a lighter bold style. I should look at the documentation.

wwmm:
>Do you mean only the rate and the latency?

For some reason these two feel exaggerated in the light theme. But they do feel better this way in the dark theme.  The ideal solution would probably be to detect if the app is running with a dark theme and do that only if the theme is dark. But I am not sure if there is a straightforward way to get this information.

Digitalone1:
> For some reason these two feel exaggerated in the light theme. But they do feel better this way in the dark theme. The ideal solution would probably be to detect if the app is running with a dark theme and do that only if the theme is dark. But I am not sure if there is a straightforward way to get this information.

Okay, for the moment I will remove them for rate and latency.

wwmm:
I decided to have a conversation with chatgpt :smile: and it suggested something that I think deserves a proper investigation. Instead of doing the highlight through the bold tag we could set the color
```
actionRateValue.text = `<pre><span style="color:${Kirigami.Theme.textColor}">${rate}</span> <span style="color:${Kirigami.Theme.disabledTextColor}">kHz</span></pre>`;
```
.For example this produced
image
and
image
. The clear advantage of this approach is that the color will automatically follow the theme used by Kirigami. It does not have to necessarily be the colors above. It is a matter of figuring out which ones would be the best combination.

Just in case you haven't done it yet install `kirigami-gallery`. It is similar to the one for gtk4 and libadwaita and shows some kirigami code examples. Including the names of the theme colors.

Digitalone1:
> I decided to have a conversation with chatgpt 😄 and it suggested something that I think deserves a proper investigation. Instead of doing the highlight through the bold tag we could set the color
> 
> ```
> actionRateValue.text = `<pre><span style="color:${Kirigami.Theme.textColor}">${rate}</span> <span style="color:${Kirigami.Theme.disabledTextColor}">kHz</span></pre>`;
> ```
> 
> The clear advantage of this approach is that the color will automatically follow the theme used by Kirigami. It does not have to necessarily be the colors above. It is a matter of figuring out which ones would be the best combination.

Okay, but the boldness seems the same in your screenshots.

If I can set the [CSS](https://doc.qt.io/qt-6/richtext-html-subset.html#css-properties) inside the HTML, I could use the `font-weight` which controls how bold is the text.

I will make some tests.

wwmm:
>Okay, but the boldness seems the same in your screenshots.

I've changed only the one for the rate value. If you look at the image for the light theme you can see it is not as strong as the one I attached before. In any case a weaker bold is also a valid solution.

Digitalone1:
At last I applied both solutions. 😅

This is the one with only the color theme:

Screenshot From 2025-09-21 17-56-18 .
Screenshot From 2025-09-21 17-56-45 .

And this is the one with font-weight at 600 (standard bold is 700):

Screenshot From 2025-09-21 18-01-49 .
Screenshot From 2025-09-21 18-02-42 .

@wwmm What do you think? This mostly depends on the monospace font installed on the system.

With font-weight at 600 there is a difference, but it's slight. Do you want me to make the PR so you can test on your system?

wwmm:
>And this is the one with font-weight at 600 (standard bold is 700):

This one looks better. Feel free to do the PR.

Digitalone1:
There's also this other solution with font-weight at 700 and disabledTextColor.

Screenshot From 2025-09-21 18-18-53 .
Screenshot From 2025-09-21 18-20-02 .

It seems too light for the light theme. Anyway, I will make the PR with font-weight 600 and Theme.textColor, then you test it as you want.

PR #4102, opened 2025-09-21
XDG desktop entry: add StartupWMClass

Digitalone1:
Running Easy Effects I noticed a new default app icon is displayed in the Gnome dash. I remember this issue is fixed adding the `StartupWMClass` to the desktop entry. Indeed I copied the system desktop entry into `~/.local/share/applications` and added that line to avoid showing the additional icon.

Maybe this trick could also fix the missing icon in the development build on KDE.

@wwmm The desktop entry is automatically installed by cmake, right? So no need to change the PKGBUILD I suppose...

### DIFF
--- a/src/contents/com.github.wwmm.easyeffects.desktop
+++ b/src/contents/com.github.wwmm.easyeffects.desktop
@@ -7,6 +7,7 @@ Keywords=limiter;compressor;reverberation;equalizer;autovolume;
 Categories=AudioVideo;Audio;
 Exec=easyeffects
 Icon=com.github.wwmm.easyeffects
+StartupWMClass=wwmm.easyeffects.github.com
 StartupNotify=true
 Terminal=false
 Type=Application
--- a/src/contents/com.github.wwmm.easyeffects.desktop.template
+++ b/src/contents/com.github.wwmm.easyeffects.desktop.template
@@ -7,6 +7,7 @@ Keywords=limiter;compressor;reverberation;equalizer;autovolume;
 Categories=AudioVideo;Audio;
 Exec=easyeffects
 Icon=com.github.wwmm.easyeffects
+StartupWMClass=wwmm.easyeffects.github.com
 StartupNotify=true
 Terminal=false
 Type=Application
--- a/util/easyeffects-dev.desktop
+++ b/util/easyeffects-dev.desktop
@@ -7,6 +7,7 @@ Keywords=limiter;compressor;reverberation;equalizer;autovolume;
 Categories=AudioVideo;Audio;
 Exec=build/src/easyeffects
 Icon=easyeffects
+StartupWMClass=wwmm.easyeffects.github.com
 StartupNotify=true
 Terminal=false
 Type=Application


wwmm:
>The desktop entry is automatically installed by cmake, right? So no need to change the PKGBUILD I suppose...

Yes. And  as the translations are being handled now the workflow is editing `com.github.wwmm.easyeffects.desktop.template` and then running `extract-translation-messages.sh` to generate the new desktop file. Otherwise we risk loosing manually edited information in the final desktop file whenever the translation template is updated.

>Maybe this trick could also fix the missing icon in the development build on KDE.

Unfortunately not
image
. I suspect this needs one of those smaller size icons we do not provide right now. I mean like the ones inside the folder `/usr/share/icons/hicolor/`. We provide a svg icon. But maybe for the window title has to be a 16x16 png or similar size.

Digitalone1:
@wwmm Could be also that the icon name in 

https://github.com/wwmm/easyeffects/blob/master/util/easyeffects-dev.desktop#L9

is different than the ones in the other desktop entries?

Digitalone1:
@wwmm I think I found the solution:

https://nicolasfella.de/posts/fixing-wayland-taskbar-icons/

TLDR: Make sure that the window correctly reports the name of the .desktop file to the compositor. For Qt applications that boils down to making sure that [QGuiApplication::desktopFileName()](https://doc.qt.io/qt-5/qguiapplication.html#desktopFileName-prop) is correct.

Updated link for Qt6 documentation: https://doc.qt.io/qt-6/qguiapplication.html#desktopFileName-prop

wwmm:
> @wwmm Could be also that the icon name in
> 
> https://github.com/wwmm/easyeffects/blob/master/util/easyeffects-dev.desktop#L9
> 
> is different than the ones in the other desktop entries?

I used that `easyeffects-dev.desktop` only through the time our pkgbuild was not working in our desktop. My latest tests have been made with the same desktop file the users will use. We can remove that `easyeffects-dev.desktop` file from our sources.

wwmm:
> @wwmm I think I found the solution:
> 
> https://nicolasfella.de/posts/fixing-wayland-taskbar-icons/
> 
> TLDR: Make sure that the window correctly reports the name of the .desktop file to the compositor. For Qt applications that boils down to making sure that [QGuiApplication::desktopFileName()](https://doc.qt.io/qt-5/qguiapplication.html#desktopFileName-prop) is correct.
> 
> Updated link for Qt6 documentation: https://doc.qt.io/qt-6/qguiapplication.html#desktopFileName-prop

Interesting. I will add this and see what happens.

Digitalone1:
> I used that `easyeffects-dev.desktop` only through the time our pkgbuild was not working in our desktop. My latest tests have been made with the same desktop file the users will use. We can remove that `easyeffects-dev.desktop` file from our sources.

Yes, I found another trick, just read above.

wwmm:
I see that the docs are using a call for a qguiapplication. But kirigami apps use qapplication.

wwmm:
chatgpt suggests that for a qapplication the alternative is `QCoreApplication::setApplicationName`. What we already use. Maybe the value is the problem. It is just "easyeffects" there.

Digitalone1:
> chatgpt suggests that for a qapplication the alternative is `QCoreApplication::setApplicationName`. What we already use. Maybe the value is the problem. It is just "easyeffects" there.

Try to set it as the same I used for StartupWMClass.

wwmm:
And now that I tried to change it I remembered why I set it to `easyeffects`. The call to `QCoreApplication::setApplicationName` also sets the name of the easyeffects folder inside `.config`. And no other kde or qt app sets their configuration folder name following the same convention as the one for the icon.

>https://nicolasfella.de/posts/fixing-wayland-taskbar-icons/

I see now that he talks about taskbar icons. This one is fine. Only the one from the window titlebar is wrong.

wwmm:
>I see now that he talks about taskbar icons. This one is fine. Only the one from the window titlebar is wrong.

That is why I suspect not providing the needed icon size may be the reason. The taskbar is fine. The the tray icon is fine and the one for the about window too. It is just the window titlebar icon that is wrong.

Digitalone1:
> That is why I suspect not providing the needed icon size may be the reason. The taskbar is fine. The the tray icon is fine and the one for the about window too. It is just the window titlebar icon that is wrong.

I don't think this is the solution. If you look the list of files for [Alligator](https://archlinux.org/packages/extra/x86_64/alligator/) (a Kirigami app), there's only the SVG icon. I suppose the icon is showing for that app, so it should be doing something we're missing.

wwmm:
> I don't think this is the solution. If you look the list of files for [Alligator](https://archlinux.org/packages/extra/x86_64/alligator/) (a Kirigami app), there's only the SVG icon. I suppose the icon is showing for that app, so it should be doing something we're missing.

Its icon works indeed.

wwmm:
Problem fixed. I saw in alligator code that when built for Android they set `QApplication::setWindowIcon(QIcon::fromTheme(QStringLiteral("alligator")));`. Doing something similar on our side fixed the problem

image

I still have no idea why we need to set this while alligator and others do not seem to need this call outside of Android. But whatever... It is working now :smile: .

Digitalone1:
Meanwhile, after the new compilation, the icon is broken on my system. 😂😂😂

image

I will try to fix it.

Digitalone1:
I think I solved it. QApplication is derived by QGuiApplication, so it can use their methods. I managed to set the displayed name on the dash and the icon is showing correctly even if the StartupWMClass is incorrect:

Screenshot From 2025-09-21 22-58-52

PR #4104, opened 2025-09-21
Use QApplication to set the app name

Digitalone1:
@wwmm This shows the correct app name and the proper icon in Gnome dash.

I commented the `setWindowIcon` you added since I'm pretty sure is unneeded and the icon will show in the title bar or your system. If I'm wrong, please, decomment it and test what happens.


wwmm:
> I commented the `setWindowIcon` you added since I'm pretty sure is unneeded and the icon will show in the title bar or your system. If I'm wrong, please, decomment it and test what happens.

At least for now it is fine without it indeed.

PR #4109, opened 2025-09-23
Plugin base: Continuous warnings on null probe array

Digitalone1:
Related to https://github.com/wwmm/easyeffects/issues/4085#issuecomment-3323535537

@wwmm I keep getting uninterrupted warnings while playing audio with my preset. Please test it.

### DIFF
--- a/src/plugin_base.cpp
+++ b/src/plugin_base.cpp
@@ -67,7 +67,7 @@ void on_process(void* userdata, spa_io_position* position) {
     return;
   }
 
-  // We had to add the following checks for vector sizes. See #4085
+  // We had to add the following checks for dummy array sizes. See #4085
   if (d->pb->dummy_left.size() != n_samples) {
     d->pb->dummy_left.resize(n_samples);
 
@@ -103,7 +103,7 @@ void on_process(void* userdata, spa_io_position* position) {
   if (in_left != nullptr) {
     left_in = std::span(in_left, n_samples);
   } else {
-    util::warning("We received a null left_in pointer. Using the dummy array instead.");
+    util::warning("processing: we received a null left_in pointer. Using the dummy array instead.");
 
     left_in = d->pb->dummy_left;
   }
@@ -111,7 +111,7 @@ void on_process(void* userdata, spa_io_position* position) {
   if (in_right != nullptr) {
     right_in = std::span(in_right, n_samples);
   } else {
-    util::warning("We received a null right_in pointer. Using the dummy array instead.");
+    util::warning("processing: we received a null right_in pointer. Using the dummy array instead.");
 
     right_in = d->pb->dummy_right;
   }
@@ -119,7 +119,7 @@ void on_process(void* userdata, spa_io_position* position) {
   if (out_left != nullptr) {
     left_out = std::span(out_left, n_samples);
   } else {
-    util::warning("We received a null left_out pointer. Using the dummy array instead.");
+    util::warning("processing: we received a null left_out pointer. Using the dummy array instead.");
 
     left_out = d->pb->dummy_left;
   }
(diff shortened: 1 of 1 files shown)


wwmm:
Showing warnings for the probe will have some side effects. They will be not null only in the plugins that actually create probes. For all the others those warnings will be shown every time PipeWire invokes the process callback. 

Now that I am thinking about it something similar may happen to the people that have microphone drivers that generate only one channel. An eternal warning will be shown for a channel that will never be available. I will try to add a count for these warnings and stopping showing them after the second or third occurrence.

wwmm:
>Showing warnings for the probe will have some side effects. They will be not null only in the plugins that actually create probes. For all the others those warnings will be shown every time PipeWire invokes the process callback.

Actually that part of the code should only be executed if a plugin set `d->pb->enable_probe`. Let me see why I suddenly have a plugin showing probe warnings non stop...

Digitalone1:
So having null pointers is not normal when the plugin uses probes, but if the user has a mono microphone, one channel will always provide a null pointer. Did I understand correctly?

wwmm:
Oh... When the limiter sidechain input is not set to external `pw_filter_get_dsp_buffer` returns a null pointer for the probes. Probably because in this case there is no link to our pw_filter ports used by the probes....

wwmm:
>but if the user has a mono microphone, one channel will always provide a null pointer. Did I understand correctly?

I still have to verify this. The problem is that I think none of microphones have a mono mode. The drivers create two channels for them even if it does not make much sense.

wwmm:
`d->pb->enable_probe == true` just means that the plugin will call the `process` overload that has probes. It does not mean the plugin actually requested links to its probes. So showing warnings for probes will be a little tricky. It is better to disable these warnings for now.

In any case this makes it clear that if there is no link to a port `pw_filter_get_dsp_buffer` returns a null pointer. So those cases where the user mic has only one channel will definitely have a permanent warning.

wwmm:
>It is better to disable these warnings for now.

Or putting a counter and stopping showing them. I will try that.

Digitalone1:
> `d->pb->enable_probe == true` just means that the plugin will call the `process` overload that has probes. It does not mean the plugin actually requested links to its probes. So showing warnings for probes will be a little tricky. It is better to disable these warnings for now.
> 
> In any case this makes it clear that if there is no link to a port `pw_filter_get_dsp_buffer` returns a null pointer. So those cases where the user mic has only one channel will definitely have a permanent warning.

Okay, so they are null if they do not use the sidechain. So okay, it does not make much sense to show them. But I suppose this happens also for Compressor, right?

wwmm:
>But I suppose this happens also for Compressor, right?

Yes

wwmm:
I've updated our master branch with something that seems a good compromise between removing the warnings and keeping them as they were. Now we print the warning once. In case PipeWire changes the quantum value we reset the "got null" state. This should be good enough for us to know something might be wrong without spamming the user logs when `pw_filter_get_dsp_buffer` always returns null.

Digitalone1:
> I've updated our master branch with something that seems a good compromise between removing the warnings and keeping them as they were. Now we print the warning once.

Nice.

> In case PipeWire changes the quantum value we reset the "got null" state. This should be good enough for us to know something might be wrong without spamming the user logs when `pw_filter_get_dsp_buffer` always returns null.

Okay, but I don't see the reset of the bool flags in the latest changes. Am I missing something?

wwmm:
> Okay, but I don't see the reset of the bool flags in the latest changes. Am I missing something?

Here https://github.com/wwmm/easyeffects/blob/d9224c583d623eb938cfdb22cd7570a32bb6b2a1/src/plugin_base.cpp#L87

Digitalone1:
> Here 
> https://github.com/wwmm/easyeffects/blob/d9224c583d623eb938cfdb22cd7570a32bb6b2a1/src/plugin_base.cpp#L87

Okay, nice.

PR #4110, opened 2025-09-23
Plugin base: wait for Pipewire to provide valid pointers

Digitalone1:
Please, test #4109 before.

@wwmm Then I wanted to know if you like this solution to get more protection from null pointers from Pipewire.

### DIFF
--- a/src/plugin_base.cpp
+++ b/src/plugin_base.cpp
@@ -89,11 +89,39 @@ void on_process(void* userdata, spa_io_position* position) {
 
   // util::warning("processing: " + util::to_string(n_samples));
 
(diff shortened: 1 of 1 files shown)


wwmm:
Based on what we are talking about in the other PR that loop will have a big problem. In some cases `pw_filter_get_dsp_buffer` will never return a non null pointer.

And as the plugins thread is supposed to run in realtime having a waiting loop may not be a good idea. And talking about realtime I wonder when xdg-desktop-portal will have a new release fixing the bug preventing pipewire from setting realtime priorities to the filter thread...

Digitalone1:
Okay, so this is not a good solution, you can discard it.

PR #4123, opened 2025-09-28
Coding style: Use doxygen style blocks for multiline comments

Digitalone1:
Comments with more than 2 lines have been updated to use the doxygen block style:

```
/**
 * Text...
 */
```

Other multiline comments similar to doxygen have been updated to use the doxygen notation which is more readable.

Other comments with 1 or 2 lines using the doxygen style have been reverted to the normal `// ` notation which uses less space. 

Very long single line comments have been split to be more readable.

### DIFF
--- a/src/autogain.cpp
+++ b/src/autogain.cpp
@@ -1,20 +1,20 @@
-/*
- *  Copyright © 2017-2025 Wellington Wallace
+/**
+ * Copyright © 2017-2025 Wellington Wallace
  *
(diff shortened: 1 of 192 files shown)


wwmm:
I wonder if there is a way to make clang-format do some of that for us.

Digitalone1:
> I wonder if there is a way to make clang-format do some of that for us.

Unfortunately not.

PR #4133, opened 2025-10-01
Fix error in translate strings

Drommer:


Drommer:
It would also be great to add plural forms for line 135 (frames). But I don't have enough knowledge to do that.

wwmm:
Thanks! Somehow I did not notice this error.

> It would also be great to add plural forms for line 135 (frames). But I don't have enough knowledge to do that.

Based on what is in [KDE's docs](https://invent.kde.org/frameworks/ki18n/-/blob/master/docs/programmers-guide.md)  it would be `unit: i18np("frame", "frames", pluginDB.minProcessingBuffer)`. But for some reason its behavior seems strange. It is switching between plural and singular as I change the parameter value in a way that is not making any sense to me.

Drommer:
In the examples I found, there is always a variable "%n" that contains a number. The choice of form depends on this value. At least, that's how I understand it.

wwmm:
> In the examples I found, there is always a variable "%n" that contains a number. The choice of form depends on this value. At least, that's how I understand it.

But somewhere in that doc they say that inserting a "%n" is optional. What is mandatory is the integer variable in the first argument. And as far as I could see there is no error when the program runs. But At least on my end sometimes it alternates between plural and singular forms for some numbers. Really strange...

Drommer:
> But somewhere in that doc they say that inserting a "%n" is optional. What is mandatory is the integer variable in the first argument. And as far as I could see there is no error when the program runs. But At least on my end sometimes it alternates between plural and singular forms for some numbers. Really strange...

I checked the implementation of SpinBox with a number and text in the KDE Plasma settings interface. Instead of `unit:` they use a construction with two functions:
```
textFromValue: function(value, locale) {
  return i18np("%1 second", "%1 seconds", value)
}
valueFromText: function(text, locale) {
  return parseInt(text)
}
```

For example: [this](https://phabricator.kde.org/D26529#change-OdZrdk9ga65v) or [this](https://github.com/KDE/plasma-workspace/blob/master/kcms/cursortheme/ui/LaunchFeedbackDialog.qml).

P.S. I also think that a translation option should be added for lines such as `unit: "dB"`. This is spelled differently in languages that use the Cyrillic alphabet. In addition, some other lines (such as filter sharpness) contain translations.

wwmm:
Our spinbox uses as value `const t = Number(decimalValue).toLocaleString(locale, 'f', control.decimals) + unit_str;`. Doing the same as the spinbox from the KDE repository isn't viable for us in my opinion.  We need the number to follow user locale but we also need to be able to control the number of decimal places. And to have units or not at the end. Putting something like  `return i18np("%1 second", "%1 seconds", value)` there will get in the way of all of this.

I still think that `unit: i18np("frame", "frames", pluginDB.minProcessingBuffer)` should have worked. This would be the ideal solution. Specially considering that we do not need to insert units through i18np in most of the cases.

wwmm:
> P.S. I also think that a translation option should be added for lines such as `unit: "dB"`. This is spelled differently in languages that use the Cyrillic alphabet. In addition, some other lines (such as filter sharpness) contain translations.

I do not see any problem in doing this. We probably just never considered necessary. Or was there a reason and I forgot about it @Digitalone1?

Drommer:
> Our spinbox uses as value `const t = Number(decimalValue).toLocaleString(locale, 'f', control.decimals) + unit_str;`. Doing the same as the spinbox from the KDE repository isn't viable for us in my opinion. We need the number to follow user locale but we also need to be able to control the number of decimal places. And to have units or not at the end. Putting something like `return i18np("%1 second", "%1 seconds", value)` there will get in the way of all of this.

I'm not suggesting changing the implementation in every SpinBox call. In most cases one- or two-letter abbreviations are used which does not require different options depending on the value. However, you can try it for this particular field. We are talking about frames. The step is 1. That means only integer values.

wwmm:
> I'm not suggesting changing the implementation in every SpinBox call.

That approach requires changes to our custom widget code. It can't be done just in the places that are needed. Even if I tried something like
```
units: function(value, locale) {
  return i18np("%1 frame", "%1 frames", value)
}
```
so that the widget code is not modified the final result would be to have the numbers displayed twice before the unit like ` 1 1 frame`. For this to not happen the widget implementation has to be modified. What is not ideal. That is why I would like to have an output of `unit: i18np("frame", "frames", pluginDB.minProcessingBuffer)` that makes sense. Right now it changes with the given `pluginDB.minProcessingBuffer` value. But in ways that are not making sense to me. For example I see
```
0 frame, 1 frames, 2 frame
```
when the number is increased and 
```
2 frames, 1 frames, 0 frame
```
when I decrease the number. It does not make sense. If `%1` were mandatory it should either show an error or warning or do nothing at all. But it does alternate between plural and singular in ways that are not right.

wwmm:
>the final result would be to have the numbers displayed twice before the unit like  1 1 frame

Out of curiosity I decided to do that and number that comes from i18np does not match the one displayed by our spinbox. This probably explains why the plural is not making sense. My guess is that the current internal implementation of our custom spinbox does not play nice with how i18np wants to work...

Drommer:
> I still think that `unit: i18np("frame", "frames", pluginDB.minProcessingBuffer)` should have worked. This would be the ideal solution. Specially considering that we do not need to insert units through i18np in most of the cases.

I made a build with the following changes. I changed 
```
unit: i18n("frames")
```
to
```
unit: i18np("frame", "frames", pluginDB.minProcessingBuffer)
```
 in the **DeepFilterNet.qml** file and also changed 
```
msgid "frames"
msgstr "кадр(-а,-ів)"
```
to
```
#, c-format
msgid "frame"
msgid_plural "frames"
msgstr[0] "кадр"
msgstr[1] "кадра"
msgstr[2] "кадрів"
```
in the **easyeffects.po** file.

Overall, it works almost correctly, but the translation always lags behind by one value. This means that the translation is not updated immediately after the value is changed. Which means that the displayed translation is correct for the previous value.

wwmm:
>Overall, it works almost correctly, but the translation always lags behind by one value. This means that the translation is not updated immediately after the value is changed. Which means that the displayed translation is correct for the previous value.

It is what I am seeing. For some weird reason the value returned by i18np is lagging by one. At this moment I do not know how this is even possible. The value saved in `pluginDB.minProcessingBuffer` is correct. The one displayed by the spinbox is also correct. But somehow the output i18np lags by one =/

wwmm:
I imagine that the issue is at `const t = Number(decimalValue).toLocaleString(locale, 'f', control.decimals) + unit_str;`. Most likely `decimalValue` and the value used by `i18np` that will eventually generate `unit_str` are not being updated at the same time.

Drommer:
I was reviewing the QtQuick documentation and noticed that it mentions `suffix`, which is intended to be used for adding units of measurement in SpinBox. It also says that when `value` is changed it sends a signal that also updates the suffix and prefix. However, I did not see any mention of `unit` there. Why do you use exactly `unit`?

wwmm:
> I was reviewing the QtQuick documentation and noticed that it mentions `suffix`, which is intended to be used for adding units of measurement in SpinBox. It also says that when `value` is changed it sends a signal that also updates the suffix and prefix. However, I did not see any mention of `unit` there. Why do you use exactly `unit`?

I suppose you are talking about this example from qml spinbox docs
```
SpinBox {
    id: spinBox
    from: -100
    value: 11
    to: 100
    editable: true
    anchors.centerIn: parent

    property string prefix: "L="
    property string suffix: "m"

    readonly property regexp numberExtractionRegExp: /\D*?(-?\d*\.?\d*)\D*$/

    validator: RegularExpressionValidator { regularExpression: numberExtractionRegExp }

    textFromValue: function(value, locale) {
        return prefix + Number(value).toLocaleString(locale, 'f', 0) + suffix
    }

    valueFromText: function(text, locale) {
        return Number.fromLocaleString(locale, numberExtractionRegExp.exec(text)[1])
    }
}
```
It is exactly the same thing we do. I just decided to name the variable as unit instead of suffix.

wwmm:
Something that I see now that will happen both with our spinbox as well as with the qml exmple above is that changes to unit/suffix will never trigger a new invocation of the callback set to `textFromValue`. Even if there is a binding from i18np to the unit/suffix parameter `textFromValue` callback has no relation to this binding at all. Only when the spinbox value is changed that callback will be called. But this happens before we update the `pluginDB.minProcessingBuffer` parameter that is used by i18np.

For this to have any chance of working our spinbox implementation will have to be changed in a way that `i18np` is used inside the `textFromValue` callback. Maybe by create an additional unit parameter dedicated to this case. This way it would be easier to change code paths inside our custom spinbox to one that uses `i18np`.

Digitalone1:
> I do not see any problem in doing this. We probably just never considered necessary. Or was there a reason and I forgot about it @Digitalone1?

I didn't consider it necessary, but it' OK to add it anyway.

PR #4147, opened 2025-10-05
Enhance PreferencesSheet with subtitle for system notifications

Digitalone1:
I'm with the smartphone. The commit message has been generated by Copilot.

I though it would be better to explain the Ignore Systems Notification option.

@wwmm Please review it and see if it could be written in a better English. If you can, format it in VS Code also since I wrote it with the Github editor from the smartphone. Thanks.


wwmm:
It seems good to me. And qml's linter is not suggesting any format change.

Issue #4150, opened 2025-10-06
[Feature Request] Changing settings from the outside at runtime


owlonpc:
I used to depend on the ability to change LSP loudness settings from scripts and such using gsettings. Since the project moved to QT, I wonder if you could expose these things through dbus instead. Thank you in advance.

![Image](https://github.com/user-attachments/assets/5d522b3c-4877-4ab3-88f9-0b5b8cc8a6aa)

wwmm:
I will try to do that through command line options. DBUS is a pain to deal with.

wwmm:
>I will try to do that through command line options.

As we do not need to expose many parameters it should be a viable approach.

owlonpc:
Thanks, for now I have made this, https://github.com/owlonpc/easyeffects-scroll-loudness. I don't want this to be the way, so I'm not closing the issue.

wwmm:
I've implemented the basic infrastructure necessary to set plugins parameters through the command line. For now only the `volume` property of the `loudness` plugin can be set but it should be easy to add support for more properties and plugins https://github.com/wwmm/easyeffects/blob/5e6cf7b76b3f5dc84cd16b7372cb65406867109b/src/local_server.cpp#L128. But I would really like to add only what is really necessary. Looking at the past most people that wanted to set properties through third party tools had as goal to create a tray icon. Now that we have one only corner cases like the one from the loudness plugin need our properties to be accessed from external tools.

Running `./easyeffects -h` a help about the command can be seen. But for the particular case of the loudness plugin volume it is something like
```
easyeffects --set-property output:loudness:0:volume:-10
```
Pipeline followed by plugin name, plugin instance, parameter name and its value.

owlonpc:
Thanks, it's great!

owlonpc:
<img width="1304" height="114" alt="Image" src="https://github.com/user-attachments/assets/5f281868-c444-404a-9a0a-a5057ab70ec9" />

Setting the properties is obviously very slow, and getting properties is not yet possible. This is getting complicated.

wwmm:
>Setting the properties is obviously very slow

I am not sure if there is much that can be done about it with the current approach. But is there a reason why this has to be super fast?

>and getting properties is not yet possible

As the new instance created in the terminal has to communicate with the other getting the values will be a little more complicated. But it should not be impossible.

Digitalone1:
I think the regex could be improved a little bit, but we can't do much. Regex is slow unfortunately...

[...]

PR #4177, opened 2025-10-14
Improve set_property regex

Digitalone1:
Related to #4150 

It shouldn't make so much difference, but it's still an improvement.

Anyway, I got a core dump when I misspelled the pipeline. For example `--set-property outpur:loudness:0:volume:5`.

```
easyeffects: command_line_parser.cpp:197	Invalid pipeline type. Must be 'input' or 'output'
Annullato                  (core dump creato) easyeffects --set-property outpur:loudness:0:volume:5
``` 

Is this normal?

Besides, maybe we need something to get the default value. For example `double volume = plugin_db->getDefaultVolume();`

Isn't this possible at the moment?

### DIFF
--- a/src/contents/kcfg/easyeffects_db_autogain.kcfg
+++ b/src/contents/kcfg/easyeffects_db_autogain.kcfg
@@ -30,7 +30,7 @@
             <default>-23</default>
         </entry>
         <entry name="referenceLabels" type="StringList">
-            <default>Momentary,Shortterm,integrated,Geometric Mean (MSI),Geometric Mean (MS),Geometric Mean (MI),Geometric Mean (SI)</default>
+            <default>Momentary,Shortterm,Integrated,Geometric Mean (MSI),Geometric Mean (MS),Geometric Mean (MI),Geometric Mean (SI)</default>
         </entry>
         <entry name="reference" type="Int">
             <label>Reference</label>


Digitalone1:
In case we have a string value with white spaces, we should write `--set-property "output:autogain:0:reference:Geometric Mean (MSI)"`. This should work, am I right?

Besides, I noticed that `Integrated` is not written with capital letter in the string list:

https://github.com/wwmm/easyeffects/blob/dc78b0d62e8a123c35afc57e963ecc27fcbc8e81/src/contents/kcfg/easyeffects_db_autogain.kcfg#L33C53-L33C73

Update: There's also Autogain Force Silence that is not loaded after it has been saved into a preset.

wwmm:
>Annullato                  (core dump creato) easyeffects --set-property outpur:loudness:0:volume:5

That happens because of `util::fatal`. Maybe we should just use a warning there.

>Besides, maybe we need something to get the default value. For example double volume = plugin_db->getDefaultVolume();

KConfig XT already generates member functions for default values. In the particular case of the volume setting it is `settings->defaultVolumeValue()`;. 

>Update: There's also Autogain Force Silence that is not loaded after it has been saved into a preset.

Oh... I forgot to add it to the preset class...

Digitalone1:
> KConfig XT already generates member functions for default values. In the particular case of the volume setting it is `settings->defaultVolumeValue()`;.

Okay. But in order to expose all the parameters as like gsettings, should we redeclare all of them or there's something much easier in which we can pass the parameter name and the value and it gets updated?

wwmm:
> Okay. But in order to expose all the parameters as like gsettings, should we redeclare all of them or there's something much easier in which we can pass the parameter name and the value and it gets updated?

Redeclaring all of them manually will require an insane amount of work. It is not worth at all considering that it is a very nice application. It will probably be a lot easier to do something similar to what we already have in https://github.com/wwmm/easyeffects/blob/master/src/kconfig_base_ee.cpp. All the database classes inherit from the KConfigBaseEE class. We just have to create a setter and a getter. The only potential annoyance is that there what we have is a QVariant and not the type actually used by the variable.

PR #4186, opened 2025-10-17
CI fixes

vchernin:
Fixes the flatpak build. Also replaces the fedora ci build with a opensuse build, as unfortunately I could not find a fedora package for webrtc-audio-processing 2. Opensuse did not even have it in the default repositories, but it was in one I found.

I did notice that the build reports confusing errors when missing some dependencies, it was a bit tricky figuring out all the qt dependencies. Some of them like `qt6-quickshapes` seem to not be in the easyeffects cmake files but are definitely needed at least with the opensuse qt packages. For example:

```
CMake Warning at /usr/lib64/cmake/Qt6/Qt6Config.cmake:243 (find_package):
  Found package configuration file:

    /usr/lib64/cmake/Qt6Graphs/Qt6GraphsConfig.cmake

  but it set Qt6Graphs_FOUND to FALSE so package "Qt6Graphs" is considered to
  be NOT FOUND.  Reason given by package:

  Qt6Graphs could not be found because dependency Qt6QuickShapesPrivate could
  not be found.

  Configuring with --debug-find-pkg=Qt6QuickShapesPrivate might reveal
  details why the package was not found.

  Configuring with -DQT_DEBUG_FIND_PACKAGE=ON will print the values of some
  of the path variables that find_package uses to try and find the package.

Call Stack (most recent call first):
  CMakeLists.txt:34 (find_package)
```

It may be best to try to depend on qt6-quickshapes directly, but I am not sure how that is best done.

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -6,6 +6,10 @@ on:
       - 'dependabot/**'
   pull_request:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 name: CI
 jobs:
   prepare:


wwmm:
Thanks!

>as unfortunately I could not find a fedora package for webrtc-audio-processing 2

Hum... As version 2.0 has been around for 9 months I thought distributions would be offering it already.

>Some of them like qt6-quickshapes seem to not be in the easyeffects cmake files but are definitely needed at least with the opensuse qt packages. 

Hum... Arch Linux does not even have a package with this name. It does not even list any package as optional qt6-graphs dependency. As far I could see from system files the package providing `QuickShapesPrivate` is `qt6-declarative`. And this one is already forced in our cmake files when find_package requires `Quick`.

This may be OpenSUSE doing to Qt packages things they shouldn't...

PR #4192, opened 2025-10-19
Add a label and a signal to notify RNNoise model changes + various minor changes

Digitalone1:
RNNoise plugin is tricky because we can end up using the standard model even if the standard model flag is unset and this state is not clearly reported in the UI leading to confusion. This is not happening only after an attempt to load a corrupted model.

For example, reset the plugin and uncheck the standard model switch. If the model list is filled by custom models, an entry is selected. This might be think that we are using the selected model, but in reality we are **still using the standard model**.

This is misleading, so we need something like the Convolver where there's a label that shows the loaded impulse response. I added a small label below the cards to show the model currently used.

This is the state when the plugin is reset and the standard model switch is unset.

<p>Schermata del 2025-10-19 17-52-22<p>

But before doing this, I added a Qt signal to inform the user about the loading status. If the model is corrupted, the standard one is used as fallback. But the user should know why the standard model has been loaded or why the model is not changing, so we need these messages to be shown.

For the label, it has been a little tricky to implement the initial state because I was not sure the signal was caught, especially when the user enters the UI after the backed has finished its construction. In the end, I added a Q_PROPERTY on the `standard_model` bool property to check it while the UI is initializing.

This is only needed for the initialization, indeed checking the `standard_model` bool flag in the plugin backend is not enough when the model is selected in the list, because it might change later when the backend tries to load it and fails. In this case we need the signals.

@wwmm: please test it. It seems good on my side.

The only remaining issue are the following messages shown on the initialization: 

```
qrc:/ui/RNNoise.qml:238:29: QML Connections: Detected function "onStandardModelLoaded" in Connections element. This is probably intended to be a signal handler but no signal of the target matches the name.
qrc:/ui/RNNoise.qml:238:29: QML Connections: Detected function "onCustomModelLoaded" in Connections element. This is probably intended to be a signal handler but no signal of the target matches the name.
``` 

How can we get rid of them?

### DIFF
--- a/src/contents/ui/RNNoise.qml
+++ b/src/contents/ui/RNNoise.qml
@@ -192,7 +192,6 @@ Kirigami.ScrollablePage {
                             onClicked: {
                                 rnnoisePage.pluginDB.modelName = name;
                                 listItemDelegate.ListView.view.currentIndex = index;
-                                appWindow.showStatus(i18n("Loaded Model: %1", name), Kirigami.MessageType.Positive);// qmllint disable
                             }
 
                             contentItem: RowLayout {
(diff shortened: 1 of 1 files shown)


wwmm:
> How can we get rid of them?

It is probably happening because qml is evaluating the connection before the property pluginBackend is initialized. Eventually it auto updates to the proper value and everything works. 

Let me see how to make the connection take this into account.

wwmm:
The easiest "fix" is using `ignoreUnknownSignals` https://doc.qt.io/qt-6/qml-qtqml-connections.html. I think it is totally safe to do that here. Qml is able to update the connection on its own once pluginBackend becomes ready. Silence this warning should be fine in this case.

I will do that now.

wwmm:
Done. As I am not sure if qml would be dumb and create a new connection instance for each delegate instance I moved the connection creation to the rnnoise window top widget.

PR #4198, opened 2025-10-21
Improve Kirigami inline messages + various minor changes

Digitalone1:
- Added a context to Kirigami inline messages.
- Added a new util to check null or undefined values.
- `showStatus` modified to specify default values and `autohide` argument.
- `qsTr` replaced with `i18n`.
- `name` property declared as string in all plugin UI.

@wwmm Please take a look at weblate since I don't see new translations anymore, maybe there's something going wrong...


wwmm:
> Please take a look at weblate since I don't see new translations anymore, maybe there's something going wrong...

Things seem fine in the maintenance field.

Digitalone1:
> Things seem fine in the maintenance field.

I see the new translations in weblate now.

I forgot to mention that I had to remove the message for the Failing to Load of a Community Preset. It was showing over the messages generated by the JSON parser containing the details of the error inside the preset.

PR #4213, opened 2025-10-25
use max of absolute samples for `get_peaks`

edwloef:
The peak meter was able to show lower peak values than were actually present if the audio was asymmetrical towards the negative side. Since clipping can happen below -1.0 as well, use the max of the absolute samples to determine the peaks.

This matches the behavior of:
- Audacity: https://github.com/audacity/audacity/blob/master/au3/libraries/lib-audio-io/AudioIO.cpp#L123
- LMMS: https://github.com/LMMS/lmms/blob/master/include/SampleFrame.h#L200


wwmm:
It makes sense. Thanks!

PR #4215, opened 2025-10-26
Convolver: inform the user about the Passthrough Mode

Digitalone1:
When the Convolver plugin backed was failing to load a kernel, the user got a positive inline message on the correct loading.

Now the user gets a negative message on failing loadings and an informative message when the plugin is reset (entering in pass-through mode).

Besides, the chart banner is now properly updated to reflect these states.

The only remaining thing (which I'm not sure how to do it), is to reset the graph at the initial state when the pass-through mode is signaled. @wwmm What do you suggest? To give en empty array to the `updateData`? Or maybe you could implement a reset method which could be more straightforward to understand?

### DIFF
--- a/src/contents/ui/Convolver.qml
+++ b/src/contents/ui/Convolver.qml
@@ -42,6 +42,23 @@ Kirigami.ScrollablePage {
     }
 
     Connections {
+        function onNewKernelLoaded(name, success) {
+            if (success) {
+                convolverChartContainer.banner.title = name;
+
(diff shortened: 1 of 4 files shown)


wwmm:
> What do you suggest? To give en empty array to the `updateData`? Or maybe you could implement a reset method which could be more straightforward to understand?

`updateData` ignores empty arrays. So that won't work. `SplineSeries` and similars have a `clear()` method for that. But only calling it is not enough. Those `QList<QPointF>` from the convolver also have to be cleared. Otherwise any event that triggers widgets redraw will bring the chart back. I will take a look at this.

PR #4218, opened 2025-10-26
Don't depend on deprecated appstream-util

ximion:
Hi!

This change drops the unneeded dependency on appstream-util, which we are trying to remove from Debian. The utility is deprecated and should no longer be used for modern metadata.

I did not touch the `util/update-release-files.sh` script, which still uses it. But it may make sense to sort that out too (you can pass `--strict` to `appstreamcli` to get strict validation behavior like with `appstream-util`).

Also, it looks like the cmake config doesn't install the metadata into `KDE_INSTALL_METAINFODIR` yet...

Thanks for considering the patch!
Best,
    Matthias

### DIFF
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -35,7 +35,7 @@ depends=(
   'libportal-qt6'
   'webrtc-audio-processing'
 )
-makedepends=('cmake' 'extra-cmake-modules' 'git' 'ninja' 'intltool' 'appstream-glib' 'mold' 'ladspa')
+makedepends=('cmake' 'extra-cmake-modules' 'git' 'ninja' 'intltool' 'appstream' 'mold' 'ladspa')
 optdepends=('calf: limiter, exciter, bass enhancer and others'
             'lsp-plugins: equalizer, compressor, delay, loudness'
             'zam-plugins: maximizer'
--- a/PKGBUILD_AUR
+++ b/PKGBUILD_AUR
@@ -35,7 +35,7 @@ depends=(
   'libportal-qt6'
   'webrtc-audio-processing'
 )
-makedepends=('cmake' 'extra-cmake-modules' 'git' 'ninja' 'intltool' 'appstream-glib' 'ladspa')
+makedepends=('cmake' 'extra-cmake-modules' 'git' 'ninja' 'intltool' 'appstream' 'ladspa')
 optdepends=('calf: limiter, exciter, bass enhancer and others'
             'lsp-plugins: equalizer, compressor, delay, loudness'
             'zam-plugins: maximizer'
--- a/debian/control
+++ b/debian/control
@@ -3,7 +3,7 @@ Section: sound
 Priority: optional
 Maintainer: Boyuan Yang <byang@debian.org>
 Build-Depends:
- appstream-util,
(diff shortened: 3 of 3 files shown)


vchernin:
Main author of `util/update-release-files.sh` here, I would be very happy to completely remove all traces of `appstream-util` from this repo. 

I almost did so in a recent set of changes, but I noticed [fedora's packaging docs](https://docs.fedoraproject.org/en-US/packaging-guidelines/AppData/) still say to use `appstream-util` for rpm packages. I just accepted it and didn't investigate further at the time. Looking again, I see those docs have not been updated in over two years, so maybe they are just out of date?

Are there any modern-ish distros that have not migrated from `appstream-util`?

ximion:
Oh, nice!

@conan-kudo Can you have a look at the Fedora docs? Those should probably be updated at this point...

> Are there any modern-ish distros that have not migrated from appstream-util?

At this point, since pretty much all of GNOME is migrated over, I don't think there is any distribution where depending exclusively on appstream / appstreamcli would be a problem :-)

wwmm:
As this PR only touches Arch's pkgbuild and Debian files it seems to me it is fine to merge it. Or am I missing something?

Conan-Kudo:
> Oh, nice!
> 
> @Conan-Kudo Can you have a look at the Fedora docs? Those should probably be updated at this point...
> 
> > Are there any modern-ish distros that have not migrated from appstream-util?
> 
> At this point, since pretty much all of GNOME is migrated over, I don't think there is any distribution where depending exclusively on appstream / appstreamcli would be a problem :-)

Fedora (and other RPM distros) still use appstream-builder for the AppStream repodata, so we have to use appstream-util for verification until we can move to appstream-generator.

vchernin:
> As this PR only touches Arch's pkgbuild and Debian files it seems to me it is fine to merge it. Or am I missing something?

I see no problem merging it, neither Arch or Debian require appstream metadata to be compatible with `appstream-util`. But since RPM distros still require it, we will have to support `appstream-util` in the script and other places for now.

ximion:
> Fedora (and other RPM distros) still use appstream-builder for the AppStream repodata, so we have to use appstream-util for verification until we can move to appstream-generator.

You have no more excuses anymore, now that asgen is C++ code ;-)
Also, appstream-util doesn't validate modern AppStream metadata anymore, and generates outdated metadata collections, so pretty much all modern features will be unavailable on Fedora and only accessible to Flatpak apps over there...

At this point, that's a bit of an issue, I'd say... (there has been only emergency fixes for appstream-glib for more than three years now, it's in very deep maintenance mode)

ximion:
@wwmm @vchernin On Debian, we get this warning about the metadata: https://appstream.debian.org/sid/main/issues/easyeffects.html

Translation data cannot be found because the name in the `translation` tag is incorrect: https://github.com/wwmm/easyeffects/blob/master/src/contents/com.github.wwmm.easyeffects.metainfo.xml.in#L40

The name should likely be `easyeffects` (same as the gettext translation domain). See https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-translation for more information :-)

PR #4240, opened 2025-10-29
Reinstate literal new lines in locale strings

Digitalone1:
I though new lines "`\n`" could be confusing for translators, but in reality they are rendered in a user friendly way in weblate, so we can reinstate them. The only things to hide from translators should be the HTML code, if possible.

### DIFF
--- a/src/contents/ui/PresetsLocalPage.qml
+++ b/src/contents/ui/PresetsLocalPage.qml
@@ -184,7 +184,7 @@ ColumnLayout {
                     id: saveDialog
 
                     title: i18n("Save Preset") // qmllint disable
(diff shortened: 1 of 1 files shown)


wwmm:
> I though new lines "`\n`" could be confusing for translators, but in reality they are rendered in a user friendly way in weblate, so we can reinstate them. The only things to hide from translators should be the HTML code, if possible.

Good to know. I was wondering if weblate was being smart or not about that.

Digitalone1:
> Good to know. I was wondering if weblate was being smart about that.

Luckily yes. Please update the translation templates.

wwmm:
> Luckily yes. Please update the translation templates.

Our github action probably will trigger that in a few seconds.

PR #4263, opened 2025-10-31
Fix some typos

Digitalone1:


Digitalone1:
@wwmm I still see the wrong translation on weblate. Can you push these changes also there? Thanks.

wwmm:
> @wwmm I still see the wrong translation on weblate. Can you push these changes also there? Thanks.

Strange. As far as I can see there is nothing pending there.

PR #4269, opened 2025-11-01
Improve margins of plugin credits

Digitalone1:
@wwmm Since you said it's not good that footer action buttons are too close to the plugin credits, I added an amount of right margin to it.

I also improved a label for the Deesser which was misleading for the translators.


wwmm:
Thanks! As the button inside Kirigami.ActionToolBar are using a flat style the plugin credits felt like it was also part of the Kirigami.ActionToolBar when they were too close.

PR #4284, opened 2025-11-05
Flatpak build QT 6.10 update

vchernin:
Some minor fixes for the build to work for QT 6.10.

Worth noting this also brings a runtime pipewire version upgrade to pipewire 1.4.8.

I did notice the desktop file may not be correct for the "devel" build that flatpak uses. The desktop file does not get exported properly by flatpak and I cannot find the application on the gnome launcher. It may just be a problem with how I'm installing the built `.flatpak` file, I will try some more to investigate when I have time.

The manual also does not open in the flatpak build, except there is not even an error message. Simply nothing happens when clicking the manual button.

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -49,7 +49,7 @@ jobs:
       FLATPAK_BUILD_PATH: flatpak_app/files/share
     needs: prepare
     container:
-      image: ghcr.io/flathub-infra/flatpak-github-actions:kde-6.9
+      image: ghcr.io/flathub-infra/flatpak-github-actions:kde-6.10
       options: --privileged
     strategy:
       matrix:
--- a/util/flatpak/com.github.wwmm.easyeffects.Devel.json
+++ b/util/flatpak/com.github.wwmm.easyeffects.Devel.json
@@ -1,7 +1,7 @@
 {
     "id": "com.github.wwmm.easyeffects.Devel",
     "runtime": "org.kde.Platform",
-    "runtime-version": "6.9",
+    "runtime-version": "6.10",
     "sdk": "org.kde.Sdk",
     "branch": "stable",
     "command": "easyeffects",
--- a/util/flatpak/easyeffects-modules.json
+++ b/util/flatpak/easyeffects-modules.json
@@ -8,8 +8,8 @@
                 {
                     "type": "git",
                     "url": "https://github.com/qt/qtgraphs.git",
-                    "tag": "v6.9.0",
-                    "commit": "ba29aa033b1878429dd1f12d783c57e516f6f32a",
+                    "tag": "v6.10.0",
+                    "commit": "4e18609678b777f7e7ce47c17764e46c20bfb608",
                     "//": "We cannot use x-checker-data here as this version has to match the qt/kde runtime"
                 }
             ],
@@ -45,7 +45,8 @@
             "config-opts": [
                 "-DCMAKE_BUILD_TYPE=Release",
                 "-DBUILD_STATIC_LIBS=OFF",
-                "-DCMAKE_INSTALL_LIBDIR=lib"
+                "-DCMAKE_INSTALL_LIBDIR=lib",
+                "-DCMAKE_POLICY_VERSION_MINIMUM=3.5"
             ],
             "sources": [
                 {
@@ -365,12 +366,8 @@
                 {
                     "type": "git",
                     "url": "https://gitlab.freedesktop.org/pulseaudio/webrtc-audio-processing.git",
-                    "tag": "v2.1",
-                    "commit": "846fe90a289f58b7c9303a635142aa2c7caa93e5",
-                    "x-checker-data": {
-                        "type": "git",
-                        "tag-pattern": "^v([\\d.]+)$"
-                    }
+                    "commit": "bc838790eeb6066d30f019c2a3516bd2b824a5c8",
+                    "//": "use latest commit for gcc 15 support"
                 }
             ]
         }


wwmm:
The class managing the manual is https://github.com/wwmm/easyeffects/blob/master/src/help_manager.cpp. In short the manual now is a collection of html files embedded in our executable. In order to show the manual we copy the embedded files to a local directory and ask `QDesktopServices::openUrl(url)` to open the index.html file. 

Considering there is no warning my guess is that the last step is the one failing. Do we need some kind of permission to ask browsers to open local web pages?

vchernin:
> Considering there is no warning my guess is that the last step is the one failing. Do we need some kind of permission to ask browsers to open local web pages?

Not that I'm aware of. Usually opening a web page is easy and works out of the box with Flatpak. Maybe it is different because it is a bunch of local html files.

wwmm:
> Not that I'm aware of. Usually opening a web page is easy and works out of the box with Flatpak. Maybe it is different because it is a bunch of local html files.

ChatGPT suggests the host browser does not have permission to load files inside the sandbox. What makes sense. But I wonder why there is no warning if that is the case.

It also suggests to call xdg-open directly `QProcess::startDetached("/usr/bin/xdg-open", arguments);`. But I am not sure if this will really work. And there are the usual concerns involved in calling an external executable directly...

wwmm:
@vchernin I've added `xdg-open` as fallback and additional warnings. Let's see if this makes any difference.

vchernin:
> @vchernin I've added `xdg-open` as fallback and additional warnings. Let's see if this makes any difference.

It is an improvement. I now see a dialog asking which browser I want to open the file with.

There is still an issue though. The way the code seems to work is it just opens the singular html file through the xdg desktop filesystem portal. Meaning while I can load the html page, it is without any styling or js or any other referenced files. Also, none of the hyperlinks work. The generated path viewable from apps not in the easyeffects flatpak sandbox will be something like `file:///run/user/1000/doc/170a8fdc/index.html`.

There aren't any warnings relating to this. What is the environment variable to enable debug logging?

The copy operation I think is correct. The issue is xdg-open is not quite the correct tool. In flaptak, a custom xdg-open is used. xdg-open is currently calling this dbus method: https://github.com/flatpak/flatpak-xdg-utils/blob/7c77bb30415aece282a54106fdb8ce3367ddd0b5/src/xdg-open.c#L125-L139 i.e. `OpenFile`. We want `OpenDirectory`. https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.OpenURI.html#org-freedesktop-portal-openuri-opendirectory. The entire directory will then be copied so that the filesystem portal can use it across apps. However, even if that means the files will all be available outside of the easyeffects flatpak sandbox, somehow we still need to then try to open the index.html file that was copied. I don't think asking firefox to open the entire directory will make sense to begin with.

So instead we likely want have to do is to use [Documents portal](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Documents.html) to copy the entire directory, and then use the openuri portal to open the webpage. But I am not sure if it is possible or practical.

Another thing I realized we need to update the ci job for the help page https://wwmm.github.io/easyeffects/. It should be trivial to update the ci file now that they are just plain old html files. If we wanted, the in-app manual button could always direct users to the help website (we could put different versions under sub paths). But supporting a offline manual makes a lot of sense to me.

wwmm:
>There aren't any warnings relating to this. What is the environment variable to enable debug logging?

Now you get logs running EasyEffects as `easyeffects --debug`. No environmental variables for that anymore.

>If we wanted, the in-app manual button could always direct users to the help website (we could put different versions under sub paths). But supporting a offline manual makes a lot of sense to me.

At some point I considered having an online manual. But as adding it offline was easy I did not take the idea ahead. And I also think it makes sense to support an offline manual.

I remember I tried to make a built-in html viewer based on Qt's webkit library but the whole thing failed miserably. The module that worked does not support javascript and the one that does did not work with the widgets we use. I don't know if this changed in the last Qt release.

vchernin:
> >If we wanted, the in-app manual button could always direct users to the help website (we could put different versions under sub paths). But supporting a offline manual makes a lot of sense to me.
> 
> 
> 
> At some point I considered having an online manual. But as adding it offline was easy I did not take the idea ahead. And I also think it makes sense to support an offline manual.
> 
> 
> 
> I remember I tried to make a built-in html viewer based on Qt's webkit library but the whole thing failed miserably. The module that worked does not support javascript and the one that does did not work with the widgets we use. I don't know if this changed in the last Qt release.

FYI Qt has migrated from WebKit to WebEngine (Chromium). https://doc.qt.io/qt-6/qtwebenginewidgets-qtwebkitportingguide.html. Maybe it works better now.

wwmm:
> FYI Qt has migrated from WebKit to WebEngine (Chromium). https://doc.qt.io/qt-6/qtwebenginewidgets-qtwebkitportingguide.html. Maybe it works better now.

I think that webengine is the one I tried. But instead of the c++ api in the link I tried its qml counterpart. In any case I will see if anything has changed.

wwmm:
@vchernin I was able to use WebEngineView. It should be fine in the flatpak package now. Initially I thought it would not work because EasyEffects was crashing because of it but it was the compiler sanatizer catching Qt bugs. As far as I can manual usability is fine.

wwmm:
Now I am wondering if the Flatpak build already has support for `qt6-webengine` or if it will fail :smile: .

vchernin:
> Now I am wondering if the Flatpak build already has support for `qt6-webengine` or if it will fail 😄 .

I will look into it. Some flatpak apps use qt webengine so it should be possible.

PR #4291, opened 2025-11-06
Update for Ukrainian Localization

Drommer:
Some minor changes.

### DIFF
--- a/po/uk/easyeffects.po
+++ b/po/uk/easyeffects.po
@@ -1,14 +1,14 @@
 # SOME DESCRIPTIVE TITLE.
 # Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
 # This file is distributed under the same license as the easyeffects package.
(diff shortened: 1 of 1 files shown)


Drommer:
I noticed that some strings for Filter and Equalizer have the same meaning but differ by one symbol (for example, “Allpass” and “All-pass”), which means it has to be translated twice.

wwmm:
> I noticed that some strings for Filter and Equalizer have the same meaning but differ by one symbol (for example, “Allpass” and “All-pass”), which means it has to be translated twice.

I agree. There is no need for this. I probably typed "All-pass" by mistake. As far as I can see LSP native window always use "Allpass". I will fix that today.

Drommer:
Also

> Band-pass -> Bandpass
> Hi-pass -> High-pass
> Hi-shelf -> High-shelf
> Ladder-rej -> Ladder-rejection
> Lo-pass -> Low-pass
> Lo-shelf -> Low-shelf

PR #4301, opened 2025-11-07
Devel mode fixes

vchernin:
Adds back support for devel mode. This should be fairly safe, but it is possible I forgot to change a variable somewhere. See mainly this last commit.

This also fixes the metainfo file which was not even being installed for a while.


wwmm:
For some reason after installation the desktop file has the wrong app name:
```
wallace@wwmm ~/myfiles/code/bash $ cat /usr/share/applications/com.github.wwmm.easyeffects.desktop 
[Desktop Entry]
Name=@APPLICATION_NAME@
GenericName=Equalizer, Compressor and Other Audio Effects
Comment=Audio Effects for PipeWire Applications
Keywords=limiter;compressor;reverberation;equalizer;autovolume;
Categories=AudioVideo;Audio;
Exec=easyeffects
Icon=@APPLICATION_ID@
StartupWMClass=@APPLICATION_NAME@
StartupNotify=true
Terminal=false
Type=Application
```

wwmm:
My guess is that
```
install(FILES 
    com.github.wwmm.easyeffects.desktop.in
    DESTINATION ${KDE_INSTALL_APPDIR}
    RENAME ${APPLICATION_ID}.desktop
)
```
is taking the wrong file. The correct file is generated inside the build folder. It probably got the one inside the source folder.

PR #4304, opened 2025-11-08
Some cleanup

vchernin:
A couple commits of random cleanup.

I found a autostart bug introduced with the recent changes in master. I first noticed it in the flatpak build, it crashes here when toggling the service mode switch in preferences. This also happens with the arch build if you comment out the non-libportal fallback code.

```bash
Core was generated by `/app/bin/easyeffects --debug --quit'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x000056032adf33d2 in Autostart::update_background_portal (this=0x0) at /run/build/easyeffects/src/autostart.cpp:109
109	  auto xdp_parent = xdp_parent_new_qt(window);
[Current thread is 1 (Thread 0x7f14662d6a40 (LWP 2))]
```

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -60,7 +60,7 @@
 #include "test_signals.hpp"
 #include "util.hpp"
 
-#if defined(__GLIBC__)
+#ifdef __GLIBC__
 #include <malloc.h>
 #endif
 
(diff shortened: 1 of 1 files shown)


wwmm:
> I found a autostart bug introduced with the recent changes in master. I first noticed it in the flatpak build, it crashes here when toggling the service mode switch in preferences. This also happens with the arch build if you comment out the non-libportal fallback code.

For some reason calling `update_background_portal()` through the lambda function is causing an "use after free". If I put update_background_portal() directly inside the qt slot there is no crash.

wwmm:
Master branch updated. It is fine for me. Let me know if additional surprises happen when running inside flatpak.

PR #4305, opened 2025-11-08
Update for Ukrainian Localization

Drommer:
I propose to correct:
> Gain boost -> Threshold Boost
> Automatic level -> Automatic Level _(both exist now)_
> LPF -> Low-pass _(you use this option in all other LSP plugins)_

### DIFF
--- a/po/uk/easyeffects.po
+++ b/po/uk/easyeffects.po
@@ -1679,7 +1679,7 @@ msgstr "Ефекти"
 
 #: src/contents/ui/PageStreamsEffects.qml:637
 msgid "Input monitoring"
-msgstr "Відстеження входу"
+msgstr "Прослуховування входу"
 
 #: src/contents/ui/PageStreamsEffects.qml:650
 msgid "Excluded apps"


Digitalone1:
You can also do it on [weblate](https://hosted.weblate.org/engage/easyeffects/). It's more straightforward.

Digitalone1:
@Drommer LPF is used as acronym by the native LSP window, so we followed it.

Drommer:
> @Drommer LPF is used as acronym by the native LSP window, so we followed it.

This abbreviation is used only once in the EasyEffects code. It appears in the Multiband Compressor and Multiband Gate. The term "low-pass" is used for other LSP plugins, even though "LPF" is used by the native window (for example, a single-band compressor and gate). Similar terms are also used by other instruments, such as "high-pass." Therefore, it makes sense to use just one term.

> You can also do it on [weblate](https://hosted.weblate.org/engage/easyeffects/). It's more straightforward.

I know. But I find it easier to edit in Notepad. One of the users on Weblate likes to mess up translations. It's not very convenient to go through all the lines every time and look for where he damaged the translation. The second pull request by him right now. I'll have to fix it again.

Issue #4309, opened 2025-11-09
Improve metadata for Flathub


vchernin:
I noticed Flathub has a few minor complaints about Easy Effects. Our issues seem to be ones originating from the metainfo file.

<details>
  <summary>Full app listing quality data</summary>

<img width="996" height="1998" alt="Image" src="https://github.com/user-attachments/assets/d58ef9b6-9ac8-43ef-a3fd-3937293daa3b" />
</details>

- [ ] [Fix primary brand colors](https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines/quality-guidelines#has-primary-brand-colors)
- [ ] [Decrease length of summary](https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines/quality-guidelines#not-too-long-1)
- [ ] [Use sentence case not title case for summary](https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines/quality-guidelines#no-weird-formatting-1)

There is some debate to be had as to what color and summary should be used.

You can [see the banner preview and try things out here](https://flathub.org/en/apps/com.github.wwmm.easyeffects/bannerpreview) (for now the old screenshots may still be visible, ignore that).

Maybe for the summary `Audio effects for any app` is better? PipeWire has long had good compatibility with non-PipeWire clients, so really it seems misleading at this point to say it is only for PipeWire applications. But of course, Easy Effects will not work at all without PipeWire as the sound daemon. Regardless Flathub says to avoid putting technical things in the summary, and instead put those in less visible spots.

Digitalone1:
`Audio effects for multimedia applications` should be enough I think.

wwmm:
>There is some debate to be had as to what color and summary should be used.

I don't have a preference. But color of what would this be?

vchernin:
> `Audio effects for multimedia applications` should be enough I think.

Flathub suggests a shorter one. (10-25 characters). Of course if we want it is fine to ignore their recommendation. Personally I think the `multimedia` part is a bit redundant since it already says `Audio`. Maybe `Application audio effects`?

> I don't have a preference. But color of what would this be?

This is for the [brand background color](https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines/quality-guidelines#good-brand-colors).

You can see examples on the flathub home page: https://flathub.org/en

Easy Effects currently has a default pink background it seems: https://flathub.org/en/apps/com.github.wwmm.easyeffects/bannerpreview

[...]

PR #4321, opened 2025-11-10
Metainfo and tray icon fixes

vchernin:
Fixes https://github.com/wwmm/easyeffects/issues/4309.

Adds some branding colors and also changes the summary text everywhere. I am not 100% sold on the summary text change, so if it feels wrong to you it can be changed back. I was mostly trying to follow Flathub's recommendations, as well as just trying to make it as concise as possible.

Now that the tray icon is actually supported, unfortunately we need to add permission for it in Flatpak as the implementation is not super secure.

### DIFF
--- a/util/flatpak/com.github.wwmm.easyeffects.Devel.json
+++ b/util/flatpak/com.github.wwmm.easyeffects.Devel.json
@@ -11,6 +11,7 @@
         "--share=ipc",
         "--socket=fallback-x11",
         "--env=QTWEBENGINEPROCESS_PATH=/app/bin/QtWebEngineProcess",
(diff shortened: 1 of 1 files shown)


wwmm:
> I am not 100% sold on the summary text change, so if it feels wrong to you it can be changed back

"Simple audio effects" seems a little strange indeed because some of the effects we provide are not exactly simple. For now I will rename it to "Global audio effects". It is still short and in a way adding global effects for the desktop is the idea. 



> Now that the tray icon is actually supported, unfortunately we need to add permission for it in Flatpak as the implementation is not super secure.

Oh... So we do need permission.

PR #4324, opened 2025-11-10
Translation for units of measurement

Drommer:
I added translations for units of measurement and one missing line.


Digitalone1:
@wwmm I don't see these ones in weblate yet. Maybe we need to update the translation templates.

PR #4326, opened 2025-11-10
Fix spacing in Filter page

CarlSchwan:


CarlSchwan:
Not tested as I am getting the following error when building:

```
/home/carl/projects/easyeffects/src/echo_canceller.hpp:22:10: fatal error: api/scoped_refptr.h: No such file or directory
   22 | #include <api/scoped_refptr.h>
```

ramdomPTM:
This header is provided by libwebrtc-audio-processing devel package.

CarlSchwan:
Yeah, looks like I'm hit by https://bugzilla.redhat.com/show_bug.cgi?id=2404671 Let see if I can compile the new version myself

CarlSchwan:
Okay now that I can compile, this will need more changes :)

This is the look in term of spacing, I want to change to:

Screenshot_20251110_142407

If you are happy with that, I'm happy to apply the same transformation to all the effects

wwmm:
>If you are happy with that, I'm happy to apply the same transformation to all the effects

If you can please do it :-). There was so much we had to do that details like this had to be left for later.

For some reason I did not understand github saw a conflict. Let's hope I did not mess with the spaces while fixing it.

PR #4333, opened 2025-11-10
Add some vertical margin to the segmented button in the footer

CarlSchwan:
Before:

image

After:

Screenshot_20251110_143844

Spacing could be increased a bit more but not 100% sure

Btw if you have any opinion on the design of the segmented button or how to evolve then a bit, I'm taking them. I originally designed then for the date picker with the hope to use them in more places but atm it's only used there and now in Easy Effects.

### DIFF
--- a/src/contents/ui/PageStreamsEffects.qml
+++ b/src/contents/ui/PageStreamsEffects.qml
@@ -489,6 +489,8 @@ Kirigami.Page {
         position: Controls.ToolBar.Footer
 
         contentItem: RowLayout {
+            spacing: Kirigami.Units.smallSpacing
+
             anchors {
                 left: parent.left
                 leftMargin: Kirigami.Units.smallSpacing
@@ -599,6 +601,8 @@ Kirigami.Page {
                 id: segmentedButton
 
                 Layout.alignment: Qt.AlignHCenter
+                Layout.topMargin: Math.round(Kirigami.Units.smallSpacing / 2)
+                Layout.bottomMargin: Math.round(Kirigami.Units.smallSpacing / 2)
 
                 readonly property bool hasEnoughWidth: appWindow.width >= Kirigami.Units.gridUnit * 40
 


vchernin:
It is nice to see experts come here :)

@wwmm could you change the repository settings to allow workflows to run for new contributors? It is a bit annoying when they don't run.

https://github.com/wwmm/easyeffects/settings/actions and set `Approval for running fork pull request workflows from contributors` to `Require approval for first-time contributors who are new to GitHub`. The workflows are all reasonably secure and isolated, they just build and run things mostly on their own. Fork workflows should not be able to do anything mildly dangerous like publish the manual.

wwmm:
> It is nice to see experts come here :)
> 
> @wwmm could you change the repository settings to allow workflows to run for new contributors? It is a bit annoying when they don't run.
> 
> https://github.com/wwmm/easyeffects/settings/actions and set `Approval for running fork pull request workflows from contributors` to `Require approval for first-time contributors who are new to GitHub`. The workflows are all reasonably secure and isolated, they just build and run things mostly on their own. Fork workflows should not be able to do anything mildly dangerous like publish the manual.

Done. This has been annoying me for some time but it id not occur to me I could change the setting :smile:

wwmm:
>Spacing could be increased a bit more but not 100% sure

At some point I tried to insert a space there. But then I realized that inserting space in the bottom bar made the space also go to the header bar and vice versa. So I decided to let the whole thing for later. It may be because I used `Kirigami.AbstractApplicationHeader` to create the bottom bar. But not using it would require much more work...

>Btw if you have any opinion on the design of the segmented button or how to evolve then a bit, I'm taking them. I originally designed then for the date picker with the hope to use them in more places but atm it's only used there and now in Easy Effects.

I lack imagination when ti comes to design or style... So I think I will just explain why I used the segmented button. Initially I put there a `Kirigami.ActionToolBar`. It worked but as time went by I felt those buttons needed to be visually distinct from the ones inside the Kirigami.ActionToolBar we use on other places.

These buttons select major sections of our interface. And they are not independent. I wanted something that would give some feeling that it was a group of buttons working together. The Kirigami.ActionToolBar style does not give that vibe. 

That being said as soon as I used it I missed the Kirigami.ActionToolBar feature where it automatically switch to  icon only based on the available horizontal space. I think there is not need for an overflow menu in this case but having the segmented button switch modes automatically would be nice. I put my own workaround to force that but I am not sure the approach is the best.

wwmm:
>But then I realized that inserting space in the bottom bar made the space also go to the header bar and vice versa.

The same happened after this patch. What isn't a bad thing. It just gives that feeling that using `Kirigami.AbstractApplicationHeader` in the bottom bar may not be a good idea because a space set to the bottom propagating to the header does not seem intended.

PR #4343, opened 2025-11-11
cmake: Use imported targets for pkgconfig dependencies

CarlSchwan:


wwmm:
Thanks! I was not aware of IMPORTED_TARGET existence.

PR #4344, opened 2025-11-11
Rework MenuAddPlugins

CarlSchwan:
- Add keyboard navigation
- Use a plain Controls.Dialog
- Fix size

image

### DIFF
--- a/src/contents/ui/MenuAddPlugins.qml
+++ b/src/contents/ui/MenuAddPlugins.qml
@@ -6,31 +6,35 @@ import "Common.js" as Common
 import ee.tags.plugin.name as TagsPluginName
 import org.kde.kirigami as Kirigami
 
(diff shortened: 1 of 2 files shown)


wwmm:
Along the way I spent quite a lot of time thinking about whether to use the overlay sheet or a dialog in these cases. After not getting to a conclusion I just used the overlay sheet and moved on.

Setting the height was also a dilemma in some situations. Does
```
implicitHeight: Math.min(Kirigami.Units.gridUnit * 40, Math.round(Controls.ApplicationWindow.window.height * 0.8))
```
come from KDE HIG?

Pointedstick:
FWIW, the KDE HIG has the following to say on the topic: https://develop.kde.org/hig/displaying_content/#page-vs-dialog-vs-overlaysheet

PR #4350, opened 2025-11-11
Convolver UI: avoid leading spaces for translated units

Digitalone1:
@wwmm I made this with the smartphone. Can you please test it and apply QML format? Thanks.


Digitalone1:
Oh, maybe `qmllint disable` was not needed. Please remove it if necessary.

wwmm:
> Oh, maybe `qmllint disable` was not needed. Please remove it if necessary.

The linter does not understand the i18n. So it is still needed if we set `UnqualifiedAccess = true`. At this moment we have it disabled because the github CI shows many false positives. But every now and then I enable it here on my computer to have an idea if we have many unqualified access.

Issue #4332, opened 2025-11-10
New UI doesnt follow qt6ct themes


fabioraitz:
### EasyEffects Version

8.0.0

### What package are you using?

Flatpak (Flathub)

### Distribution

Linux Mint 22.2

### Describe the bug

The new 8.0 version, with QT, wont follow qt6ct theming on linux mint.

wwmm:
An important detail is that we use Qt, qml and kirigami. We do not force any style on our side. What you are seeing may be a limitation of qml or kirigami.

HessDavid:
I'm having the same issue. 
It's the only QT app that seems to ignore the colors set by qt5ct and qt6ct.
It does use the correct font set by qt5ct/qt6ct tho.

Shorkerrr:
Also got same issue as above

Digitalone1:
> I'm having the same issue. It's the only QT app that seems to ignore the colors set by qt5ct and qt6ct. It does use the correct font set by qt5ct/qt6ct tho.

Do you use other Kirigami apps that follow qt6ct?

HessDavid:
> Do you use other Kirigami apps that follow qt6ct?

I normally don't, but I just tried to install another one (Angelfish) and it seems to follow the QT theme set by qt5ct/qt6ct

Shorkerrr:
> > Do you use other Kirigami apps that follow qt6ct?
> 
> I normally don't, but I just tried to install another one (Angelfish) and it seems to follow the QT theme set by qt5ct/qt6ct

Can confirm that this phenomena is happening on my system too, aswell as the KDE systemsettings app. all but easyeffects seem to theme just fine

Digitalone1:
> I normally don't, but I just tried to install another one (Angelfish) and it seems to follow the QT theme set by qt5ct/qt6ct

Then we are missing something. @wwmm will try to figure out what.

wwmm:
> Then we are missing something. [@wwmm](https://github.com/wwmm) will try to figure out what.

It is possible. At some point during the year it took to finish the Qt port there was one occasion where Kirigami  changed how the app should setup theme configuration. Maybe they changed this again and we do not know.

I will try to do some tests with qt6ct. But the fact KDE developers do not recommended it to be used inside KDE https://discuss.kde.org/t/how-to-get-kate-to-respect-qt6ct-window-color-theme-again/28948/8 because it can break things is a little concerning...

CarlSchwan:
[...]

PR #4353, opened 2025-11-11
Dont hide the text for Effects button

dphaldes:
- Made the button text of effects text visible, since it conveys the functionality better. 
- Also added text based on checked/active state

After 
image image

### DIFF
--- a/src/contents/ui/main.qml
+++ b/src/contents/ui/main.qml
@@ -436,10 +436,9 @@ Kirigami.ApplicationWindow {
                 overflowIconName: "overflow-menu-left"
                 actions: [
                     Kirigami.Action {
-                        text: i18n("Turn effects on/off") // qmllint disable
+                        text: checked === true ? i18n("Active") : i18n("Inactive")// qmllint disable
                         icon.name: "system-shutdown-symbolic"
                         icon.color: checked === true ? Kirigami.Theme.positiveTextColor : Kirigami.Theme.negativeTextColor
-                        displayHint: Kirigami.DisplayHint.IconOnly
                         checkable: true
                         checked: !DB.Manager.main.bypass
                         onTriggered: {


Pointedstick:
In general, checkable buttons shouldn't change their text depending on whether or not they're checked. This can make it unclear what pressing the button will do.

If you want to change the text depending on the state, the button shouldn't be checkable.

Looks like this case may not be mentioned in https://develop.kde.org/hig/getting_input/ or https://develop.kde.org/hig/text_and_labels/. I'll look into that.

Digitalone1:
> In general, checkable buttons shouldn't change their text depending on whether or not they're checked. This can make it unclear what pressing the button will do.

I agree. At this point, if the label should remain visible, I would suggest only `Effects`. 

I think the `Turn effects on/off` was more for showing the tooltip to the user when the label was hidden. I'd like to show the tooltip anyway and maybe we can implement it later.

Drommer:
What about "Bypass"? This is a general term used by most plug-ins.

wwmm:
> What about "Bypass"? This is a general term used by most plug-ins.

We had a discussing in this direction a few years ago. Mostly the issue was not having an icon that clearly conveyed the bypass idea. As the current button feels more like turn on/off  the word bypass feels strange. But maybe now there is a good bypass icon. I do not know.

dphaldes:
> I agree. At this point, if the label should remain visible, I would suggest only Effects.

I have a started a discussion about this on KDE VDG channel. The problem is that Color Highlights are not sufficient to convey button states, especially for Qt apps since they are themeable and thus raises a UX issues. 
While the HIG doesn't say anything about this, as Nate said, we already use similar mechanism in plasma already (and mobile). 
Changing icons and text is also part of several design languages like Google's Material and Microsoft's Fluent.

gwuen:
> The problem is that Color Highlights are not sufficient to convey button states, especially for Qt apps since they are themeable and thus raises a UX issues.

I agree. The contrast is somehow even worse on GNOME, even though the theme should be the same as on KDE:

light dark

Nothing besides the green icon makes it clear the button is toggled 'on'. The outline could very well be part of the general design instead of indicating the button's state, same for its slightly different color. The GTK version had better contrast in this regard.

So as someone with trouble telling red and green apart, it's not easy recognizing the current state at a glance. Having a bigger, more visible button somewhat alleviates this problem, but I would much prefer something like a [switch](https://doc.qt.io/qt-6/qtquickcontrols-buttons.html#switch-control).

Digitalone1:
There wasn't the label and you wanted it. We added it, but now you don't like the button because is toggled by default. So we maybe will change it to a switch (which is not possible since that is an action toolbar) and some other one will come here saying they want another thing. We can't please everyone. This is a Kirigami app and you know it's not the best on Gnome. At some point you need to make do with what you have in life.

gwuen:
> There wasn't the label and you wanted it. We added it, but now you don't like the button because is toggled by default. So we maybe will change it to a switch (which is not possible since that is an action toolbar) and some other one will come here saying they want another thing. We can't please everyone.

For one, I think you're mixing multiple people up here. I wasn't the one who requested the label in the first place, and also I don't have any problem with it being toggled on by default.

> This is a Kirigami app and you know it's not the best on Gnome. At some point you need to make do with what you have in life.

This is *not* about lacking integration or aesthetics. My issue is about accessibility and usability of the app. I won't complain about the toolkit the app is using as long as it doesn't cause troubles.

Really didn't expect such a rude answer from you, I was never disrespectful to any of you in any way, and I'm doing my best to help out to the best of my abilities when pointing out issues.

Digitalone1:
@gwuen It's not a rude comment. We already made a lot to make the experience better on Gnome. I know the libadwaita version looked better on Gnome, but we can't pretend more by a Kirigami app made primarily for KDE.

So, no, I don't think think we will change the check buttons in switch buttons. Sorry.

gwuen:
> I know the libadwaita version looked better on Gnome, but we can't pretend more by a Kirigami app made primarily for KDE.

I don't know why you think this is about aesthetics. As you can see in my screenshots, the contrast is evidently worse on GNOME. I don't know what's causing this, I'm not theming anything, but I would also consider this an issue if it looked this way on KDE.

This is simply about accessibility of the app. I would much prefer if it looked exactly the way it did on KDE.

wwmm:
I know this whole thing can be tiring but let's try to calm down. Ideally I think that instead of a switch or different labels it would be better to have an icon that becomes slightly different when in the off state. Mabe creating a custom icon based on the current one for that is viable.

Digitalone1:
> I know this whole thing can be tiring but let's try to calm down. Ideally I think that instead of a switch or different labels it would be better to have an icon that becomes slightly different when in the off state. Mabe creating a custom icon based on the current one for that is viable.

The problem is not the icon, he says "the contrast is evidently worse on GNOME" because "it's not easy recognizing the current state at a glance". The check buttons are not rendered in a best way on Gnome, but there's nothing we can do.

Digitalone1:
> I don't know why you think this is about aesthetics. As you can see in my screenshots, the contrast is evidently worse on GNOME. I don't know what's causing this, I'm not theming anything, but I would also consider this an issue if it looked this way on KDE.
> 
> This is simply about accessibility of the app. I would much prefer if it looked exactly the way it did on KDE.

I understood what you are saying. I'm telling you that, maybe, there's no fix for this issue. Or, maybe, what you think will fix it won't please other users.

gwuen:
> The problem is not the icon, he says "the contrast is evidently worse on GNOME" because "it's not easy recognizing the current state at a glance".

No, I think @wwmm understood me correctly. As @dphaldes already explained:

> The problem is that Color Highlights are not sufficient to convey button states, especially for Qt apps since they are themeable and thus raises a UX issues.

So if the button's color highlight can't reliably convey its state, an alternative would be to have different icons for different states. I think that's a good solution, no matter the theme or platform.

---

> The check buttons are not rendered in a best way on Gnome, but there's nothing we can do.

I figured out this has nothing to do with GNOME, but rather not having the [Breeze theme](https://github.com/KDE/breeze) installed. This line

https://github.com/wwmm/easyeffects/blob/5203af4841a9c94e06a4776fd25242ceef6e1c35/src/main.cpp#L321

only forces Breeze if it's present on the system, but falls back to the built-in Qt theme and doesn't report an error otherwise. This is why the app might look different on non-KDE systems. A quick fix could be to list Breeze as a hard dependency, but I don't know if it should be considered one. Especially considering people already expressed interest in [changing the UI theme](https://github.com/wwmm/easyeffects/issues/4332), so this is likely not the best solution.

---

> I'm telling you that, maybe, there's no fix for this issue. Or, maybe, what you think will fix it won't please other users.

Well, the thing is, I don't consider accessibility some personal preference. If I'm not mistaken, KDE [strives to be accessible](https://develop.kde.org/hig/accessibility/) as much as GNOME, though the means to the end may be slightly different.

But I don't want to dwell on this point, because I think we have already found a good solution by simply using distinct icons. Cheers!

Digitalone1:
> So if the button's color highlight can't reliably convey its state, an alternative would be to have different icons for different states. I think that's a good solution, no matter the theme or platform.

The Effect button is not the only check button we have. We have plenty of check buttons throughout the plugins UI. We can't use different icons for all of them.
 
> I figured out this has nothing to do with GNOME, but rather not having the [Breeze theme](https://github.com/KDE/breeze) installed. 

Did you install the breeze theme and tried it in Easy Effects? We could suggest this option to other users.

gwuen:
> We have plenty of check buttons throughout the plugins UI.

Are you sure? All the other controls I found look fine, the current state is always clear in some way:

screenshot

Either way, having more potentially problematic buttons shouldn't keep us from fixing the most prominent example.

> Did you install the breeze theme and tried it in Easy Effects? We could suggest this option to other users.

Yes, tested on Arch Linux, this works.

Digitalone1:
> Are you sure? All the other controls I found look fine, the current state is always clear in some way:

See the check buttons in the footer of some plugins or the mute/bypass buttons of some other ones.

PR #4354, opened 2025-11-11
Respect user-defined Qt style instead of forcing Breeze (fixes #4332)

### LINKED ISSUE #4332: New UI doesnt follow qt6ct themes

Lunix-420:
This PR removes the hardcoded `QApplication::setStyle("breeze")` call and updates the Quick Controls style initialization logic.
Previously, EasyEffects forced the Breeze widget style regardless of user or system configuration, which broke theming under qt6ct. (see #4332)

The new logic:
- Keeps breeze as fallback icon theme only (no forced widget style).
- applies `org.kde.desktop` only if no `QT_QUICK_CONTROLS_STYLE` is set and `QQuickStyle::name()` is empty.

Allows user-configured system QT themes to take effect properly.

There still seem to be a few of icons that don’t theme correctly, but I’m looking into that. Also, I’m not sure if this has any negative impact on systems that don’t have a system theme as I can't really test that on my machine.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -300,17 +300,10 @@ int main(int argc, char* argv[]) {
 
   KColorSchemeManager::instance();
 
-  // Keep system widget colors
-  if (qEnvironmentVariableIsEmpty("QT_STYLE_OVERRIDE"))
-   QApplication::setStyle(QStyleFactory::create(QApplication::style()->objectName()));
- 
-  // Force icons to breeze regardless of color theme
-  QIcon::setThemeName(QStringLiteral("breeze"));
-  QIcon::setFallbackThemeName(QStringLiteral("breeze"));
- 
-  // Keep org.kde.desktop for QQC2 style, but don’t override user env
-  if (qEnvironmentVariableIsEmpty("QT_QUICK_CONTROLS_STYLE") && QQuickStyle::name().isEmpty())
-   QQuickStyle::setStyle(QStringLiteral("org.kde.desktop"));
+  QIcon::setFallbackThemeName(QStringLiteral("breeze")); 
+  if (qEnvironmentVariableIsEmpty("QT_QUICK_CONTROLS_STYLE") && QQuickStyle::name().isEmpty()) { 
+   QQuickStyle::setStyle(QStringLiteral("org.kde.desktop")); 
+  }
 
   // Parsing command line options
 


Digitalone1:
Sorry, but for me this is not good.

I tried these changes on Gnome 49 and this is the result:

<p>Schermata del 2025-11-11 15-04-18</p>

While this is the original one we always had:

<p>image</p>

This is just bad, I can't even see some buttons since I can't distinguish them from the background. Take a look at the plugin list, the menu button for bypass and remove controls are barely visible in the first screenshot. Icons for action buttons are not visible either.

Maybe we should force the Breeze style if we are outside KDE/Qt environments? 

Update. This is the Autogain comparison:

<p>Schermata del 2025-11-11 15-21-56</p>

<p>image</p>

And the footer:

<p>image</p>

<p>image</p>

wwmm:
I am looking at @Digitalone1 images and I have to agree that on gnome this is a downgrade. It is clear that we will need the guidance of a Kirigami developer on this. There is probably a reason why Kirigami's tutorial forces the Breeze theme.

HessDavid:
Yeah, something is not quite right with the icons on this. The weird thing to me is, that it doesn't seem to affect all icons.

> Maybe we should force the Breeze style if we are outside KDE/Qt environments?

I don't think so, forcing the breeze style with is what causes it to ignore the system theme. I think ideally, it should force the breeze icons, without forcing the entire breeze style, but not sure.

wwmm:
Right now I am talking to pipewire's developers on their matrix channel trying to find a good way to implement preset autoloading. So I am not sure about when I will be able to look at this theme situation.

Digitalone1:
> I don't think so, forcing the breeze style with is what causes it to ignore the system theme. I think ideally, it should force the breeze icons, without forcing the entire breeze style, but not sure.

Yes, but we don't live in an ideal world unfortunately, so since the system theme is not good outside KDE, forcing the Breeze icon only outside KDE does not seem a bad idea. Which desktop environment do you use?

HessDavid:
> Which desktop environment do you use?

I'm using Hyprland.

Digitalone1:
> I'm using Hyprland.

Okay, you are outside KDE and you use qt6ct, but not all users install qt6ct to use a specific theme. 

Besides I use qBittorrent as Qt6 app and it is good without qt6ct, so I think they force their own theme.

HessDavid:
> Okay, you are outside KDE and you use qt6ct, but not all users install qt6ct to use a specific theme.

I don't know much about KDE, but I was under the impression that almost all qt6ct users are on something other than KDE. From what I understand, KDE has its own first-party system for managing Qt themes and actually discourages the use of qt6ct. Outside of KDE though, on most enviroments qt6ct is basically the only reliable way to theme Qt apps.

> Besides I use qBittorrent as Qt6 app and it is good without qt6ct, so I think they force their own theme.

I don’t think qBittorrent forces any specific style by default. It will pick up the qt6ct theme if present, but it also has its own internal theming system that can override the system style when enabled.

Digitalone1:
> I don't know much about KDE, but I was under the impression that almost all qt6ct users are on something other than KDE. From what I understand, KDE has its own first-party system for managing Qt themes and actually discourages the use of qt6ct. Outside of KDE though, on most enviroments qt6ct is basically the only reliable way to theme Qt apps.

I suppose most of the people don't have qt6ct installed. So I think we have to care about how they see the application on the first launch without forcing them to install qt6ct. In this case, if we force the Breeze theme, I don't think it's worse of showing a bad user interface.

I'm afraid we could resolve this issue diversifying between desktop environments, just like we did for the XDG Global Shortcuts.   

> I don’t think qBittorrent forces any specific style by default. It will pick up the qt6ct theme if present, but it also has its own internal theming system that can override the system style when enabled.

I don't know, but I remember how horrible it was some years ago, so they made something to improve its look under Gnome without qt6ct.

HessDavid:
> I suppose most of the people don't have qt6ct installed. So I think we have to care about how they see the application on the first launch without forcing them to install qt6ct. In this case, if we force the Breeze theme, I don't think it's worse of showing a bad user interface.

Yeah, I totally get that, seeing a messy UI on first launch is a bigger problem than the app not following the theme. Maybe we could check if `QT_QPA_PLATFORMTHEME` is set to `qt5ct` or `qt6ct` and only force Breeze if it’s not? I think that should basically force Breeze for everyone not using qt6ct for theming, while still letting users change it if they do use qt6ct. I assume nobody sets that environment variable without actually using it.
﻿
Also, I think there’s a small misunderstanding. It’s not that all icons look fine under qt5ct/qt6ct. It’s just that ~3 miscolored icons felt like such a big improvement over everything being completely out of place that I didn’t notice it at first, only after the mention of the GNOME situation did I realize it. So there is definitely some issue here.
﻿
> I don't know, but I remember how horrible it was some years ago, so they made something to improve its look under Gnome without qt6ct.


I only switched to Linux about two years ago, so I can’t really comment on the older situation. But I can say that right now qBittorrent looks way better than I remember it from when I was using Windows.

Digitalone1:
> Yeah, I totally get that, seeing a messy UI on first launch is a bigger problem than the app not following the theme. Maybe we could check if `QT_QPA_PLATFORMTHEME` is set to `qt5ct` or `qt6ct` and only force Breeze if it’s not? I think that should basically force Breeze for everyone not using qt6ct for theming, while still letting users change it if they do use qt6ct.

Yes, it does not seem a bad idea to me.

wwmm:
>Yeah, I totally get that, seeing a messy UI on first launch is a bigger problem than the app not following the theme. Maybe we could check if QT_QPA_PLATFORMTHEME is set to qt5ct or qt6ct and only force Breeze if it’s not? I think that should basically force Breeze for everyone not using qt6ct for theming, while still letting users change it if they do use qt6ct.

I do not have any problem with this workaround. But since we are trying to "better follow Kirigami's guidelines" it would be good to also understand why the procedure in its tutorial is not working well in this case.

vchernin:
From trying to understand https://github.com/wwmm/easyeffects/issues/4460 I learned that indeed Breeze is the only theme that is regularly tested with Kirigami, and so for an app as complex as Easy Effects we really should go with the safe option by default. So option 1, which is the default now, is to use Breeze without modifying anything to support other tools. So I think so long as the default is very clearly Breeze we can try to do this.

> Yeah, I totally get that, seeing a messy UI on first launch is a bigger problem than the app not following the theme. Maybe we could check if `QT_QPA_PLATFORMTHEME` is set to `qt5ct` or `qt6ct` and only force Breeze if it’s not? I think that should basically force Breeze for everyone not using qt6ct for theming, while still letting users change it if they do use qt6ct. I assume nobody sets that environment variable without actually using it.

I think this makes a lot of sense, but I see two problems.

1. Why does it need to check for `qt5ct` _or_ `qt6ct`? Surely since this is a Qt6 app we should only care if `QT_QPA_PLATFORMTHEME` is set to `qt6ct`? Can `QT_QPA_PLATFORMTHEME` simultaneously be trying to configure for both Qt5 and Qt6? What if someone has a theming setup for Qt5 and has set `QT_QPA_PLATFORMTHEME=qt5ct`, but has no decent theme setup for Qt6?
2. To avoid confusion with trying to parse and understand another environment variable, would a new preference option be better? Something in Preferences under `Style`, maybe titled `Use default Breeze theme` which is on by default, with the subtext `By default Breeze is used to provide the look and feel of the application. Disable to allow complete customization with tools like qt6ct. You must restart the application for this setting to take effect.` It can even be in experimental options considering it is unclear how well qt6ct works for Easy Effects even under ideal circumstances.

wwmm:
>It can even be in experimental options considering it is unclear how well qt6ct works for Easy Effects even under ideal circumstances.

Probably. Specially considering that this kind of settings requires an app restart. What right now has to be done manually by the user.

Digitalone1:
I'm not against forcing the Breeze theme. If you decide this way, it's okay and I will install it. However, as an Arch user, I have to say that installing half of KDE packages only for breeze is not the best.

wwmm:
> I'm not against forcing the Breeze theme. If you decide this way, it's okay and I will install it. However, as an Arch user, I have to say that installing half of KDE packages only for breeze is not the best.

Unless I misunderstood something you would not have to do that. The idea would be to just put the call `QApplication::setStyle(QStringLiteral("breeze"));` behind a configuration option. This way people that want to use qt6ct would have a chance to do so after restarting EE.

Digitalone1:
> Unless I misunderstood something you would not have to do that. The idea would be to just put the call `QApplication::setStyle(QStringLiteral("breeze"));` behind a configuration option. This way people that want to use qt6ct would have a chance to do so after restarting EE.

`breeze` has been added to dependencies here https://github.com/wwmm/easyeffects/blob/master/PKGBUILD_AUR#L20

wwmm:
> > Unless I misunderstood something you would not have to do that. The idea would be to just put the call `QApplication::setStyle(QStringLiteral("breeze"));` behind a configuration option. This way people that want to use qt6ct would have a chance to do so after restarting EE.
> 
> `breeze` has been added to dependencies here https://github.com/wwmm/easyeffects/blob/master/PKGBUILD_AUR#L20

Oh... For now let's remove it from there.

wwmm:
I talked today to Kirigami devs in their matrix channel and the suggestion was not calling `QApplication::setStyle(QStringLiteral("breeze"));`. As we know this may not be the best on GNOME my suggestion is doing what this PR is doing but with the extra addition of a configuration option that allows the user to force `QApplication::setStyle(QStringLiteral("breeze"))` if necessary. What do you think?

wwmm:
> I talked today to Kirigami devs in their matrix channel and the suggestion was not calling `QApplication::setStyle(QStringLiteral("breeze"));`. As we know this may not be the best on GNOME my suggestion is doing what this PR is doing but with the extra addition of a configuration option that allows the user to force `QApplication::setStyle(QStringLiteral("breeze"))` if necessary. What do you think?

Or doing the other what around and force by default. And implementing an option that allows this to be disabled.

Digitalone1:
It's very hard to decide. @wwmm do what you think it's the best option. As long as I have a configuration to make it look better, I'll use it.

wwmm:
> It's very hard to decide. @wwmm do what you think it's the best option. As long as I have a configuration to make it look better, I'll use it.

Initially I was inclined to not do the call by default. But after reading

>Yeah, I totally get that, seeing a messy UI on first launch is a bigger problem than the app not following the theme. 

again I think it may be better to put on the people that feel the need to switch themes the burden of using the configuration setting and restarting EasyEffects.

vchernin:
> > `breeze` has been added to dependencies here https://github.com/wwmm/easyeffects/blob/master/PKGBUILD_AUR#L20
> 
> 
> 
> Oh... For now let's remove it from there.

If we have users not installing the breeze theme we will get more bugs like this https://github.com/wwmm/easyeffects/issues/4460. The kirigami developers were fairly clear that breeze is the only theme that was regularly tested. Can we maybe make `breeze` an optional dependency? So at least users can uninstall breeze if they really don't want it?

> Or doing the other what around and force by default. And implementing an option that allows this to be disabled.

I think the default should be what we do now, which is to force the breeze theme. It seems too risky to not force it as otherwise things can be inconsistent. If users want to disable the forcing they can do it with a preferences option. 

I did some reading on kde theming, and maybe this situation will get better with "Union" in the future: https://thelibre.news/the-future-of-kde-styling-and-design/. If that happens we can revisit this.

wwmm:
>Can we maybe make breeze an optional dependency?

I do not see any problem.

wwmm:
Configuration option added. Now it has to be tested by people that use qt6ct (I don't).

wwmm:
> Configuration option added. Now it has to be tested by people that use qt6ct (I don't).

It is in the `Style` section.

wwmm:
As the configuration option is in place I think we can close this PR and keep the discussion at #4332.

PR #4356, opened 2025-11-11
Use a column view for the PipeWierePage

CarlSchwan:
desktop

image

mobile:

image

with https://github.com/wwmm/easyeffects/pull/4355 applied


wwmm:
I see errors like 
```
file:///usr/lib/qt6/qml/org/kde/kirigami/Page.qml:199: Error: Cannot assign [undefined] to QQuickItem*
```
Does this mean I should have waited for the draft to be ready before merging this?

CarlSchwan:
The error is harmless, but I submitted a patch upstream to fix it: https://invent.kde.org/frameworks/kirigami/-/merge_requests/1960

PR #4362, opened 2025-11-11
Use ActionToolBar instead of SegmentedButton

CarlSchwan:
Looks a bit better 

image

### DIFF
--- a/src/contents/ui/PageStreamsEffects.qml
+++ b/src/contents/ui/PageStreamsEffects.qml
@@ -53,7 +53,7 @@ Kirigami.Page {
     MenuAddPlugins {
         id: menuAddPlugins
 
-	    parent: pageStreamsEffects.Controls.Overlay.overlay
+        parent: pageStreamsEffects.Controls.Overlay.overlay
         streamDB: pageStreamsEffects.streamDB
     }
 
@@ -608,7 +608,7 @@ Kirigami.Page {
                 }
             }
 
-            Components.SegmentedButton {
+            Kirigami.ActionToolBar {
                 id: segmentedButton
 
                 Layout.alignment: Qt.AlignHCenter
--- a/src/contents/ui/main.qml
+++ b/src/contents/ui/main.qml
@@ -439,7 +439,7 @@ Kirigami.ApplicationWindow {
                         text: i18n("Go back") // qmllint disable
                         icon.name: "draw-arrow-back-symbolic"
                         displayHint: Kirigami.DisplayHint.IconOnly
-                        visible: (pageStack.currentItem as Kirigami.Page).showBackButton
+                        visible: (pageStack.currentItem as Kirigami.Page)?.showBackButton ?? false
                         onTriggered: {
                             pageStack.currentItem.goBack();
                         }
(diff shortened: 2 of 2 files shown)


dphaldes:
I was working on a similar PR but using [NavigationTabBar](https://api.kde.org/qml-org-kde-kirigami-navigationtabbar.html) since that looks slightly nicer and is what the HIG mentions, albeit takes more vertical space. 

What do you think? I think HIG should also mention action toolbar since this implementation looks quite functional

CarlSchwan:
> What do you think? I think HIG should also mention action toolbar since this implementation looks quite functional

I don't think there is enough space for the navigation tab bar here, that component is very handy on mobile. But on desktop unless there is a lot of space available, I would recommend against using it.

Ultimately I would like to add an option to actiontoolbar to look more similar to SegmentedButton while still using standard button design and then deprecate SegmentedButton.

Digitalone1:
@CarlSchwan Thanks for your contribution. I'll ask you here since you posted a proper screenshot. In the plugins list (the sidebar), don't you find the input device row too attached to the top border? Maybe we need a little of top margin?

This is how it looks on Gnome:

<p>
image
</p>

Also, can we do something to adapt the button to the translated strings when the list is empty?

<p>
Schermata del 2025-11-11 18-34-24
</p>

dphaldes:
> Ultimately I would like to add an option to actiontoolbar to look more similar to SegmentedButton while still using standard button design and then deprecate SegmentedButton.

I think I prefer this style over segmented button looks since it is very close to window and panel edges. I think segmented buttons are best reserved for in view controls like forms instead of navigation

CarlSchwan:
> @CarlSchwan Thanks for your contribution. I'll ask you here since you posted a proper screenshot. In the plugins list (the sidebar), don't you find the input device row too attached to the top border? Maybe we need a little of top margin?

Yes, let me fix that in a seperate PR :)

wwmm:
>I don't think there is enough space for the navigation tab bar here

It is what I felt when I tried it. Suddenly the header and the footer became huge. I tried to see if there was a way for it to have icons beside labels but it seems that can't be configured. So I gave up on the navigation tab bar.

wwmm:
Now that I am paying attention to it bringing back the action toolbar made the buttons not centered. Both in the header as well as in the footer. Strange. I used an action bar before but this did not happen.

Digitalone1:
This caused a regression on Gnome.

Before:

<p>
image
</p>

Now:

<p>
Schermata del 2025-11-11 19-31-30
</p>

I have to say I like the previous one. I could also accept the new one, but can we do something to avoid the selected button to collapse on the window bottom border?

Also, this bug has returned: https://github.com/wwmm/easyeffects/issues/4096

wwmm:
For now I will revert this.

wwmm:
And the buttons are centered again. But I am quite sure this was also the case with the action bar before. I wonder why now it does not want to center.

dphaldes:
> This caused a regression on Gnome.

Are you on the default theme ?

Digitalone1:
> Are you on the default theme ?

Yes, Adwaita.

PR #4363, opened 2025-11-11
Use KirigamiSettings.ConfigurationView for the preference dialog

CarlSchwan:
image

### DIFF
--- a/src/contents/ui/PreferencesSheet.qml
+++ b/src/contents/ui/PreferencesSheet.qml
@@ -1,35 +1,52 @@
 import QtQml // Despite of what Qt extension says this import is needed. We crash without it
 import QtQuick
-import QtQuick.Controls as Controls
 import QtQuick.Layouts
 import ee.database as DB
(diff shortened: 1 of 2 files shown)


wwmm:
I have to admit that the fact the search works only for the side panel felt a little disappointing :smile: . I do not know how libadwaita does it but there the preference widget search field somehow filters the preference entries shown in the window. While porting I considered doing what adwaita does but it seemed complicated so I gave up on the idea.

CarlSchwan:
> I have to admit that the fact the search works only for the side panel felt a little disappointing 😄 . I do not know how libadwaita does it but there the preference widget search field somehow filters the preference entries shown in the window. While porting I considered doing what adwaita does but it seemed complicated so I gave up on the idea.

Yeah... I wanted to fix that at some point, but as always too much to do, not enough time. Maybe a quick and dirty solution would be to add some optional keywords to the modules, but ideally the search would work inside the pages too.

PR #4365, opened 2025-11-11
Silence out qmllint errors about i18n

CarlSchwan:
Use the newly introduced .contextProperties.ini. Unfortunately this only works in Qt 6.11 (unreleased), but it doesn't hurt to have it already :)


wwmm:
Thanks! At least there will be a good solution for this. I thought it was really bizarre there was no way to make it ignore some keywords.

PR #4366, opened 2025-11-11
Center shortcuts sheet

CarlSchwan:
Screenshot_20251111_181531


wwmm:
I've noticed a peculiar behavior I am not sure if it was already there or not. If we maximize the window and bring it back to the original size while the shortcuts overlay sheet is visible it shows the vertical scrollbar and never removes it even if there is enough vertical space to show the whole sheet.

wwmm:
>and never removes it even if there is enough vertical space to show the whole sheet.

I mean. If we maximize the window again it is removed while it stays maximized. But before the window was maximized the sheet was totally capable of displaying the whole content in the original window size. Somehow after the maximization it can't.

PR #4368, opened 2025-11-11
Pass GitHub Token to Flatpak lint workflow

vchernin:
This will avoid unnecessary failures and rate limiting. I’m surprised I did not think of this ages ago.


CarlSchwan:
can't approve but +1  ;)

PR #4380, opened 2025-11-11
Update for Ukrainian Localization & Units of Measurement

Drommer:


wwmm:
I had to fix a conflict in weblate because the other Ukrainian translator had things there. At some point I think that you will have to talk to that other person and find a common ground :smile:

PR #4381, opened 2025-11-11
Global bypass button: show label and inline message

Digitalone1:
This is an attempt to solve #4353 remaining on the On/Off logic without reintroducing the global bypass concept.

The previous tooltip is retained and an inline message is shown. The message for the disabled effects uses "informative" style, not the "negative" one because it's not an error.

@wwmm If you don't like it, just discard this PR.

### DIFF
--- a/src/contents/ui/main.qml
+++ b/src/contents/ui/main.qml
@@ -447,15 +447,19 @@ Kirigami.ApplicationWindow {
                         }
                     },
                     Kirigami.Action {
-                        text: i18n("Turn effects on/off") // qmllint disable
+                        text: i18n("Effects") // qmllint disable
+                        tooltip: i18n("Turn effects on/off") // qmllint disable
                         icon.name: "system-shutdown-symbolic"
                         icon.color: checked === true ? Kirigami.Theme.positiveTextColor : Kirigami.Theme.negativeTextColor
-                        displayHint: Kirigami.DisplayHint.IconOnly
+                        displayHint: Kirigami.DisplayHint.KeepVisible
                         checkable: true
                         checked: !DB.Manager.main.bypass
                         onTriggered: {
-                            if (checked !== !DB.Manager.main.bypass)
+                            if (checked !== !DB.Manager.main.bypass) {
                                 DB.Manager.main.bypass = !checked;
+                            }
+
+                            checked ? appWindow.showStatus(i18n("Audio effects are enabled."), Kirigami.MessageType.Positive) : appWindow.showStatus(i18n("Audio effects are disabled.")); // qmllint disable
                         }
                     },
                     Kirigami.Action {


wwmm:
Let's use for a while and see how it feels.

Drommer:
Maybe it makes sense to implement the display of the caption as optional? Not only for this button but in general. Just like it is realized in KDE applications.

wwmm:
> Maybe it makes sense to implement the display of the caption as optional? Not only for this button but in general. Just like it is realized in KDE applications.

If we keep the button text it is something to consider. But I think that more important than that would be to have an icon that represented well the on state and another that represented well the off state. But as far as I remember Breeze does not have it.

Digitalone1:
> Maybe it makes sense to implement the display of the caption as optional? Not only for this button but in general. Just like it is realized in KDE applications.

Not in general. Some plugin controls are very specific and you cannot understand them only with an icon (most of the time I don't even understand some of them with the label since I have to read the documentation, but that is another matter...).

Drommer:
@Digitalone1 I'm not very familiar with QML, but is it possible to set the initial state by default with the option to change it on the user's side? It seems that at least in Dolphin it works this way.

wwmm:
> @Digitalone1 I'm not very familiar with QML, but is it possible to set the initial state by default with the option to change it on the user's side? It seems that at least in Dolphin it works this way.

I did not understand. State of what exactly?

Digitalone1:
> @Digitalone1 I'm not very familiar with QML, but is it possible to set the initial state by default with the option to change it on the user's side? It seems that at least in Dolphin it works this way.

I don't think it's a good idea. Fox example, see the adaptive intensity checkbutton. How do you explain its meaning without the label? There's no icon to do that.

> I did not understand. State of what exactly?

Showing the label near the icon of a button.

Drommer:
@wwmm I meant to set the state (show or hide the button label) by default, but at the same time implement the ability for the user to disable the label (or enable it when it is disabled by default).
@Digitalone1 Some plugins (such as the multiband compressor) have many options in the form of buttons. Some of them do not have enough space to display due to their labels and are hidden in a submenu. But if the user already knows what these buttons do and wants to turn off the labels so that all icons are visible without the submenu.

wwmm:
I think that either we show the label or we don't. Specially because we wouldn't add configuration option to the other buttons. There are too many buttons to handle.

Digitalone1:
qBittorrent does this, but only for the main menu. The other buttons have the fixed label. But considering that we have space in the top bar and the Kirigami logic will autohide them on restricted sizes, I don't see the point to make this optional, even only for the top bar.

> @Digitalone1 Some plugins (such as the multiband compressor) have many options in the form of buttons. Some of them do not have enough space to display due to their labels and are hidden in a submenu. But if the user already knows what these buttons do and wants to turn off the labels so that all icons are visible without the submenu.

If they are too many, I think they could be rearranged, but not hide the label.

Drommer:
The ability to change the order would be a compromise solution.

I also thought that the button for hiding the effects panel (the one next to the add button) would be more suitable labeled "Hide" instead of "Close".

Digitalone1:
> I also thought that the button for hiding the effects panel (the one next to the add button) would be more suitable labeled "Hide" instead of "Close".

Yes. Or maybe `Reduce`, since it's not completely hided.

Issue #1039, opened 2021-07-22
Bypass state in each effect isn't saved between sessions and in presets


AlexFolland:
**To reproduce:**

1. Add an effect.
2. Make it do something cool that you want to be able to enable whenever you want, but usually leave disabled with "Bypass".
3. Save it in an input or output preset, depending on what it is.
4. Restart the computer.

**Observed behavior:**

- The "Bypass" state in each effect is not saved between sessions and in presets.
    - It resets to disabled, meaning the effect is enabled.
- This is a new regression as of EasyEffects.
    - In PulseEffects, a disabled effect stayed disabled between sessions and could be saved as disabled in presets.

**Expected behavior:**

A disabled effect ("Bypass" enabled) is expected to stay that way between sessions and be saved that way in presets.

**Additional commentary:**

- As one example, I like using the pitch effect for certain music to make it sound hilariously demonic, and being able to toggle it on whenever I want.  With this issue, every time I restart the computer, the pitch effect is enabled when I want it to be disabled by default.
- I personally prefer the previous system where all effects were readily visible to be toggled with a quite-obvious switch in the main list instead of hidden in an effect selection context menu and then only togglable with a "Bypass" button that's only visible when visiting each effect's settings.
- Otherwise, EasyEffects is awesome and I love the new name, especially as the person who suggested it in my brainstorming list!  I'm so excited that you adopted it and did all the hard work it took to rebrand.  I have insane amounts of respect.

wwmm:
>This is a new regression as of EasyEffects. 

Not a regression in the usual sense. It was intentional. The reason is that in this context a bypass is typically used just to temporarily listen to the difference the effect is making and not as a way to disable the plugin permanently. The idea is doing that by removing the plugin from the list.

>I personally prefer the previous system where all effects were readily visible to be toggled

It was that way because dealing with GStreamer pipeline is hard. Totally removing the plugin from the pipeline was a pain. So I had a super complex structure where each plugin had a container that was always in the pipeline and the "bypass" was adding or removing a child bin from that one. That is why that list had all plugins. The corresponding "top level bins" were never removed from the pipeline.

The other side effect of the previous design is that all our gtk widgets are loaded all the time. For people that only wanted to use 1 or 2 plugins it was a waste of system memory. It is not that an additional 100 to 150 MB is big by today standards. But it is not nice to waste memory. 

[...]

PR #4387, opened 2025-11-12
Update for Ukrainian Localization & Units of Measurement

Drommer:
1. Some of the changes I made last time were lost due to careless merging. So I restored it.
2. I also have a question: is it possible to replace "-inf" in SpinBox with the relevant symbol (or make this inscription available for translation and do it there)? In Ukrainian language, for example, there is no text abbreviation for this symbol but it looks a little strange in English among localized inscriptions.

Also, three questions about the interface.
1. Is it possible to make the SpinBox align to the bottom edge when the text is transferred to a new line (see screenshot 1)?
2. I think that for this submenu it is necessary to set the automatic size (see screenshot 2).
3. Regarding the "Effects" button. Perhaps it would be worth changing the status of the button (pressed or unpressed) so that it appears pressed when the effects are turned off? The enabled mode is implied by default so there is no need to draw attention to it. 
Знімок_20251112_053000
Знімок_20251112_053126

### DIFF
--- a/po/fr/easyeffects.po
+++ b/po/fr/easyeffects.po
@@ -8,7 +8,7 @@ msgstr ""
 "Project-Id-Version: \n"
 "Report-Msgid-Bugs-To: https://github.com/wwmm/easyeffects/issues\n"
 "POT-Creation-Date: 2025-11-11 22:38+0000\n"
-"PO-Revision-Date: 2025-11-10 14:51+0000\n"
+"PO-Revision-Date: 2025-11-12 11:51+0000\n"
 "Last-Translator: Bundy01 <bundy01@users.noreply.hosted.weblate.org>\n"
 "Language-Team: French <https://hosted.weblate.org/projects/easyeffects/main/"
 "fr/>\n"
@@ -296,7 +296,7 @@ msgstr "Quantité"
 #: src/contents/ui/StereoTools.qml:159 src/contents/ui/StereoTools.qml:333
 #: src/contents/ui/StereoTools.qml:351
 msgid "dB"
-msgstr ""
+msgstr "dB"
 
 #: src/contents/ui/BassEnhancer.qml:108 src/contents/ui/BassEnhancer.qml:168
 #: src/contents/ui/Exciter.qml:112 src/contents/ui/Exciter.qml:172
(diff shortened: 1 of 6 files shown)


Digitalone1:
> 2. I also have a question: is it possible to replace "-inf" in SpinBox with the relevant symbol (or make this inscription available for translation and do it there)? In Ukrainian language, for example, there is no text abbreviation for this symbol but it looks a little strange in English among localized inscriptions.

Maybe it's better to make it available for translations.

> 3. Regarding the "Effects" button. Perhaps it would be worth changing the status of the button (pressed or unpressed) so that it appears pressed when the effects are turned off? The enabled mode is implied by default so there is no need to draw attention to it.

I don't see a real problem if this button is pressed by default. Anyway, if we do as you say, we would reintroduce the Global Bypass concept which will bring back all the long debate that we had to deal with the GTK version. We used this button just to avoid problems, but it seems we have to face them either ways.

Take a look at this issue: https://github.com/wwmm/easyeffects/issues/1039

Digitalone1:
> 1. Is it possible to make the SpinBox align to the bottom edge when the text is transferred to a new line (see screenshot 1)?

I don't want to sound rude, but I have to ask you: Can't you reduce the translations? The originals are `F1 split` and `F2 peak`, just translate split and peak and retain F1/F2. Same for the others in the same plugin.

> 2. I think that for this submenu it is necessary to set the automatic size (see screenshot 2).

Yes, and also the wrap mode.

Digitalone1:
I added the wrap mode for the list items, but I think the problem with the checkbutton is that the label is too long. Try to reduce the translation to `Стандартна модель` (Standard model).

Drommer:
> I don't want to sound rude, but I have to ask you: Can't you reduce the translations?

This does not only occur in this specific location. This screenshot is just an example. However, if it is difficult to implement, then let it be as it is.

Digitalone1:
> This does not only occur in this specific location. This screenshot is just an example. However, if it is difficult to implement, then let it be as it is.

It's difficult because in general we want the widgets to be aligned at the center. Changing this convention will bring out other issues.

wwmm:
>Regarding the "Effects" button. Perhaps it would be worth changing the status of the button (pressed or unpressed) so that it appears pressed when the effects are turned off? The enabled mode is implied by default so there is no need to draw attention to it.

I think this would add confusion because in the other locations a pressed button means enable something. Having it pressed with the meaning changed to disabled feels weird.

Digitalone1:
@Drommer Please open a separate issue for the `-inf` problem, so I can remember to investigate it.

wwmm:
>I also have a question: is it possible to replace "-inf" in SpinBox with the relevant symbol (or make this inscription available for translation and do it there)?

IT is better to just allow it in the translation. Typing `-Inf` is definitely more obvious than trying to figure out how to type -∞

Issue #4325, opened 2025-11-10
It now takes two mouse clicks to toggle an effect


starlightpoppy:
As a KDE user I appreciate the new UI and the move to Qt. But it now takes two mouse clicks to toggle an effect instead of one. Hope something can be done about it.

blakkd:
+1 Cinnamon X11 user

blakkd:
I also find the new UI sexier, but find it less convenient about increasing or decreasing input output. I really miss a slider instead of only entering or clicking +/- and maintaining, which can take a long time.


> Also, but I can't tell if it's just on my side, entering a dB number isn't taken into account and the number reset to its previous value if clicking outside of the field or pressing enter to validate the value. So for now I can only use the slow +/- method

Edit: I just figured out the issue actually only occurs when inputting float numbers using a `.` as separator, `,` is working fine. But maybe allowing both could be cool

Edit 2: fixed, see below ;)

Digitalone1:
I was pretty sure this issue was going to be opened after the release.

Anyway, showing all the buttons may increase the sidebar width, taking in consideration that in locale translations the effect names could be longer than the originals in English.

Before we had a mixed style where the buttons were shown only when there was enough space (so some plugins showed them, others were hidden behind the menu button). Don't know which is the best approach, but I'm fine with the actual one even if directly clicking on the plugin bypass button would be more comfortable. 

> Edit: I just figured out the "entering a dB number isn't taken into account" issue actually only occurs when inputting float numbers using a `.` as separator, `,` is working fine. But maybe allowing both could be cool

No, it's confusing since parameters are rendered by the locale settings, so we expect the values are input by the locale convention either. Which is your locale? Do you notice a mismatch?

craftfever:
> I also find the new UI sexier, but find it less convenient about increasing or decreasing input output. I really miss a slider instead of only entering or clicking +/- and maintaining, which can take a long time. Also, but I can't tell if it's just on my side, entering a dB number isn't taken into account and the number reset to its previous value if clicking outside of the field or pressing enter to validate the value. So for now I can only use the slow +/- method
> 
> Edit: I just figured out the "entering a dB number isn't taken into account" issue actually only occurs when inputting float numbers using a `.` as separator, `,` is working fine. But maybe allowing both could be cool

[...]

PR #4404, opened 2025-11-12
Various improvements

### LINKED ISSUE #4400: Ctrl+q doesn't quit the EasyEffects service on GNOME
### EasyEffects Version

8.0.0

Arch (easyeffects)

### Distribution

Arch Linux

### Describe the bug

On GNOME, <kbd>Ctrl</kbd>+<kbd>q</kbd> doesn't quit the EasyEffects service, but only closes the visible window if the service mode is enabled. The default shortcut for terminating the service is <kbd>Ctrl</kbd>+<kbd>t</kbd> at the moment instead, which works:

https://github.com/wwmm/easyeffects/blob/91949b3267bc44c0e1fbe24c73eb702992fa19ec/src/contents/ui/main.qml#L180-L182

### Expected Behavior

As in previous (GTK) versions, I'd like to see <kbd>Ctrl</kbd>+<kbd>q</kbd> terminate the EasyEffects service.

### Additional Information

Because of some limitation with Qt, <kbd>Ctrl</kbd>+<kbd>q</kbd> does not work as intended outside of KDE (already discussed in [#2960](https://github.com/wwmm/easyeffects/issues/2960#issuecomment-3232096006)). I can also confirm this is still the case on GNOME 49 (by manually compiling with <kbd>Ctrl</kbd>+<kbd>q</kbd> as shortcut).

So the current behavior is somewhat intentional, but I still consider it a bug, as this is not only inconsistent across desktop environments, but <kbd>Ctrl</kbd>+<kbd>q</kbd> is also the de-facto standard to quit GTK applications (not just the visible window). Therefore, I also doubt this is some limitation of GNOME; <kbd>Ctrl</kbd>+<kbd>q</kbd> is not a global shortcut by default either.

Digitalone1:
- Improve `populatePluginsStack` readability returning early if plugins list is empty.
- Added a new option to show fixed buttons in plugins list. A new RowLayout is added and the one to be shown is decided by the database setting. I had also to make new functions for `onTriggered`/`onClicked` signals. @wwmm It works to me, but please test it and ensure the callback functions are working good. Related to #4325.
- RNNoise models list is not shown properly when entry names are too long. I had to add the fillWidth Layout to restrict it inside the container. Also added `elide` and `wrapMode` properties to the entries of other lists so we are sure the text is not overflowing if too long.
- Changed back the CTRL+T shortcut to CTRL+Q. Thanks to ChatGPT I find the `GlobalDrawer` QML class to catch the CTRL+Q quit event from Gnome, so now it terminates the service rather than closing only the window. @wwmm Please test it on your system. You might also consider to remove the previous CTRL+Q `Shortcut` class since there's the duplicate in `GlobalDrawer` (maybe the two conflicts and quit the application two times on KDE). Fixes #4400.

### DIFF
--- a/src/contents/ui/ShortcutsSheet.qml
+++ b/src/contents/ui/ShortcutsSheet.qml
@@ -131,7 +131,7 @@ Kirigami.OverlaySheet {
                     }
 
                     Kirigami.Chip {
-                        text: "T"
+                        text: "Q"
                         closable: false
                         checkable: false
                         down: false
--- a/src/contents/ui/main.qml
+++ b/src/contents/ui/main.qml
@@ -177,8 +177,7 @@ Kirigami.ApplicationWindow {
     }
 
     Shortcut {
-        // This one replaces CTRL+Q that might not work outside KDE.
-        sequences: ["Ctrl+T"]
+        sequences: ["Ctrl+Q"]
         onActivated: Qt.quit()
     }
 
@@ -194,6 +193,25 @@ Kirigami.ApplicationWindow {
         onActivated: {
             DB.Manager.main.processAllOutputs = !DB.Manager.main.processAllOutputs;
         }
+    }
+
+    globalDrawer: Kirigami.GlobalDrawer {
+        /**
+         * The usual way to declare the "shut down the service" Ctrl+Q shortcut
+         * with the Shortcut QML class does not work on Gnome since it closes
+         * only the visible window and does not quit the Easy Effects service.
+         * With this Kirigami.GlobalDrawer class instead we can implement a
+         * StandardKey.Quit action that catches the CTRL+Q event on Gnome, so
+         * we can terminate the service with the keyboard shortcut also there.
+         */
+        actions: [
+            Kirigami.Action {
+                shortcut: StandardKey.Quit
+                onTriggered: {
+                    Qt.quit();
+                }
+            }
+        ]
     }
 
     PreferencesSheet {


wwmm:
So far I did not see any side effect in setting `globalDrawer: Kirigami.GlobalDrawer`. But it feels like an overkill to set a global drawer just to catch a quit key that in the end calls the same `Qt.quit()`. There has to be a better way to solve this... Did you try
```
Shortcut {
        sequences: StandardKey.Quit
        onActivated: Qt.quit()
    }
```
? I see in kirigami application window source code that they create a standard quit action that instead of `Qt.quit()` calls the application window `close()` method https://github.com/KDE/kirigami/blob/c50158a6f251db02773078332c6512f3701d69f6/src/controls/AbstractApplicationWindow.qml#L175
It is also worth testing if using it isn't enough.

gwuen:
I can't talk about the technical side of things, but the new solution works for me on GNOME. Thanks!

wwmm:
>Added a new option to show fixed buttons in plugins list. A new RowLayout is added and the one to be shown is decided by the database setting. I had also to make new functions for onTriggered/onClicked signals. @wwmm It works to me, but please test it and ensure the callback functions are working good

I've found a way to do this without the extra rowlayout. If for some reason it does not work as well as the one from this PR we can go back to it later.

gwuen:
The new option is shown in the Background Service section, I guess this was not the intention?

wwmm:
> The new option is shown in the Background Service section, I guess this was not the intention?

We will need to create a new section for settings like this one. None of the ones currently available is a good fit.

wwmm:
Master branch updated with a new preferences window section named `Style`. I have the feeling we probably will need more settings fitting this section...

Digitalone1:
> The new option is shown in the Background Service section, I guess this was not the intention?

No, it was a mistake of mine.

> So far I did not see any side effect in setting `globalDrawer: Kirigami.GlobalDrawer`. But it feels like an overkill to set a global drawer just to catch a quit key that in the end calls the same `Qt.quit()`. There has to be a better way to solve this... 

I'll try to set the `StandardKey.Quit` in Shortcut when I have time. I wonder also if we can set multiple sequences in a single Shortcut instance. I'll try that too. 

> ? I see in kirigami application window source code that they create a standard quit action that instead of `Qt.quit()` calls the application window `close()` method https://github.com/KDE/kirigami/blob/c50158a6f251db02773078332c6512f3701d69f6/src/controls/AbstractApplicationWindow.qml#L175 It is also worth testing if using it isn't enough.

Okay. Which is the difference between `Qt.quit` and `close`?

wwmm:
>Okay. Which is the difference between Qt.quit and close?

I did not investigate that. At first sight I would not expect "close" to finish the whole app. But they are using it in a signal that is triggered by `StandardKey.Quit`. So I imagine that it does more than just close the application window.

Digitalone1:
@wwmm I tried to follow the documentation: https://doc.qt.io/qt-6/qml-qtquick-shortcut.html#sequences-prop

`sequences` can take an array with multiple keys. I tried:

- [StandardKey.Quit, "Ctrl+Q"]
- [StandardKey.Quit] alone 
- [StandardKey.Quit] and ["Ctrl+Q"] separated with `Qt.quit()` on both
- [StandardKey.Quit] and ["Ctrl+Q"] separated with `Qt.quit()` for the first and `appWindow.close()` for the latter.

I also tried the `sequence` alone which takes a string: `sequence: StandardKey.Quit`, but unfortunately neither of these work without the `GlobalDrawer`.

If you want, I can move all the shortcuts in the GlobalDrawer so we have an unified place to list them.

Digitalone1:
I tried to remove the Shortcut class with `CTRL+Q`, deleted the `StandardKey.Quit` and added `CTRL+Q` in the GlobalDrawer:

```
        actions: [
            Kirigami.Action {
                shortcut: "Ctrl+Q"
                onTriggered: {
                    Qt.quit();
                }
            }
        ]
```

And it works either! It seems the Shortcut class has some incompatibility with Gnome.

What do you want to do? I suggest to move all the shortcuts in the GlobalDrawer. It seems more appropriate since the GlobalDrawer is a Kirigami class while Shortcuts if from Qt Quick.

Digitalone1:
Update. Unfortunately, we have to remove the GlobalDrawer because of #4423.

PR #4419, opened 2025-11-13
Various improvements

### LINKED ISSUE #4415: output-gain of bass_enhancer overrides input-gain value when loading saved preset
### EasyEffects Version

8.0.0

Flatpak (Flathub)

### Distribution

Fedora Linux 43 (Workstation Edition)

### Describe the bug

I have a preset with bass enhancer with input gain as 0 dB and output gain as -2 dB. This works, and appears to save as expected:
```json
{
    "output": {
        "bass_enhancer#0": {
            "amount": 2.000000000000016,
            "blend": 0.0,
            "bypass": false,
            "floor": 10.0,
            "floor-active": false,
            "harmonics": 8.499999999999986,
            "input-gain": 0.0,
            "output-gain": -2.0,
            "scope": 100.0
        },
        "blocklist": [],
        "plugins_order": [
            "bass_enhancer#0"
        ]
    }
}
```
However, when I load the preset, input-gain is ignored and is set to the value of output-gain. This appears both in the UI and when listening. If I save this loaded preset, the input-gain value is now the same as output-gain. I have not tested if this issue appears with any other effects.

Edit: it appears that this is inconsistent. I had it load the preset correctly after doing a few logouts on my system. Then I did something and now it doesn't work again.

### Expected Behavior

input-gain should be the value it was set to in the preset save file when loading it. I'm fairly certain this was the case before 8.0.0.

### Debug Log

_No response_

### Additional Information

_No response_

### LINKED ISSUE #4417: Fields accepting dB values turn `-` (`U+002D : HYPHEN-MINUS`) → `−` (`U+2212 : MINUS SIGN`) → <nothing> with certain system localization formats
### EasyEffects Version

8.0.0

Flatpak (Flathub)

### Distribution

Fedora Linux 43 (Workstation Edition)

### Describe the bug

If I set my system format to `Suomi (Suomi)` (`Finland (Finnish)`) (system language can stay English), the following happens:

 I try to set either input-gain or output-gain in bass_enhancer to `-2,00 dB`, but if I press enter or unfocus the input box, the minus character `-` (`U+002D : HYPHEN-MINUS`) changes to  `−` (`U+2212 : MINUS SIGN`), and if I press enter or unfocus again, the minus character is removed and the value becomes positive.

If I set my system format to `United Kingdom (English)`, the format these fields want to be is:
`-2.00 dB` (minus character is `U+002D : HYPHEN-MINUS {hyphen, dash; minus sign}`)

When the system format is `Suomi (Suomi)`, the format these fields want to be is:
`−2,00 dB` (minus character is `U+2212 : MINUS SIGN`)

The `.` vs `,` for decimals creates no problems.

### Expected Behavior

The minus character in `−2,00 dB` (`U+2212 : MINUS SIGN`) is not removed during field validation (or whatever it's called).

### Debug Log

_No response_

### Additional Information

_No response_

Digitalone1:
I fixed the `-inf` locale, but then I realized this won't work as intended because we have a regex validator that prevents us to input whichever string that is not a number or `-inf` or `-∞`. So I think what @Drommer wants is not possible. 

We should complicate the regext validator for locale strings which I don't think it's worth. Maybe the best we can to is to show `-∞` directly rather than `-inf`. In case a solution is not found, I will remove the localization in a next PR. 

Other changes:
- Improved the number validator to handle minus/plus signs used in other localizations. Fixes #4417.
- Fixed a typo which prevented the input gain to be applied in the Bass Enhancer after loading a preset. Fixes #4415.
- In Bass Enhancer and Crystalizer preset managers, use the macros rather than the plain settings methods.
- Use more Javascript template literals for string concatenations in various QML files.

### DIFF
--- a/src/contents/ui/DelegateStreamsList.qml
+++ b/src/contents/ui/DelegateStreamsList.qml
@@ -74,7 +74,7 @@ Kirigami.AbstractCard {
                             const rate = root.rate.toLocaleString(Qt.locale(), 'f', 1);
                             const latency = root.latency.toLocaleString(Qt.locale(), 'f', 1);
 
-                            return `${root.state} · ` + `${root.format} · ${rate} kHz · ${root.nVolumeChannels} ` + i18n("channels") + ` · ${latency} ms`; // qmllint disable
+                            return `${root.state} · ` + `${root.format} · ${rate} kHz · ${root.nVolumeChannels} ${i18n("channels")} · ${latency} ms`; // qmllint disable
                         }
                         color: Kirigami.Theme.disabledTextColor
                     }
@@ -158,7 +158,7 @@ Kirigami.AbstractCard {
 
                 Controls.Label {
                     Layout.alignment: Qt.AlignHCenter
-                    text: Math.round(volumeSlider.value) + "%"
+                    text: `${Math.round(volumeSlider.value)}%`
                 }
             }
         }
--- a/src/contents/ui/EeProgressBar.qml
+++ b/src/contents/ui/EeProgressBar.qml
@@ -123,7 +123,7 @@ Control {
                 Layout.rightMargin: Kirigami.Units.smallSpacing
                 horizontalAlignment: Qt.AlignRight
                 verticalAlignment: Qt.AlignVCenter
-                text: Number(control.displayValue).toLocaleString(Qt.locale(), 'f', control.decimals) + " " + control.unit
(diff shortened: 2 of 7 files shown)


Digitalone1:
@wwmm qmllint says `import ee.autostart` is unused in `main.qml`. Take a look at it.

wwmm:
> @wwmm qmllint says `import ee.autostart` is unused in `main.qml`. Take a look at it.

It is sued. But as this comes from c++ the qmllint can't properly parse it.

PR #4422, opened 2025-11-13
Use RowLayout of ToolButton instead of SegmentedButton

CarlSchwan:
image

Second try of https://github.com/wwmm/easyeffects/pull/4362

@Digitalone1 please test on gnome ;)

### DIFF
--- a/src/contents/ui/PageStreamsEffects.qml
+++ b/src/contents/ui/PageStreamsEffects.qml
@@ -612,7 +612,7 @@ Kirigami.Page {
                 }
             }
 
-            Components.SegmentedButton {
+            RowLayout {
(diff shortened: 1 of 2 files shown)


Digitalone1:
> @Digitalone1 please test on gnome ;)

I prefer the previous one, but it's okay anyway. 

<p>image</p>

Can be merged in my opinion.

wwmm:
This will need more code 
image
. As it is right now it is possible to have all buttons unchecked.

CarlSchwan:
> This will need more code image . As it is right now it is possible to have all buttons unchecked.

I need to add a buttongroup, doing that asap

PR #4423, opened 2025-11-13
Remove global drawer

CarlSchwan:
It's not needed for the quit action and create a useless floating button.

### DIFF
--- a/src/contents/ui/main.qml
+++ b/src/contents/ui/main.qml
@@ -195,23 +195,19 @@ Kirigami.ApplicationWindow {
         }
     }
 
-    globalDrawer: Kirigami.GlobalDrawer {
+    Kirigami.Action {
         /**
          * The usual way to declare the "shut down the service" Ctrl+Q shortcut
          * with the Shortcut QML class does not work on Gnome since it closes
          * only the visible window and does not quit the Easy Effects service.
-         * With this Kirigami.GlobalDrawer class instead we can implement a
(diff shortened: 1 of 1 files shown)


wwmm:
Only now I've  noticed the float button. It is weird indeed. That workaround was put there because for some reason the shortcut was not killing the app on gnome. It just closed it's window.

Digitalone1:
I noticed it only now. I agree to remove it. And with this I think we should reinstate the CTRL+T shortcut (along with the CTRL+Q).

Digitalone1:
@CarlSchwan I didn't know we could leave a `Kirigami.Action` without the GlobalDrawer. This makes the CTRL+Q work on Gnome. Thanks. 

So no need to reinstate the CTRL+T. Tested on Gnome.

gwuen:
> It's not needed for the quit action

That's not quite correct, it was intended as workaround to get <kbd>Ctrl</kbd>+<kbd>q</kbd> to work for the quit action on non-KDE platforms. But I agree it should be removed if it comes with this side effect.

Digitalone1:
> That's not quite correct, it was intended as workaround to get Ctrl+q to work for the quit action on non-KDE platforms. But I agree it should be removed if it comes with this side effect.

It's working, I tested it on Gnome. He removed the GlobalDrawer, but the Kirigami.Action to catch the CTRL+Q event on Gnome is still there.

Issue #4252, opened 2025-10-30
Inconsistent menu/buttons UI on effects


Jorenar:
### EasyEffects Version

8.0.0

### What package are you using?

AUR (easyeffects-git)

### Distribution

Arch Linux

### Describe the bug

When the name of an effect is on the shorter side, instead of menu, on hover we get buttons on the tile:

<img width="200" height="40" alt="Image" src="https://github.com/user-attachments/assets/6bbb79f0-ca06-4c7b-998f-116c9751fadb" />

<img width="200" height="40" alt="Image" src="https://github.com/user-attachments/assets/12ba4d08-e60e-4ebe-92fc-30859721b843" />

This may lead to user accidentally choosing wrong action.

### Expected Behavior

The UI should be consistent.

### Additional Information

Semi-related issue: #3301

wwmm:
Hum... I like that it is not necessary to open the menu if there is enough space for the button to be visible...

Jorenar:
Perhaps it should be configurable?

Something like:
* dynamic
* always buttons
* always menu

wwmm:
>Perhaps it should be configurable?

Maybe. But there is always the risk of overwhelming the user with options... For now I will go back to the previous behavior. Always show the buttons if there is space for them. And if not show the menu. Visually it looks a little weird when one plugin name is big enough for force its buttons to be in the menu while the others are not. But it is probably better than the user clicking on the wrong button.

wwmm:
> Visually it looks a little weird when one plugin name is big enough for force its buttons to be in the menu while the others are not.

That feels too inconsistent. Let's just show the menu always and see if it gets too annoying or not. Master branch updated with this.

Jorenar:
Okay! Thanks!

Issue #3301, opened 2024-08-09
[Request] Make effect removal more intentional


ParticleMon:
With the Remove and Power icons close together, one can unintentionally remove instead of reorder or toggle ("toggle" being more concise than "Enable/disable"--a separate suggestion) an effect.

Some suggestions for adjusting how an effect is removed:
* Confirm removal similarly to saving a preset
* Move Remove to a context menu
* Drag the reorder handle to the right, change the cursor as when reordering, and change the tooltip to "Remove <Effect>"

Easy Effects 7.1.6

wwmm:
>and change the tooltip to "Remove "

Although it isn't a big deal to do this I do not see the reason why this would bed better...

>Confirm removal similarly to saving a preset

I think this makes the usability too annoying... And it is not like the presets where clicking in the wrong button would destroy data forever.

wwmm:
We can close this issue.

PR #4425, opened 2025-11-13
Rework list section delegate

CarlSchwan:
Use a customized version of Kirigami.ListSectionHeader with extra support for a compact layout.

And simplify a bit the sizing logic to make the item's width depends on the ListView width and not the other way around.

image

image

### DIFF
--- a/src/contents/ui/DelegatePluginsList.qml
+++ b/src/contents/ui/DelegatePluginsList.qml
@@ -86,6 +86,7 @@ Item {
                 Layout.fillWidth: true
                 text: DB.Manager.main.collapsePluginsList === false ? delegateItem.translatedName : delegateItem.translatedName.charAt(0)
                 color: bypass ? Kirigami.Theme.disabledTextColor : Kirigami.Theme.textColor
+                elide: Text.ElideRight
             }
 
             Kirigami.ActionToolBar {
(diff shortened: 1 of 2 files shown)


Digitalone1:
We had the plugins list similar to the one in the screenshot, but before the release we had to change it because a user didn't like the fact that some buttons were visible, others not.

Digitalone1:
See https://github.com/wwmm/easyeffects/issues/4252 and https://github.com/wwmm/easyeffects/issues/3301

CarlSchwan:
> See #4252 and #3301

Pushed a commit which should fix this by making the text elide :)

image

There is some APIs in Kirigami when using a ColumnView to make the sidebar resizable by the user. I will need to see if I can add that in a later pull request.

Digitalone1:
> Pushed a commit which should fix this by making the text elide :)

I have to say that choosing between seeing the buttons and seeing the whole effect name, I would choose the latter. 

> There is some APIs in Kirigami when using a ColumnView to make the sidebar resizable by the user. I will need to see if I can add that in a later pull request.

Yes, making it resizable will be the best. But I hope the app could also remember the sidebar width across different sessions, not having to set the desired size every time.

Thanks for your contribution anyway.

PR #4439, opened 2025-11-14
Use qml modules

CarlSchwan:
This now compiles the qml to pre-cached cpp files

Everything seems to work fine but I might have missed something

### DIFF
--- a/src/contents/ui/Main.qml
+++ b/src/contents/ui/Main.qml
@@ -8,7 +8,6 @@ import ee.pipeline as Pipeline
 import ee.pipewire as PW
 import ee.presets as Presets
 import org.kde.kirigami as Kirigami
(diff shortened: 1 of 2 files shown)


CarlSchwan:
Still need to port everything away from qmlRegisterSingletonInstance to QML_SINGLETON but that's a follow up ;)

wwmm:
Is it ok to have all qml files assoficated to
```
qt_add_qml_module(easyeffects
    URI ee.type.presets
```
? Most of them do not have anything to do with presets. Maybe it is better a new URI just for this. And maybe the other qt_add_qml_module calls have become unnecessary. I added them only to satisfy qmllinter.

CarlSchwan:
We could likely just use one module for everything. This would be a bit simpler indeed :)

PR #4467, opened 2025-11-16
Fix swapped Access and API data fields in client list delegate

techmanwalker:
The “Access” and “API” labels were displaying each other’s values due to a field mix-up.
This commit assigns each label to the correct root property.


wwmm:
Thanks!

Issue #4450, opened 2025-11-15
The `--hide-window` option sometimes doesn't work.


BYT0723:
### EasyEffects Version

8.0.0

### What package are you using?

Arch (easyeffects)

### Distribution

Arch linux

### Describe the bug

After using `easyeffect --service-mode --hide-window` many times to start, a window may appear, especially when I start dwm using this command in autostart.sh, the window will appear every time

https://github.com/user-attachments/assets/201ba9f9-3c71-4054-aadd-be2b6c9fb541

wwmm:
I tried several times but somehow this does not happen on my computer. Intriguing. My guess is that the timing between some events is slightly different on you machine. Maybe instead of using this command line option to hide our window after an autostart we should create a configuration option that handles the window visibility specifically when easyeffects autostarts.

wwmm:
>Maybe instead of using this command line option to hide our window after an autostart we should create a configuration option that handles the window visibility specifically when easyeffects autostarts.

Now that I am thinking about it again it may not be viable to fix this problem this way... We need to know somehow we are autostarting so we do not hide the window in cases where it should be shown...

wwmm:
@BYT0723 I've updated our master branch with some changes that should help. But as I could not reproduce the issue it is hard to be sure.

wwmm:
You can test through the AUR package.

BYT0723:
<img width="1011" height="1120" alt="Image" src="https://github.com/user-attachments/assets/4d9491c5-59a5-4cff-8672-5862f76cc5fe" /> 
I think it may be because DWM has forced XMapWindow.

BYT0723:
> You can test through the AUR package.

OK, I'll try it.

BYT0723:
https://github.com/user-attachments/assets/33eee5fc-3982-4452-9032-c73198b5b4b8

Sorry, Professor, the issue of --hide-window not working still occurs occasionally.

<img width="1531" height="450" alt="Image" src="https://github.com/user-attachments/assets/aca04f96-9fa5-4ee0-9db1-0616121c3ad9" />

wwmm:
Hum... Maybe Qt is not serializing the events I thought it would...

wwmm:
Master branch updated again. Let me know if anything has changed.

BYT0723:
> Master branch updated again. Let me know if anything has changed.

Okay, I was tested `--hide-window` in pure DWM, The problem can still be reproduced, but the probability has decreased.

[...]

PR #4491, opened 2025-11-17
fix `--hide-window` occasionally don't work in dwm 

BYT0723:
fix issue #4450

I tried many times. Initially, I thought it was due to DWM's forced map request, but that wasn't the case. Qt shouldn't send a map request to the Xorg server when creating an object. Explicitly declaring `visible: false` in the QML can prevent sending map requests. This worked for me, and I've tested it on X11-based DWM/XFCE4/KDE, and it works correctly.


wwmm:
As far as I can see it seems ok to do that. There was no side effects at least here on my computer running KDE and wayland.

PR #4496, opened 2025-11-18
Fix manual for F1 shortcut

parona-source:
Help manager was removed.

Fixes: f71b953d848c43c45d00280103c98fb64fb991e5
See-Also: 6507b8bcc20cb5c59640d3fd7b69be0f240694fb

Btw would you receptive to the idea of making qtwebengine optional? Unfortunately it would require restoring the helpmanager due to how the docs are embedded. Or maybe open "https://wwmm.github.io/easyeffects/" instead in the browser?

Qtwebengine is a big package with high build requirements which make it a hard pill to swallow if its only used for the manual. On gentoo it would require dropping support several architectures in easyeffects as its hard to justify support in qtwebengine for them.

### DIFF
--- a/src/contents/ui/Main.qml
+++ b/src/contents/ui/Main.qml
@@ -161,7 +161,7 @@ Kirigami.ApplicationWindow {
     Shortcut {
         sequences: ["F1"]
         onActivated: {
-            Help.Manager.openManual();
+            helpSheet.open();
         }
     }
 


Digitalone1:
> Btw would you receptive to the idea of making qtwebengine optional?

On Flatpak I think we will keep it. For distributions instead I think it's possible to make it optional, but I don't know if there's a way to check runtime if the dependency is installed and fallback to send the URL to the browser in case the webengine is missing.

parona-source:
> > Btw would you receptive to the idea of making qtwebengine optional?
> 
> On Flatpak I think we will keep it. For distributions instead I think it's possible to make it optional, but I don't know if there's a way to check runtime if the dependency is installed and fallback to send the URL to the browser in case the webengine is missing.

I was playing around the idea locally. 

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index cb4393fef..7cbb21aaa 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,7 +51,12 @@ find_package(Qt${QT_MAJOR_VERSION} REQUIRED NO_MODULE COMPONENTS
     Quick
     QuickControls2
     Widgets
-    WebEngineQuick
+)
+
+find_package(Qt${QT_MAJOR_VERSION}WebEngineQuick NO_MODULE)
+set_package_properties(Qt${QT_MAJOR_VERSION}WebEngineQuick PROPERTIES
+    TYPE OPTIONAL
+    PURPOSE "Show Manual pages in the application"
 )
 
 find_package(KF${QT_MAJOR_VERSION} REQUIRED COMPONENTS
diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
index 885d9a30f..7bde4b290 100644
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -53,7 +53,6 @@ qt_add_qml_module(easyeffects
         contents/ui/Expander.qml
         contents/ui/Filter.qml
         contents/ui/Gate.qml
-        contents/ui/HelpSheet.qml
         contents/ui/LevelMeter.qml
         contents/ui/Limiter.qml
         contents/ui/Loudness.qml
@@ -238,7 +237,6 @@ target_link_libraries(easyeffects PRIVATE
     Qt::Quick
     Qt::QuickControls2
     Qt::Widgets
-    Qt::WebEngineQuick
     TBB::tbb
     GSL::gsl
     GSL::gslcblas
@@ -262,6 +260,15 @@ target_link_libraries(easyeffects PRIVATE
     ${LIBZITACONVOLVER}
 )
 
+if (Qt${QT_MAJOR_VERSION}WebEngineQuick_FOUND)
+    qt_target_qml_sources(easyeffects
+        QML_FILES
+        contents/ui/HelpSheet.qml
+    )
+    target_link_libraries(easyeffects PRIVATE Qt::WebEngineQuick)
+    target_compile_definitions(easyeffects PUBLIC -DHAVE_WEBENGINE)
+endif (Qt${QT_MAJOR_VERSION}WebEngineQuick_FOUND)
+
 target_compile_definitions(easyeffects PRIVATE QT_NO_KEYWORDS=1)
 # target_compile_definitions(easyeffects PRIVATE QT_NO_KEYWORDS=1 QT_QML_DEBUG=1)
 
diff --git a/src/main.cpp b/src/main.cpp
index 54c736db9..ec5ed4921 100644
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -370,6 +370,12 @@ int main(int argc, char* argv[]) {
 
   QQmlApplicationEngine engine;
 
+#ifdef HAVE_WEBENGINE
+  engine.rootContext()->setContextProperty("HAVE_WEBENGINE", QVariant(true));
+#else
+  engine.rootContext()->setContextProperty("HAVE_WEBENGINE", QVariant(false));
+#endif
+
   // Starting the local socket server
 
   local_server->startServer();  // it has to be done after "QApplication app(argc, argv)"
```

With this you could use this in qml.

```
if (HAVE_WEBENGINE) {
} else {
}
```

https://stackoverflow.com/a/53336610

Just unsure on the details in qml, either restoring the helpmanager and using that or doing something else. This approach would be automagic in that it would build with qtwebengine support unless it wasn't installed at build time or you use `-DCMAKE_DISABLE_FIND_PACKAGE_Qt6WebEngineQuick=ON`

wwmm:
>Btw would you receptive to the idea of making qtwebengine optional? 

Ideally I would like to not have to use it at all. But doing that outside of flatpak while still using it inside flatpak is not ideal. Increasing the code path difference between the two is not good. We already have to deal autostart in a different way inside flatpak's sandbox.

Maybe we  should just give up on the idea of having an offline manual and just open an online website like gimp does.

Issue #4506, opened 2025-11-19
Some pages are misaligned


wxq1884482:
### EasyEffects Version

8.0.3

### What package are you using?

Arch (easyeffects)

### Distribution

Arch Linux

### Describe the bug

Adjusting the scaling factor to less than 1 using gnome-tweaks may cause slight misalignment of the page.

<img width="736" height="381" alt="Image" src="https://github.com/user-attachments/assets/bebe60c7-d83c-41d3-833d-f6613817b7d8" />

### Expected Behavior

Adjust the page zoom, my device is set to 0.85 scale.

### Debug Log

[admin1@archlinux ~]$ export G_MESSAGES_DEBUG=easyeffects
[admin1@archlinux ~]$ easyeffects 
easyeffects: presets_directory_manager.cpp:140	Old /home/admin1/.config/easyeffects/input directory detected. Migrating its files to /home/admin1/.local/share/easyeffects/input
easyeffects: presets_directory_manager.cpp:149	Could not copy any file. Aborting migration of /home/admin1/.config/easyeffects/input
easyeffects: presets_directory_manager.cpp:140	Old /home/admin1/.config/easyeffects/output directory detected. Migrating its files to /home/admin1/.local/share/easyeffects/output
easyeffects: presets_directory_manager.cpp:149	Could not copy any file. Aborting migration of /home/admin1/.config/easyeffects/output
easyeffects: presets_directory_manager.cpp:140	Old /home/admin1/.config/easyeffects/rnnoise directory detected. Migrating its files to /home/admin1/.local/share/easyeffects/rnnoise
easyeffects: presets_directory_manager.cpp:149	Could not copy any file. Aborting migration of /home/admin1/.config/easyeffects/rnnoise
easyeffects: presets_directory_manager.cpp:140	Old /home/admin1/.config/easyeffects/autoload/input directory detected. Migrating its files to /home/admin1/.local/share/easyeffects/autoload/input
easyeffects: presets_directory_manager.cpp:149	Could not copy any file. Aborting migration of /home/admin1/.config/easyeffects/autoload/input
easyeffects: presets_directory_manager.cpp:140	Old /home/admin1/.config/easyeffects/autoload/output directory detected. Migrating its files to /home/admin1/.local/share/easyeffects/autoload/output
easyeffects: presets_directory_manager.cpp:149	Could not copy any file. Aborting migration of /home/admin1/.config/easyeffects/autoload/output
[...]

PR #4512, opened 2025-11-19
Various improvements

Digitalone1:
- Print the presets on the command line in a prettier list, similar to how mpv prints its profiles.
- Do not show the window when the bypass is toggled on the command line (and show an error on invalid bypass state).
- Do not show the window when a preset is set on the command line (and show an error when the preset does not exist). 
- Added vertical alignment to EeSpinBox. Related to #4506.

@wwmm Talking about the command line, `Q_EMIT onHideWindow()` is emitted on `gapplication-service`. Should we emit it also on `service-mode`?

Besides, since the native window option is an experimental setting, should we follow as we did in libadwaita where we hide the button rather than only disabling it?

### DIFF
--- a/src/contents/ui/EeSpinBox.qml
+++ b/src/contents/ui/EeSpinBox.qml
@@ -219,6 +219,7 @@ FormCard.AbstractFormDelegate {
                 id: textInputSpinBox
 
                 z: 2
+                verticalAlignment: TextInput.AlignVCenter
                 font: spinbox.font
                 color: control.enabled ? Kirigami.Theme.textColor : Kirigami.Theme.disabledTextColor
                 selectionColor: Kirigami.Theme.highlightColor


lucasrsnd (approved):

wwmm:
>Talking about the command line, Q_EMIT onHideWindow() is emitted on gapplication-service. Should we emit it also on service-mode?

Probably not. In most of situations the service-mode option will be used together with --hide-window. But it is totally valid to want service mode without hiding the window. The option gapplication-service is here just for compatibility with old autostart files created for the gtk code. And in this case the desktop file won't have --hide-window there.

>Besides, since the native window option is an experimental setting, should we follow as we did in libadwaita where we hide the button rather than only disabling it?

As native window usability is more reliable than it was in the past I think that just disabling the button so people know the functionality exists is better.

PR #4526, opened 2025-11-21
Add flatpak socket x11 permission

vchernin:
Native plugin windows only support x11. However, we still want the main window to be using wayland, so allow the wayland permission as well.


vchernin:
Closing for now, as this needs further review.

PR #4534, opened 2025-11-22
global blocklist setting

### LINKED ISSUE #4532: Adding/Removing app in blocklist not persisted
### EasyEffects Version

8.0.4

Arch (easyeffects)

### Distribution

Arch Linux

### Describe the bug

Maybe it's a migration issue since I was using easyeffects before v8, but currently I have both:

`.config/easyeffects/db/easyeffectsrc`
```config
[StreamOutputs]
blocklist=...
```

and a blocklist per output at 
`.local/share/easyeffects/output/*.json`
```config
{
  "output": {
     ...
    "blocklist": [ ... ]
```

therefore when i add/remove an app in the exclusion list its not persisted to the output ".json", only to `easyeffectsrc`. Thus on relaunch `easyeffectsrc` is restored back to what's in the (obsolete) output .json
I have to manually edit all  output json files to persist changes

### Expected Behavior

Blocklist should be persisted

### Debug Log

```
Paste your log here
```
</details>

### Additional Information

_No response_

danicc097:
Fixes #4532 

If enabled, it will always use `easyeffectsrc`'s blocklist regardless of the preset.

image

### DIFF
--- a/src/contents/kcfg/easyeffects_db.kcfg
+++ b/src/contents/kcfg/easyeffects_db.kcfg
@@ -1,7 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <kcfg xmlns="http://www.kde.org/standards/kcfg/1.0"
-    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-    xsi:schemaLocation="http://www.kde.org/standards/kcfg/1.0 http://www.kde.org/standards/kcfg/1.0/kcfg.xsd">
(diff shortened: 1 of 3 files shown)


wwmm:
Thanks! I think I will just change the labels a little. And the location of the key in our database. I should have created a section dedicated to presets keys a long time ago.

wwmm:
I also renamed the database key. So you will have to enable the setting again in EE preferences window.

PR #4538, opened 2025-11-22
Improve APO/GraphicEQ preset management

### LINKED ISSUE #4530: "Export APO" feature allows to export EQ settings for the left channel only when splitting channels.
### EasyEffects Version

8.0.3 (cachyos-extra-v3/easyeffects 8.0.3-1.1)

Other (specify below)

### Distribution

CachyOS Linux

### Describe the bug

In the new version, I was pleasantly surprised to discover that easyeffects now allows to export equalizer settings by channel (The save window didn't appear before, only a warning message). However, it turns out that the settings are only exported for one channel (left). I expected that selecting the desired channel and clicking the export button would save the settings for that channel, but that's not what happens.

### Expected Behavior

_No response_

### Debug Log

```
Paste your log here
```
</details>

### Additional Information

_No response_

### LINKED ISSUE #4533: Equalizer: Can't see which band is disabled in the main view
@wwmm I don't remember if we had this issue also in the GTK version. Now that I'm testing the new changes to the APO import/export methods, I'd like to know which band is enabled or not, but I realized that to see this I have to go in the band menu. Could we have a quick way to know which band is enabled avoiding to check inside the band menu?

Besides, I forget that the new added bands (when the band spinbox is increased) are disabled by default, so I though they were enabled. Maybe this can be confusing also for other users that may assume the new added bands are effective while they're not.

Trying to solve this issue, I first thought to disable the gain slider, but this seems too disruptive since a user may still want to change the gain of a disabled band without having to go into the band menu.

At last I changed the enabled property of the "band number" label, making it disabled when the filter is Off.   

<p>Image</p>

@wwmm What do you think? Is it okay for you? We could also extend this behavior to the frequency and quality label, so it's more highlighted.

Digitalone1:
- Fixes #4533. Let's see how it looks. I prefer this way, but we can still polish it in the future.  
- Fixes #4530. In Equalizer Split Channels mode, the exported preset is built from the selected channel.

In Split Channels mode, the inline message informs the user in which channel the preset has been imported (left or right).

A bug has been also fixed which occurred in an edge case. For example, if I first import an APO/GraphicEQ preset of 16 bands in the left channel and then I import another APO/GraphicEQ preset of 10 bands in the right channel, the band number was reduced to 10, altering the configuration in the left channel which is an unwanted behavior. In Split Channels we do not want to change the other channel when we import to the selected one. This has been fixed.

Besides, I saw that the maximum EQ bands is retrieved by the default setting, but this is not quite robust since in the future the default may change. So I changed this to use the `settings->getMaxValue()` and a new equalizer tag for the `numBands` has been added.

@wwmm Please, test on your system.

### DIFF
--- a/util/NEWS.yaml
+++ b/util/NEWS.yaml
@@ -5,9 +5,10 @@ Date: UNRELEASED_DATE
 Description: 
 - Features∶
 - The labels of an Equalizer band in the main view are greyed out when the type of the filter is set to "Off", so the user can quickly know which band is enabled or not.
+- Added the ability to enable a global excluded apps list. In this mode the blocklist set by the user is always valid and do not depend on the blocklist value set in the loaded preset.
 
 - Bug fixes∶
-- Fixed a bug that occurred in the import of APO/GraphicEq presets. In certain conditions, when Equalizer Split Channels mode was enabled, the preset imported to a single channel was altering the configuration in the other channel.  
+- Fixed a bug that occurred in the import of APO/GraphicEq presets. In certain conditions, when Equalizer Split Channels mode was enabled, the preset imported to a single channel was altering the configuration in the other channel.
 
 - Other notes∶
 - 


wwmm:
As far as I can see it seems fine. The only thing I noticed may be happening for quite some time. If the equalizer number of bands is larger than the one in the apo preset the import operation does not changes the number of bands to the one the preset uses. They are just disabled.

If the current number of bands the import does increase them accordingly. It is not a big deal because in the end everything works. It is just a matter of removing the unused bands.

Digitalone1:
> As far as I can see it seems fine. The only thing I noticed may be happening for quite some time. If the equalizer number of bands is larger than the one in the apo preset the import operation does not changes the number of bands to the one the preset uses. They are just disabled.
> 
> If the current number of bands the import does increase them accordingly. It is not a big deal because in the end everything works. It is just a matter of removing the unused bands.

Did you try disabling the split channels? In that case the number should change. I tested it, but I will give another try later.

wwmm:
>Did you try disabling the split channels? In that case the number should change. I tested it, but I will give another try later.

No. I did the tests with split channels enabled.

Digitalone1:
> No. I did the tests with split channels enabled.

I tried it. When split channels is disabled, the number correspond to the APO bands. What you say is intended in split channels because if you have 10 bands in the right channel and then import a 6 bands APO preset in the left channel, you don't want the band number to be 6 because the configuration in the right channel will be deleted.

wwmm:
> > No. I did the tests with split channels enabled.
> 
> I tried it. When split channels is disabled, the number correspond to the APO bands. What you say is intended in split channels because if you have 10 bands in the right channel and then import a 6 bands APO preset in the left channel, you don't want the band number to be 6 because the configuration in the right channel will be deleted.

Oh... In that case it is better to keep it this way.

PR #4571, opened 2025-11-28
Optimize: Tighten plugin DB lifecycle and clean up small typos

oxyzenQ:
Tighten plugin DB lifecycle and clean up small typos

- prune stale plugin configuration entries when rebuilding sie/soe maps
  to avoid long-lived KConfigSkeleton allocations
- fix small typo in shell script

Tests:
- cmake -S . -B build
- cmake --build build -j$(nproc)

Optimization applied: 
create_plugin_db
 now tracks active keys and deletes stale plugin entries, preventing long-lived config objects from lingering in the sie/soe maps (@src/db_manager.cpp#167-345). Qt UI bindings remain intact because the maps keep current keys and emit the same change signals.

### DIFF
--- a/po/ga/easyeffects.po
+++ b/po/ga/easyeffects.po
@@ -8,7 +8,7 @@ msgstr ""
 "Project-Id-Version: easyeffects\n"
 "Report-Msgid-Bugs-To: https://github.com/wwmm/easyeffects/issues\n"
 "POT-Creation-Date: 2025-11-28 00:47+0000\n"
-"PO-Revision-Date: 2025-11-27 05:01+0000\n"
+"PO-Revision-Date: 2025-11-28 14:00+0000\n"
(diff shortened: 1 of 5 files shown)


wwmm:
> Qt UI bindings remain intact because the maps keep current keys and emit the same change signals.

Hum... But I wonder if there is no chance a plugin backend or a qml file won't try to use the deleted database instance. Depending on which place gets the `pluginsChanged` event first this may become a problem. Let's test this for some time and see what happens.

wwmm:
> But I wonder if there is no chance a plugin backend or a qml file won't try to use the deleted database instance.

Unfortunately that happens. Probably because the database manager class gets `pluginsChanged` before this signal gets to the other slots that also use it. An easy way to see the crash is alternating between a preset file that has no filters and one that has.

oxyzenQ:
@wwmm thanks

PR #4574, opened 2025-11-28
Add more template units + various improvements

Digitalone1:
- Added more template units and replaced them throughout the QML code.
- Fixed the missing automatic gain control in Echo Canceller preset.
- Added the ability to customize the unit separation in EeSpinBox and EeProgressBar (useful for degree values, see the Stereo Phase in Stereo Tool).
- Fixed the Gate level meters which had a wrong row spacing, unlike the other level meters in other plugins.
- Fixed the wrong alignment of the low/high pass frequency spinboxes.

I was also wondering if we can add an option to start the application with a custom language. There are other Qt apps that can do this, like qBittorrent. This way, if a translation is incomplete, the user can switch to English for a better view.

After the merge, please update the translation templates.

### DIFF
--- a/src/contents/ui/Convolver.qml
+++ b/src/contents/ui/Convolver.qml
@@ -384,7 +384,7 @@ Kirigami.ScrollablePage {
                         id: irDuration
 
                         Layout.alignment: Qt.AlignHCenter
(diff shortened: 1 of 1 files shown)


wwmm:
> I was also wondering if we can add an option to start the application with a custom language. There are other Qt apps that can do this, like qBittorrent. This way, if a translation is incomplete, the user can switch to English for a better view.

Other than setting the `LANG` enviroment variable and manually restarting the application nothing else comes to my mind right now.

PR #4635, opened 2025-12-08
feat: add plugin properties and active preset to local server 

danicc097:
allows dynamic access to all plugin parameters + active preset via the local socket. 

why: 
e.g. I want to have some system shortcuts to modify some plugin parameters, with logic also per current preset.  It's cumbersome to either maintain multiple presets or modify the appropriate json on the fly and reapply, and in my case  also makes my dotfiles state dirty. The `easyeffects -a` flag might also return stale data too when I switch devices and then trigger the shortcuts

I don't know if you're interested in the autogenerated docs, but it's easier to find something there instead of digging in the kcfg files if this were to be merged. I thought of adding a `get_schema` instead, but I feel the docs page is much easier to deal with.

I also added the realtime active preset to the cli now that's possible

### DIFF
--- a/src/command_line_parser.cpp
+++ b/src/command_line_parser.cpp
@@ -104,22 +104,18 @@ void CommandLineParser::process_events() {
 
     const auto value = parser->value("active-preset");
 
-    if (value == "input") {
-      auto preset = DbMain::lastLoadedInputPreset();
+    if (value == "input" || value == "output") {
+      if (is_primary) {  // no daemon running
+        auto preset = (value == "input") ? DbMain::lastLoadedInputPreset() : DbMain::lastLoadedOutputPreset();
 
-      if (preset.length() < 1) {
-        preset = QString("None");
+        if (preset.length() < 1) {
+          preset = QString("None");
+        }
+
+        std::cout << preset.toStdString() << '\n';
+      } else {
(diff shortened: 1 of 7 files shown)


wwmm:
Tonight I will take a look at this. But "active preset" is something that does not exist in easyeffects. Presets are just an easy way to load a group of settings. After that the preset file is not used anymore. So all the places where this term was used should be renamed to something like "last loaded preset" or "last used preset".

danicc097:
> Tonight I will take a look at this. But "active preset" is something that does not exist in easyeffects. Presets are just an easy way to load a group of settings. After that the preset file is not used anymore. So all the places where this term was used should be renamed to something like "last loaded preset" or "last used preset".

agreed. we can use last loaded preset which is what db uses, and maybe also add new cli flags since there still is active-preset and active-presets . maybe we can add last-loaded-preset and last-loaded-presets so there's no more confusion, keep the shorthand -a and -s and show a deprecation notice on active-preset and active-presets

wwmm:
@danicc097 is it possible to make the python script to generate the properties doc in the markdown format used by mkdocs? I am moving the documention to this framework. It will more practical than handling html files directly.

danicc097:
Sure, I'll take a look later. I also saw there were no docs about the local server itself, will add a mini section about it alongside

PR #4637, opened 2025-12-08
Fixed "ENABLE_RNNOISE" option having no effect at all

RubisetCie:
Pretty self-explainatory, I guess... :)

During the switch from Meson/GTK to CMake/Qt, the implementation of the option to enable or disable `rnnoise` wasn't properly ported: it was there, but setting it to `OFF` had no effect and still tried to build the `rnnoise` related effects & presets.

This PR fixes it, please don't hesitate to give suggestions and feedbacks.


wwmm:
The idea behind this option has always been just disable enough code to make the easyeffects compile in systems without rnnoise. Is it really necessary  to disable the other parts for that? If possible I would like to avoid inserting this check in so many different places.

RubisetCie:
> The idea behind this option has always been just disable enough code to make the easyeffects compile in systems without rnnoise. Is it really necessary to disable the other parts for that? If possible I would like to avoid inserting this check in so many different places.

I assure you I tried to put the minimum number of checks I could to be able to compile the project without `rnnoise`...

I understand what you mean, it's split accross several source files, but with the way `rnnoise` is integrated to `easyeffects`, I don't see how I could've do differently... or else, maybe with some refactorings.

Edit: possibly, if you're refering to the changes in `presets_community_manager.cpp`, I can remove a couple of checks, yes. Would you want me to do it?

wwmm:
> Edit: possibly, if you're refering to the changes in `presets_community_manager.cpp`, I can remove a couple of checks, yes. Would you want me to do it?

These ones too. But the others also seem strange. For example the ones related to the database calls. They do not need rnnoise functions to work. So that part of the code should also compile without rnnoise. But to be honest I still do not see why any of the code outside of `rnnoise.cpp` and `rnnoise.hpp` would need to be disabled for the compilation to succeed.

wwmm:
I am doing some tests and I see that there were definitely regressions in the cmake files. RNNoise should not be a required package. So at least that is broken indeed.

wwmm:
I fixed the regressions that made rnnoise required on my side. At least on my computer it now compiles without rnnoise installed. Let me know how things are on your computer and if something else needs to be disabled for easyeffects to compile.

RubisetCie:
> I fixed the regressions that made rnnoise required on my side. At least on my computer it now compiles without rnnoise installed. Let me know how things are on your computer and if something else needs to be disabled for easyeffects to compile.

Yes, it works on my end too. Thank you very much. :)

I'll close my pull request then, however I just have two questions/remarks:

- Wouldn't you prefer to keep the conditional around `pkg_check_modules(LIBRNNOISE rnnoise IMPORTED_TARGET)` instead of just setting it as non-required?
- Regarding the argument that the reason `librnnoise` is optional is that it isn't commonly available on all distros, what about `webrtc-audio-processing-2`? Could it be made optional too?

wwmm:
> * Wouldn't you prefer to keep the conditional around `pkg_check_modules(LIBRNNOISE rnnoise IMPORTED_TARGET)` instead of just setting it as non-required?

It does not seem necessary. CMake is doing what it is supposed to do when the required flag is not there.

> * Regarding the argument that the reason `librnnoise` is optional is that it isn't commonly available on all distros, what about `webrtc-audio-processing-2`? Could it be made optional too?

I think we should not do this to the webrtc-2 library. Its situation is quite different from the one that involved RNNoise. WebRTC 2 is the last stable release. Package maintainers just have to do what they do to any other library and use the last stable release. If they are afraid that PipeWire is going to have some kind of problem with webrtc 2 they should do the same Arch Linux is doing. Provide a legacy webrtc-1 package for pipewire and the last release for the other packages. Forcing other projects to use the older webrtc version just because PipeWire is still using is not a reasonable approach.

Even RNNoise should not need this flag anymore. A few years ago it had a license distributions like Debian considered incompatible with them. But if I remember well RNNoise changed its license to something that should be compatible with these distributions.

RubisetCie:
OK, I understand. Thank you for answering to my questions.

PR #4641, opened 2025-12-09
Remove qtwebengine dependency from more places

vchernin:
Pretty self explanatory.

One thing I may look into tomorrow if I have time, is to try publishing both releases and master manual versions to github pages. So something like `wwmm.github.io/easyeffects/v8.0.7` and `wwmm.github.io/easyeffects/master` and  `wwmm.github.io/easyeffects/latest` can all be valid urls to go to. So even though we don't have an offline manual anymore, users on older versions can still get the exact manual from that release. 

Unless perhaps it is better to always give the latest release version manual? How often does manual content change in a way that would confuse users on older versions?

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -176,8 +176,6 @@ jobs:
         qt6-graphs-devel
         qt6-quickshapes-devel
         qt6-quickcontrols2-devel
-        qt6-webengine
-        qt6-webenginequick-devel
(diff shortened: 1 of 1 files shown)


wwmm:
> Unless perhaps it is better to always give the latest release version manual? How often does manual content change in a way that would confuse users on older versions?

It has been quite a long time since our manual received big updates. There is actually quite a lot of outdated information there. I wish people volunteered to help with this.

PR #4655, opened 2025-12-10
Option to disable Flatpak / LibPortal requirement

RubisetCie:
Since XDG portals are useless if one is not building EasyEffects for Flatpak, this pull request adds a quick option to control the requirement of the later (`ENABLE_FLATPAK`).

It was used only for the autostart feature, but there was already a fallback, so no functionality should be impacted.

Please tell me what you think.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -272,11 +272,11 @@ if(ENABLE_SANITIZER)
     target_link_options(easyeffects PRIVATE "-fsanitize=address")
 endif(ENABLE_SANITIZER)
 
-if(LIBRNNOISE_FOUND AND ENABLE_RNNOISE)
+if(ENABLE_RNNOISE)
     MESSAGE(STATUS "Enabling rnnoise")
     target_compile_definitions(easyeffects PRIVATE ENABLE_RNNOISE=1)
     target_link_libraries(easyeffects PRIVATE PkgConfig::LIBRNNOISE)
-endif(LIBRNNOISE_FOUND AND ENABLE_RNNOISE)
+endif(ENABLE_RNNOISE)
 
 if(LIBPORTAL_FOUND AND LIBPORTALQT_FOUND AND ENABLE_LIBPORTAL)
     MESSAGE(STATUS "Flatpak build")


wwmm:
Initially the idea was not having a fallback and handle the autostart entirely with libportal. But it still has some limitations like not allowing entra desktop file options to be set.

In any case "enable flatpak" isn't a good name. It is better to call it "ENABLE_LIBPORTAL" like @vchernin had done in the past.

RubisetCie:
> Initially the idea was not having a fallback and handle the autostart entirely with libportal. But it still has some limitations like not allowing entra desktop file options to be set.
> 
> In any case "enable flatpak" isn't a good name. It is better to call it "ENABLE_LIBPORTAL" like @vchernin had done in the past.

Alright, I will rename it accordingly. :)

Arguably, I wouldn't be fond of using libportal just to handle the autostart... as it introduces no less than 4 extra dependencies: **libportal** and it's qt6 backend (which isn't yet available on Debian), **GLib** and **GIO**... for a job as trivial as it is, I'd suggest keeping the original way (moreover, it's still XDG compliant).

vchernin:
I think this is fine but I wanted to clarify a couple things.

 > Since XDG portals are useless if one is not building EasyEffects for Flatpak, this pull request adds a quick option to control the requirement of the later (`ENABLE_FLATPAK`).

I should point out the premise here is wrong, XDG portals are used in other places. For example one is already used for the Global shortcuts feature. But libportal does not provide support for the global shortcuts portal so that works via manual dbus calls. Other things like the color scheme setting (dark mode) also work via a portal, but that is handled by Qt/Kirigami.

Long term trying to make libportal optional may be more difficult if more functionality is added. Unfortunately I don’t think libportal is maintained as well as it could be (e.g. no global shortcuts support still). But perhaps one day a lot of packages will use it so trying to remove it entirely from your system may be futile. However that day is not today so I think it is fine to go ahead with this.

> It was used only for the autostart feature, but there was already a fallback, so no functionality should be impacted.
> 
> Please tell me what you think.

As long as there are clear and distinct code paths I think it should be fine. Flatpak must always try to use the portal and just outright fail if it can’t. Non-Flatpak should never use the portal in my opinion even if it is available. Just to keep things consistent.

src/CMakeLists.txt
```diff
@@ -274,11 +272,17 @@ if(ENABLE_SANITIZER)
     target_link_options(easyeffects PRIVATE "-fsanitize=address")
 endif(ENABLE_SANITIZER)
 
-if(ENABLE_RNNOISE)
+if(LIBRNNOISE_FOUND AND ENABLE_RNNOISE)
```
vchernin: This looks like an auto feature anti-pattern to me. We should always try to build with rnnoise if the feature is enabled. 

If the library is not found, but the feature is enabled we should be outright failing the build. Which is what is currently done without this change. Reason being it is easy to accidentally forget a dependency. If packages like Debian need to build without rnnoise they should explicitly disable the feature with the build flag. I find it is much clearer this way.

src/CMakeLists.txt
```diff
@@ -274,11 +272,17 @@ if(ENABLE_SANITIZER)
...
     MESSAGE(STATUS "Enabling rnnoise")
     target_compile_definitions(easyeffects PRIVATE ENABLE_RNNOISE=1)
     target_link_libraries(easyeffects PRIVATE PkgConfig::LIBRNNOISE)
-endif(ENABLE_RNNOISE)
+endif(LIBRNNOISE_FOUND AND ENABLE_RNNOISE)
+
+if(LIBPORTAL_FOUND AND LIBPORTALQT_FOUND AND ENABLE_LIBPORTAL)
```
vchernin: Same thing here, we should not check if libportal is found here at all. Only check if the feature is enabled.

RubisetCie:
> As long as there are clear and distinct code paths I think it should be fine. Flatpak must always try to use the portal and just outright fail if it can’t. Non-Flatpak should never use the portal in my opinion even if it is available. Just to keep things consistent.

I understand, thank you for the clarification.

src/CMakeLists.txt
RubisetCie: I tend to agree with you. However, I did it like that because I understood it was the preferred way of doing dependency checks by the author... maybe I misunderstood, though.

What would you think is better @wwmm, leaving `rnnoise` and `libportal` as non-REQUIRED and check if found later, or else, skip the pkg_config call ahead of time if the option is disabled?

wwmm: I think that using `LIBRNNOISE_FOUND` is not necessary. Under normal circumstances I would like to not even have to make rnnoise optional. But some distributions still seem to think its license is a problem.

The main idea behind `ENABLE_RNNOISE` is just to give these distributions a way to force the build to work without rnnoise. As far as EasyEffects design is concerned making it optional was never the intention.

wwmm:
>Arguably, I wouldn't be fond of using libportal just to handle the autostart... as it introduces no less than 4 extra dependencies: libportal and it's qt6 backend (which isn't yet available on Debian), GLib and GIO... for a job as trivial as it is, I'd suggest keeping the original way (moreover, it's still XDG compliant).

What I do not like about the current situation is that it makes it harder to fix bugs. On top of the fact that I do not use the Flatpak package there is the different code path from one package to another. I'd rather have a few extra libraries as dependencies than this is an extra layer of annoyance when trying to fix bugs. But libportal is not allowing us to do what we need right now.

RubisetCie:
> What I do not like about the current situation is that it makes it harder to fix bugs. On top of the fact that I do not use the Flatpak package there is the different code path from one package to another.

Understandable concern.
However, for what it's worth, I tend to think the convenience of having optional dependencies makes it worth it...

And in this case, it's just five small new code paths that spans accross a single file... so IMO it's okay, pretty manageable. :D

> I'd rather have a few extra libraries as dependencies than this is an extra layer of annoyance when trying to fix bugs

In my experience, the lesser the dependencies, the easier it is to fix bugs (and avoid vulnerabilities). I'm fine having `libportal` optional (I think it's reasonable), but if this is a concern of yours, how about ditching it for the time being, then?

wwmm:
> In my experience, the lesser the dependencies, the easier it is to fix bugs (and avoid vulnerabilities). 

In this particular case quite a lot of users had problems with the app autostart in the flatpak package. Not that there were much we could do about it...

>I'm fine having `libportal` optional (I think it's reasonable), but if this is a concern of yours, how about ditching it for the time being, then?

As things are right now we are not using it outside of flatpak. So going back to the old behavior where the flatpak was the one enabling libportal is ok. Just remove the `LIBRNNOISE_FOUND` and this PR is fine.

wwmm:
> Just remove the `LIBRNNOISE_FOUND` and this PR is fine.

And the other "_FOUND" that came with it.

PR #4678, opened 2025-12-16
Effects list sidebar: increase top/bottom padding

Digitalone1:
I remember that top and bottom bars were shorter, so the padding was increased, but the same has not been done for the header in the sidebar. Indeed `Add effect` and `Close` buttons feels too attached to the top and bottom borders.

This is the original with small spacing:

<p>
Schermata del 2025-12-16 10-00-24
</p>

This is with medium spacing:

<p>
Schermata del 2025-12-16 10-01-07
</p>

This is with large spacing:

<p>
Schermata del 2025-12-16 10-03-47
</p>

@wwmm  I feel large spacing is better. and follows the padding in the top and bottom bars of the window app. But feel free to change it if it seems out of place on your system.

### DIFF
--- a/src/contents/ui/PageStreamsEffects.qml
+++ b/src/contents/ui/PageStreamsEffects.qml
@@ -336,8 +336,8 @@ Kirigami.Page {
                 GridLayout {
                     columns: DbMain.collapsePluginsList === true ? 1 : 2
                     Layout.fillWidth: true
-                    Layout.topMargin: Kirigami.Units.smallSpacing
-                    Layout.bottomMargin: Kirigami.Units.smallSpacing
+                    Layout.topMargin: Kirigami.Units.largeSpacing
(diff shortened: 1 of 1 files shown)


wwmm:
I don't mind keeping it with the large space. I just wonder which of the HIG space guidelines would apply here https://develop.kde.org/hig/layout_and_nav. I suppose this is one of those moments where it makes sense to ask HIG people what they think.

wwmm:
> I suppose this is one of those moments where it makes sense to ask HIG people what they think.

Question asked.

wwmm:
According to one of the main KDE developers involved with KDE HIG development medium space is what would make more sense in this place as the functionality is similar to what a toolbar does. That being the case I think that using medium should make things more consistent with other parts of the app.

PR #4732, opened 2026-01-05
Update copyright notice. 

Digitalone1:
For the new year.

@wwmm I wonder if we need this notice also for QML files.


wwmm:
> I wonder if we need this notice also for QML files.

Kirigami puts copyright information in their qml files. For example https://github.com/KDE/kirigami/blob/master/src/controls/ActionToolBar.qml. But I don't mind keeping things as they are now. If it is easy to do and you are willing to do it feel free to go ahead.

PR #4739, opened 2026-01-07
Fix app name

Digitalone1:
I see a lot of `EasyEffects Manual` labels in the HTML pages. I suppose changing the title in `mkdocs.yaml`, they will be regenerated with the proper label, right?

Please update translation templates.


wwmm:
> I suppose changing the title in `mkdocs.yaml`, they will be regenerated with the proper label, right?

After changes to the yaml or the markdown files it is necessary to run `mkdocs build` in the project root for the html files to be updated. I am not sure if the github action that deploys the page is doing that. I did it now.

PR #4749, opened 2026-01-09
Restore window state on main window reopen

uenatole:
Comment copy:
```
// NOTE: When the window is reopened, its state is irritatingly lost.
//       This happens both after restarting the application and when switching its visibility through the tray.
//       Until this issue is fixed with xx-session-restore-v1/etc, we can use the approach used in the qBittorrent:
//       save the window properties to a file when hiding it and restore it when showing it.
```

### DIFF
--- a/src/contents/ui/Main.qml
+++ b/src/contents/ui/Main.qml
@@ -32,6 +32,7 @@ import org.kde.kirigami as Kirigami
 Kirigami.ApplicationWindow {
     id: appWindow
 
+    visible: false
     width: DbMain.width
     height: DbMain.height
     title: applicationName


Digitalone1:
Hello. Thanks for your contribution.

What do you mean by "window state"? I do not see issues restoring the window. For example, if I set the application tab and close the window, when I reopen Easy Effect I see the application tab again. Maybe you mean restoring the window position?

uenatole:
> Hello. Thanks for your contribution.
> 
> What do you mean by "window state"? I do not see issues restoring the window. For example, if I set the application tab and close the window, when I reopen Easy Effect I see the application tab again. Maybe you mean restoring the window position?

I meant that after window reopen it loose its window state. If window was maximized, it will loose its maximized state after reopen, and it's not good.

Digitalone1:
> I meant that after window reopen it loose its window state. If window was maximized, it will loose its maximized state after reopen, and it's not good.

I think this is an issue of the Wayland compositor you are using because on my system the maximized state is retained. The only thing not saved is the window position, but this is a well know Wayland issue, at least on Gnome. Which desktop environment do you use?

uenatole:
> I think this is an issue of the Wayland compositor you are using because on my system the maximized state is retained. The only thing not saved is the window position, but this is a well know Wayland issue, at least on Gnome. Which desktop environment do you use?

KWin. Again, this works at qBittorrent just because it saves window geometry and window state to preferences and restores it on reopen, even if native platform doesn't support it. With KWin there are no problem with restoring geometry (without position component), but every time window is hidden it loose its maximized state.

uenatole:
see: https://bugs.kde.org/show_bug.cgi?id=437089.

Digitalone1:
The main developer @wwmm uses KDE too and it's weird he didn't notice this issue. And it's even more weird the maximized state is retained on Gnome and not in KDE. 

@uenatole Is the window size restored? If yes, I think it's only a matter of saving the maximized state (like we were doing in the GTK version).

uenatole:
> The main developer @wwmm uses KDE too and it's weird he didn't notice this issue. And it's even more weird the maximized state is retained on Gnome and not in KDE.
> 
> @uenatole Is the window size restored? If yes, I think it's only a matter of saving the maximized state (like we were doing in the GTK version).

Yes, window geometry restores properly. That's why the only thing I save is window state.

Digitalone1:
> Yes, window geometry restores properly. That's why the only thing I save is window state.

Okay, you are saving [this](https://doc.qt.io/qt-6/qt.html#WindowState-enum). But I wonder if there's a better solution to save this value. We bound the window width/height to the corresponding database properties through QML. Maybe we could save this value in the same way avoiding to implement an additional C++ class.

uenatole:
> > Yes, window geometry restores properly. That's why the only thing I save is window state.
> 
> Okay, you are saving [this](https://doc.qt.io/qt-6/qt.html#WindowState-enum). But I wonder if there's a better solution to save this value. We bound the window width/height to the corresponding database properties through QML. Maybe we could save this value in the same way avoiding to implement an additional C++ class.

I'm not familiar with QML, so I can't to provide any alternative solution but yes, there are `Window.visibility` property that can be used for this.

UPD: done; please, check the latest commit.

Digitalone1:
Okay, nice. Can you please switch the comment style to 

```
\**
 * Line 1
 * Line 2
 * ...
 *\
```

Thanks.

uenatole:
> Okay, nice. Can you please switch the comment style to
> 
> ```
> \**
>  * Line 1
>  * Line 2
>  * ...
>  *\
> ```
> 
> Thanks.

done

wwmm:
Messing with the window visibility will bring the problem fixed by https://github.com/wwmm/easyeffects/pull/4491 back.

wwmm:
> Messing with the window visibility will bring the problem fixed by #4491 back.

Nevermind. The other issue was related to `visible` and not `visibility`.

wwmm:
That being said this PR removes `visible: false`. And that will definitely bring https://github.com/wwmm/easyeffects/pull/4491 back.

wwmm:
I put the `visible: false` there again. I think that now we can merge this.

wwmm:
Suddenly I started to see the warning

>qrc:/qt/qml/ee/ui/contents/ui/Main.qml:32:1: QML Main: Conflicting properties 'visible' and 'visibility'

And the window was not shown. After recompiling with `appWindow.visibility = DbMain.visibility;` removed from the code things wen back to normal. Even if `appWindow.visibility = DbMain.visibility;` is added again.

I will disable the code that sets `appWindow.visibility` until we understand what the hell is going on. Suddenly having the window in an state where it is impossible to see it is not good.

Digitalone1:
I wonder why Qt does not have a simple way to set the maximized state as we were doing on the GTK version.

wwmm:
> I wonder why Qt does not have a simple way to set the maximized state as we were doing on the GTK version.

It is actually simpler than it was in gtk. Setting `visibility: Window.Maximized` is enough for that. And for the other possible states https://doc.qt.io/qt-6/qwindow.html#Visibility-enum.

As removing `visible:false` was not enough to make things right I suspect that setting `visibility` inside `onVisibleChanged` may not be the right way to do it.

wwmm:
Out of curiosity I did the test
```
visible: false
visibility: Window.Maximized
```
and that guarantees the warning although the window is still shown. So somehow the fact we need to set `visible: false` is related.

It is not a surprise that `visible: false` conflicts with maximized. These are opposed operations. But that is a sign that handling window state in our case is not going to be so simple. Just removing the visibility management from `onVisibleChanged` is probably not going to be enough.

uenatole:
The confict of `visible` and `visibility` was the reason I removed `visible : false`. If it causes the issues in some environments like `dwm` we should to set initial visibility state from the point it comes into contact with `--hide-window` and other initial options. For reference you may look into kde bug was mentioned earlier. The solution was provided differs with mine (a bit).

PR #4770, opened 2026-01-12
Crystalizer UI: use sentence case following KDE HIG

Digitalone1:
As the title says.


wwmm:
Thanks! I keep forgetting about this :smile:

PR #4787, opened 2026-01-16
Do external plugin initializations on main thread

openglfreak:
We can't really know which library may use fftw, and any plugin that uses fftw that isn't initialized on the main thread may crash the application.

This should fix the inconsistent but frequent crashes I've been getting when starting EasyEffects.

### DIFF
--- a/src/stereo_tools.cpp
+++ b/src/stereo_tools.cpp
@@ -18,6 +18,7 @@
  */
 
 #include "stereo_tools.hpp"
+#include <QApplication>
 #include <algorithm>
 #include <cstddef>
 #include <format>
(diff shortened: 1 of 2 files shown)


wwmm:
I think this touches quite a lot of things that do not need to be touched. And in a way ti throws away some of the advantages of the workser thread. I have seen in many situations the main window animation freezing because heavy plugin initializations were being done in the main thread. This PR is taking this to the extreme.

wwmm:
>This should fix the inconsistent but frequent crashes I've been getting when starting EasyEffects.

Hum... The crashes related to the fftw stopped on my side after using a global mutex shared by the code that directly or indirectly uses fftw. If there are other plugins using it it is better to make more use of the global mutex where necessary than moving what is in the worker thread to the main thread.

wwmm:
The changes related to `std::ranges::copy` are fine. Wrapping all the `lv2_wrapper->create_instance` calls with `QMetaObject::invokeMethod` feels like an overkill. But there should not be a downside in doing that.

But I prefer to keep what was in the worker thread in the worker thread. In the past the calls related to fftw were being done in the main thread. Multiple fftw instances can still cause crashes even if they are initialized in the same thread. The only thing that stopped that was the global mutex used for them. So all that moving those calls to the main thread will achieve is bringing back the freezing that could happen to the main window animations in some situations.

wwmm:
For now let's merge everything except bcd3f935c89e092d6951712b7d68c6805ce7f2e2. The other changes do not have any downside as far as I can see. If moving things out of the worker thread is still necessary on your side we should investigate why. If possible I would like to avoid putting too much work in the main thread.

PR #4802, opened 2026-01-20
Prefer std::clamp in place of min and max combinations

Digitalone1:
While testing this I got a crash:

```
easyeffects: resampler.cpp:34	error while initializing speex resampler: Invalid argument.
easyeffects: resampler.cpp:34	error while initializing speex resampler: Invalid argument.
easyeffects: resampler.cpp:34	error while initializing speex resampler: Invalid argument.
easyeffects: resampler.cpp:34	error while initializing speex resampler: Invalid argument.
Qt has caught an exception thrown from an event handler. Throwing
exceptions from an event handler is not supported in Qt.
You must not let any exception whatsoever propagate through Qt code.
terminate called after throwing an instance of 'std::length_error'
  what():  vector::_M_default_append
Annullato                  (core dump creato)
```

I cant reproduce it anymore. But I didn't touch anything in the resampler. How is it possible?

### DIFF
--- a/src/convolver_kernel_fft.cpp
+++ b/src/convolver_kernel_fft.cpp
@@ -69,8 +69,8 @@ auto ConvolverKernelFFT::calculate_fft(std::vector<float> kernel_L,
   auto linear_spectrum_R = util::interpolate(freq_axis, spectrum_R, linear_freq_axis);
 
   // Initialize logarithmic frequency axis
-  auto max_freq = std::ranges::max(freq_axis);
-  auto min_freq = std::ranges::min(freq_axis);
+  const auto max_freq = std::ranges::max(freq_axis);
+  const auto min_freq = std::ranges::min(freq_axis);
 
   util::debug(std::format("Min fft frequency: {}", min_freq));
   util::debug(std::format("Max fft frequency: {}", max_freq));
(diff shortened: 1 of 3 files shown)


wwmm:
>I cant reproduce it anymore. But I didn't touch anything in the resampler. How is it possible?

You may still have the coredump related to the crash. Take a look at the output of `coredumpctl list` and after finding the correct PID do `sudo coredumpctl info PID`.

wwmm:
>How is it possible?

Most likely a bad sampling rate as input value. We do not use the resampler in many places. Which plugins were you using at the time?

wwmm:
>resampler.cpp:34	error while initializing speex resampler: Invalid argument.

This message is only shown when `std::make_unique<Resampler>` runs. Considering that there were 4 of them in sequence my guess is that the operation failed when the convolver tried to resample a kernel file. But we already have protection for `rate == 0` there. Strange...

Digitalone1:
Unfortunately I rebooted the system. I remember that I was loading a temporary preset with the deep noise reduction which uses the ladspa wrapper, but the app crashed.

wwmm:
> Unfortunately I rebooted the system. I remember that I was loading a temporary preset with the deep noise reduction which uses the ladspa wrapper, but the app crashed.

I see. The deepnoise plugin setup function is called by `resetHistory`. But there is no protection for bad rate values in the deepnoise setup function like there is for other plugins like the crystalizer. I will add a check for `rate == 0` there too.

PR #4898, opened 2026-02-28
Improve autoload preset dialog readability for long device names

antoinecellerier:
## Problem

Long device names such as `Alder Lake PCH-P High Definition Audio Controller` and device identifiers like `alsa_output.pci-0000_00_1f.3-platform-skl_hda_dsp_generic.HiFi__Speaker__sink` were truncated in the autoload tab, making it impossible to distinguish between multiple outputs on the same sound card.

## Changes

- **Wider dialog**: Increased preset dialog width from 30 to 40 grid units (90% of window width)
- **Text wrapping**: Autoload list entry labels (Device, Description, Route, Local preset) now wrap instead of being truncated with ellipsis
- **Layout constraints**: Added `maximumWidth` to contentItem FlexColumn and `Layout.fillWidth` to GridLayout so wrapping is properly constrained
- **Dropdown tooltip**: Added tooltip on hover for truncated device names in the combo box dropdown, using `contentItem.labelItem.truncated` to only show when text is actually elided

## Files changed

- `src/contents/ui/PresetsDialog.qml` — dialog width
- `src/contents/ui/PresetsAutoloadPage.qml` — wrapping, layout constraints, tooltip

## Use of AI

These changes were generated using GitHub Copilot CLI. I'll admit I do not have prior knowledge of QML nor this repo. From looking at the proposed changes and the resulting rendering in UI they seem right but do let me know if they're garbage.

## Screenshots

Before the changes: (easyeffects 8.1.2 on Debian Sid running Gnome)
image
image

After the changes:
image
image

### DIFF
--- a/src/contents/ui/PresetsAutoloadPage.qml
+++ b/src/contents/ui/PresetsAutoloadPage.qml
@@ -23,6 +23,7 @@ import QtQuick.Layouts
 import ee.pipewire as PW
 import ee.presets as Presets
 import org.kde.kirigami as Kirigami
+import org.kde.kirigamiaddons.delegates as Delegates
 import org.kde.kirigamiaddons.formcard as FormCard
 
 ColumnLayout {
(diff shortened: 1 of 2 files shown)


wwmm:
The code seems fine. And showing a tooltip when the label is not fully visible is a good idea. Thanks!

The only thing that concerns me a little is if the larger dialog width deviates too much from KDE's HIG. But the truth is that super long device names and small width are not a good match.

PR #4911, opened 2026-03-06
Merge

gunjan-ghangare:
Merge


Digitalone1:
Hello. Thanks for this PR. @wwmm will review it.

Anyway, please add missing licence text on top of the new files as Copilot says (you can copy it from other files).

gunjan-ghangare:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/wwmm/easyeffects/pull/4911#pullrequestreview-3901893780)

wwmm:
Tonight I will have time to take a look at this. For now what I think that has to change is the plugin name. Immersive surround will give users the wrong idea that they can improve their surrounding audio. But that is not what the plugin does. And easyeffects does not even support surround.

Virtual surround may also be misleading because usually the idea would be to process true surround so that a stereo audio would feel like surround. I think that "Fake Surround" is a name that indicates better what will be done.

wwmm:
I am doing some tests and so far I could not notice any difference in the sound processed by this plugin. At first I thought it was because I as using my desktop speakers. But even with my bluetooth headphone I can't notice any difference at all. There is level meters activity. But the plugin seems to do nothing.

PR #4992, opened 2026-04-16
fix: correct spelling typos and add codespell config

oxyzenQ:
Fixed typos:
- continously → continuously
- orignal → original
- unneded → unneeded
- convertion → conversion

Added .codespellrc for future automated typo checking."


wwmm:
Thanks!

>Added .codespellrc for future automated typo checking.

Interesting. What does the check? A Github action? OR a program installed in our PC?

oxyzenQ:
> Thanks!
> 
> > Added .codespellrc for future automated typo checking.
> 
> Interesting. What does the check? A Github action? OR a program installed in our PC?

It uses codespell, which is just a small CLI tool to detect common spelling mistakes in code and text.

The .codespellrc file is only a configuration file (ignore words, paths, etc.), it doesn’t run anything by itself.

It can be used in two ways:

Locally: developers install and run codespell on their machine
CI/CD: for example via a GitHub Actions workflow that runs it automatically on pull requests

In this PR, I only added the config.

PR #5001, opened 2026-04-21
Crosstalk Canceller plugin effect

alankila:
Crosstalk is defined as channels mixing together, and here it refers to the in-air mixing that occurs between the left and right speakers, as the sound from left speaker can reach around the head and enters the right ear and vice versa.

This has numerous effects. In the center channel, or when both speakers play the same sound, there is always a time-delayed copy being received in both ears as the channels mix together with a delay. This is akin to an early echo from nearby reflective surface. In case of mono sound from single channel, the contralateral ear receives the sound with time delay that informs about the genuine position of the loudspeaker, and thus adds a localization cue which usually restricts the sound origin to the plane betwen the speakers.

Crosstalk of this type can be reduced by estimating the sound perception in the right ear and playing correctly timed cancellation waveform from the right channel which accounts to some degree by HRTF estimate the difference in the planar incident angle of the sound waveform. Thus, it tends to change the sound of the phantom center and also tends to wrap the sound more around the listener's head, widening the soundstage.

This can be considered a digital version of the a large sheet of absorbing material in front of the listener that absorbs sound crossing the center plane.

High end commercial approaches such as BACCH in addition provide more tuning and far more complicated processing, and realtime camera that observes head position and tailors the timing to create crosstalk cancellation at the listening location. This is a very simple version, with an adjustable but user controlled crosstalk timing. The time parameter should be set to around 320 us for equilateral triangle setup, and grows if listener is closer to speakers, and reduces if further away.

The effect should be the last in the processing chain. It is assumed that the speakers face the listener at similar angle, are at similar distance, and are of same type, and therefore the direct sound characteristics will be the same. Other processing can interfere with the cancellation waveform.


alankila:
I used AI (Qwen 3.5 122B) to perform the forward porting from the past GTK+ version to this Qt version. I've only run it for a day against Ubuntu's easyeffects, but it seems to work the same.

I then forward ported it once more to the master, but I haven't been able to build and execute this master branch on the 26.04 Ubuntu either. I hope that you have patience of reviewing and testing out this PR version, as I still don't have all the dependency versions available.

wwmm:
I am not familiar with the algorithm. So I can't judge that part. As far as the code is concerned it is almost fine. The only bug I saw so far is presets parameters not being applied. But it should not be hard to fix. I will take a look at it. The other changes are more about making the code in line with what is done on other plugins. And fix some non critical clang-tidy warnings.

Could you add a explanation about this plugin to https://github.com/wwmm/easyeffects/tree/master/src/contents/docs/plugins ? Those files are shown in our online manual https://wwmm.github.io/easyeffects/

wwmm:
And before I forget, thanks for the contribution! :-)

wwmm:
Based on you explanation it may be better to show to the user `Crosstalk Canceller` instead of Localization Cue Correction

wwmm:
I renamed the plugin to Crosstalk Canceller. I also added an empty file that should be used for its documentation https://github.com/wwmm/easyeffects/blob/master/src/contents/docs/plugins/crosstalkcanceller.md

PR #5070, opened 2026-05-09
Expand spectrum analyzer frequency limits.

troyBORG:
Allow minimum frequency down to 1 Hz and maximum frequency up to 25 kHz.

My audio setup is....

| Item | Model | Frequency response |
|---|---|---:|
| Subwoofer | Klipsch Reference Premiere RP-1200SW | 16.5 Hz – 138 Hz ±3 dB |
| Bookshelf speakers | Klipsch Reference Premiere RP-500M II | 50 Hz – 25,000 Hz ±3 dB |

My sub can go down alot lower then the 20 Hz limit that was set.  So I thought.  Why not let us go all the way down?
Same with the upper limit.  It was only 20 Khz.  But my speakers, (well probably the tweaters) can go higher..

### DIFF
--- a/src/contents/kcfg/easyeffects_db_spectrum.kcfg
+++ b/src/contents/kcfg/easyeffects_db_spectrum.kcfg
@@ -48,14 +48,14 @@
         </entry>
         <entry name="minimumFrequency" type="Int">
             <label></label>
-            <min>20</min>
+            <min>1</min>
             <max>21900</max>
             <default>20</default>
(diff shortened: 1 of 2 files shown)


troyBORG:
Oh.  I found a bug when I tried setting it to 25000 it just crashed..   
I should have tested before doing pull request..   Fixing now

wwmm:
It seems fine. Thanks!

PR #5081, opened 2026-05-14
Fixed RNNoise timeout on empty input

jmarewski:
Fixed a bug where RNNoise exceeded the real‑time time limit (rttime), causing EasyEffects to be terminated.

## Reproduce the "crash"
 - turn on RNNoise
 - start a recording application
 - mute/unmute the input device a few times

### DIFF
--- a/src/rnnoise.cpp
+++ b/src/rnnoise.cpp
@@ -190,7 +190,11 @@ void RNNoise::process(std::span<float>& left_in,
   }
 
 #ifdef ENABLE_RNNOISE
-  if (!rnnoise_ready) {
+  auto eps = 1e-6F;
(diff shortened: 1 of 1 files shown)


Digitalone1:
Thank you. `eps` can be made `constexpr`.

wwmm:
> Fixed a bug where RNNoise exceeded the real‑time time limit (rttime), causing EasyEffects to be terminated.
> ## Reproduce the "crash"
> 
>     * turn on RNNoise
> 
>     * start a recording application
> 
>     * mute/unmute the input device a few times

I can't reproduce the crash on my computer. But if the issue is rnnoise spending too much time in pipewire's realtime thread because of denormalized numbers I imagine that the crash is not going to happen on all cpu models.

In any case the fix  seems simple enough to be merged.

wwmm:
> Thank you. `eps` can be made `constexpr`.

I did that now.

PR #5093, opened 2026-05-20
Fix: Replace semicolon with comma in QT_QPA_PLATFORM to prevent startup crash

ghost:
This PR resolves an issue where EasyEffects crashes on startup with a `SIGABRT` due to the Qt 6 platform plugin failing to initialize. 

Some desktop environments (or environment wrappers) mistakenly set `QT_QPA_PLATFORM` using a semicolon as a delimiter (e.g., `wayland;xcb`). While Qt 5 tolerated this format, Qt 6 strictly requires a comma-separated list (e.g., `wayland,xcb`) for fallback platform plugins. When encountering a semicolon, Qt 6 fails to parse the string, cannot find the appropriate plugin, and aborts immediately on launch.

To prevent this, this patch adds a quick check in `main.cpp` before the `QApplication` is initialized. It dynamically replaces any semicolons found in `QT_QPA_PLATFORM` with commas, ensuring that Qt 6 can parse the fallback plugins correctly.

### Related Issues
* Fixes Bugzilla Bug: [#2479830](https://bugzilla.redhat.com/show_bug.cgi?id=2479830)

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -344,6 +344,12 @@ int main(int argc, char* argv[]) {
   QCoreApplication::setAttribute(Qt::AA_ShareOpenGLContexts);
   QLoggingCategory::setFilterRules("easyeffects.debug=false");
 
+  auto qt_qpa_platform = qgetenv("QT_QPA_PLATFORM");
+  if (qt_qpa_platform.contains(';')) {
+    qt_qpa_platform.replace(';', ',');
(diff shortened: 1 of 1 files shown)


wwmm:
I was not aware of it. Ideally these wrappers should stop doing that. But it will probably be easier to workaround this on our side.

wwmm:
Thanks!

PR #5097, opened 2026-05-21
flatpak: Fix qt6-graphs duplicate libraries ldconfig warning

### LINKED ISSUE #5096: ldconfig warnings on launch: QtGraphs libraries bundled as plain files in /app/lib instead of symlinks
### EasyEffects Version

8.2.2

Flatpak (Flathub)

### Distribution

Debian 13.5

### Describe the bug

Launching EasyEffects emits two `ldconfig` warnings on first run per session:

```
ldconfig: /app/lib/libQt6Graphs.so.6 is not a symbolic link
ldconfig: /app/lib/libQt6GraphsWidgets.so.6 is not a symbolic link
```

The QtGraphs libraries are bundled twice in the Flatpak with different layouts.

In `/app/lib/x86_64-linux-gnu/` the SONAME chain is correct (symlinks to the versioned files):

```
libQt6Graphs.so.6        -> libQt6Graphs.so.6.10.2
libQt6GraphsWidgets.so.6 -> libQt6GraphsWidgets.so.6.10.2
```

But in `/app/lib/` the same SONAMEs exist as plain regular files, with no adjacent `.so.6.10.2` and not being symlinks:

```
-rwxr-xr-x libQt6Graphs.so.6        (3044512 bytes)
-rwxr-xr-x libQt6GraphsWidgets.so.6 (177064 bytes)
```

When `ldconfig` scans `/app/lib`, it finds versioned SONAMEs that by convention should be symlinks and warns. Impact is cosmetic: the libraries are still registered, the app starts, and presets load correctly.

Flatpak commit: `656e5e1b94e093e67edc8c8a8f1ff223e8bd6d5d3b5ad323d7f615e117082cec`

### Expected Behavior

No `ldconfig` warnings on launch. The QtGraphs libraries should follow the standard SONAME convention in `/app/lib` as well — either by not duplicating them as flat regular files (keeping only the correctly-symlinked copies in `/app/lib/x86_64-linux-gnu/`), or by reconstructing the proper chain (`libQt6Graphs.so.6 -> libQt6Graphs.so.6.10.2` plus the versioned file) in `/app/lib`.

### Debug Log

Not a `G_MESSAGES_DEBUG` log — the warnings come from `ldconfig` inside the sandbox, on stderr at launch:

```
$ flatpak run com.github.wwmm.easyeffects
ldconfig: /app/lib/libQt6Graphs.so.6 is not a symbolic link
ldconfig: /app/lib/libQt6GraphsWidgets.so.6 is not a symbolic link
```

### Additional Information

The warnings appear only on the first launch per session, since `ldconfig` builds the sandbox cache once. The epoch mtime (1970) and link count 2 on the flat files are consistent with OSTree hardlink dedup, confirming they originate from the build rather than the host.

Verified with:

```
find /var/lib/flatpak -name 'libQt6Graphs*.so.6' -exec ls -la {} +
```

ghost:
Fixes #5096.

### Description

Launching EasyEffects previously emitted two `ldconfig` warnings on the first run per session regarding the `Qt6Graphs` libraries not being symbolic links:
```text
ldconfig: /app/lib/libQt6Graphs.so.6 is not a symbolic link
ldconfig: /app/lib/libQt6GraphsWidgets.so.6 is not a symbolic link

### DIFF
--- a/util/flatpak/easyeffects-modules.json
+++ b/util/flatpak/easyeffects-modules.json
@@ -14,7 +14,7 @@
                 }
             ],
             "post-install": [
-                "install $FLATPAK_DEST/lib/*-linux-gnu/*.so $FLATPAK_DEST/lib/*-linux-gnu/*.so.* $FLATPAK_DEST/lib"
+                "cp -a $FLATPAK_DEST/lib/*-linux-gnu/*.so $FLATPAK_DEST/lib/*-linux-gnu/*.so.* $FLATPAK_DEST/lib"
             ]
         },
         {


wwmm:
Ok. It seems fine to do that.

tuliom:
@wwmm Heads up: this PR got created by an autonomous agentic AI. This system has produced incorrect diagnoses and fixes in multiple cases.

I believe this PR is OK.
I reviewed the other PR they created and noticed you have already reverted it.

I suggest to be careful with more contributions coming from that Github account. ;-)

More information:
- https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/SFVETHOYKQAO7KKLEXCK4IBT4WVPRE6F/
- https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/PRTKEA2XJW2EILP3OVV3IQVHCFYX4TQG/

wwmm:
> I suggest to be careful with more contributions coming from that Github account. ;-)

I suspected it was something like that. Thanks for confirming it :-).

PR #5109, opened 2026-05-24
Prevent the saturation icon from shifting the right-aligned toolbar information.

pparaxan:
Fixes an issue where the saturation icon moves the information aligned at the right of the toolbar found below.

This wasn't an issue prev., as the information was at the left, and the saturation icon had all the space for itself.


wwmm:
Thanks! I forgot to check how that was.

PR #5143, opened 2026-06-19
Add "Duplicate" button

ChacePratt:
# Why?
Having a duplicate button would save significant amounts of time while adding multiple of the same effect (I found that when I was adding filters for EQing my speakers it would be a nice UX improvement to have a "Duplicate" button)

# To-Do
1. Test further (I was unable to successfully build a testing Flatpak of EasyEffects with every available effect - I'm new to this stuff, and I tried for hours, I promise...) 
**There's a chance one effect being duplicated will break the logic I built!**
**The try catch I made could be used to easily see if there is a problem while copying values of effects**
2. Add translations (currently every translation is marked with "empty" that isn't English)

### DIFF
--- a/po/af/easyeffects.po
+++ b/po/af/easyeffects.po
@@ -1230,12 +1230,17 @@ msgctxt "@info:placeholder"
 msgid "No description set"
 msgstr ""
 
-#: src/contents/ui/DelegatePluginsList.qml:119
+#: src/contents/ui/DelegatePluginsList.qml:160
 #, fuzzy
 msgid "Toggle this effect"
 msgstr "EasyEffects"
 
-#: src/contents/ui/DelegatePluginsList.qml:127
+#, empty
+#: src/contents/ui/DelegatePluginsList.qml:168
+msgid "Duplicate this effect"
+msgstr ""
+
+#: src/contents/ui/DelegatePluginsList.qml:174
 #, fuzzy
 msgid "Remove this effect"
 msgstr "EasyEffects"
--- a/po/ar/easyeffects.po
+++ b/po/ar/easyeffects.po
@@ -1283,11 +1283,16 @@ msgctxt "@info:placeholder"
 msgid "No description set"
 msgstr ""
 
-#: src/contents/ui/DelegatePluginsList.qml:119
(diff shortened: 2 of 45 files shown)


wwmm:
This approach will probably work. But it is more efficient to handle the duplicated instance creation and data copy in the c++ backend. Being more specific in the [db_manager](https://github.com/wwmm/easyeffects/blob/master/src/db_manager.cpp) class. After the call to `create_plugin_db` and the data copy is done it would be just a matter of updating the database plugins list array. Changes to it are automatically propagated to easyeffects window. There is no need to deal with the qml plugins list directly.

One of the reasons why I think it is better to handle this on c++ side is that some plugins like the multiband gate and compressor really have a huge amount of parameters. Although it is mostly just copy doing all of this in qml thread is probably not going to be nice to weak computers.

wwmm:
>After the call to create_plugin_db and the data copy is done it would be just a matter of updating the database plugins list array.

Actually it is not even needed to call it. After an update to the plugins list the `::pluginsChanged` callbacks already defined in this class are automatically called. Inside of them could be the call that copies the data from one plugins instance to another.

ChacePratt:
> This approach will probably work. But it is more efficient to handle the duplicated instance creation and data copy in the c++ backend. Being more specific in the [db_manager](https://github.com/wwmm/easyeffects/blob/master/src/db_manager.cpp) class. After the call to `create_plugin_db` and the data copy is done it would be just a matter of updating the database plugins list array. Changes to it are automatically propagated to easyeffects window. There is no need to deal with the qml plugins list directly.
> 
> One of the reasons why I think it is better to handle this on c++ side is that some plugins like the multiband gate and compressor really have a huge amount of parameters. Although it is mostly just copy doing all of this in qml thread is probably not going to be nice to weak computers.

I’ll work on getting this implemented. I’m on vacation right now and have no access to a laptop but when I get home I’ll update db_manager from the onDuplicate QML function.
I had initially considered doing it this way but I guess I missed the plugin list in db_manager. I’m new to C++ so I’m still working out how you program with it.

Thank you for the suggestion!!

PR #5164, opened 2026-07-04
Add audio sharing global shortcut

Freezermod:
## Summary

This adds a new global shortcut for toggling audio sharing:

- `Meta/Super + Ctrl + O` toggles audio sharing ( You may change it if you find it's too close to the other one )
- the shortcut is registered through the existing GlobalShortcuts portal flow
- the shortcut help sheet now lists the new binding
- shortcut portal descriptions now use the translated description field instead of the internal shortcut id

## Details

The new shortcut toggles `DbStreamOutputs::linkToVirtualSource`, which is the same setting used by the existing
audio sharing UI action.

## Testing

Tested locally with:

- `cmake --build build-arch-test`
- `DESTDIR=/tmp/easyeffects-arch-test cmake --install build-arch-test`
- Flatpak build using the existing development Flatpak manifest
- Flatpak appstream test passed

Also manually tested the generated build and confirmed the shortcut works.

### DIFF
--- a/src/contents/ui/ShortcutsSheet.qml
+++ b/src/contents/ui/ShortcutsSheet.qml
@@ -399,6 +399,46 @@ Kirigami.OverlaySheet {
                     text: i18n("Toggle microphone monitoring") // qmllint disable
                 }
 
+                RowLayout {
+                    Layout.alignment: Qt.AlignRight
+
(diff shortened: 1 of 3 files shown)


wwmm:
Thanks!

PR #5232, opened 2026-08-08
fix preset autoload on volume changes

### LINKED ISSUE #4729: Preset auto reloads when changing volume with fallback preset enabled
### EasyEffects Version

8.0.9

Arch (easyeffects)

### Distribution

Arch Linux

### Describe the bug

When changing the volume with preset autoloading enabled the preset automatically reloads every tick causing audio stutters. This behaviour only happens if the default preset is enabled disabling it fixes the issue.

https://github.com/user-attachments/assets/9c166624-f75a-46ed-9416-f0c36b3dea14

### Expected Behavior

The presets switches only when actually switching devices not always

### Debug Log

_No response_

### Additional Information

_No response_

MagnusCardell:
Fixes #4729.

PipeWire's Route param includes the profile device it belongs to. EasyEffects currently discards it and applies each enumerated route to every node on the card. Volume changes update route volume props, so cards with several nodes cycle through unrelated routes and reload autoload presets.

Keep card.profile.device for each node and only apply its matching route. Fall back to the old behavior when either id is missing.

### DIFF
--- a/src/pw_device_manager.cpp
+++ b/src/pw_device_manager.cpp
@@ -167,11 +167,12 @@ void DeviceManager::on_device_event_param(void* object,
 
   enum spa_direction direction {};
   enum spa_param_availability available {};
+  int32_t route_device_id = -1;
 
   if (spa_pod_parse_object(param, SPA_TYPE_OBJECT_ParamRoute, nullptr, SPA_PARAM_ROUTE_direction,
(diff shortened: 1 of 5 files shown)


wwmm:
As my system did not have the preset reload problem I can't say how good this is to fix #5232. But using the profile device tag seems a good thing to do. And as far as I can see there is no regressions on my side.

wwmm:
I wonder since when `card.profile.device` became part of the node properties. I do not remember seeing it there before. It is a good thing they did that.

PR #5246, opened 2026-08-20
feat: implement Mid-Side Equalizer effect

### LINKED ISSUE #5234: MidSide Equalizer
Hello! Is it possible to add the mid side equalizer from https://lsp-plug.in/?page=manuals&section=para_equalizer_x32_ms
This would be very helpful for advanced effects, thank you!

Additional: in another issue I saw that it said easyeffects does not support adding new effects because it would need implementation of the GUI, but potentially could the support simply be added by having the plugin configured from the Show native window button (not EasyEffects) itself? This would prevent any GUI coding

kairosci:
Adds a Mid-Side Equalizer plugin enabling independent parametric equalization of the mid and side audio channels. The implementation includes a QML interface, preset management, APO import support, and full database integration with automatic plugin registration and icon handling. This expands the available processing options for advanced audio manipulation and stereo imaging control. Closes #5234.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -43,6 +43,7 @@ qt_add_qml_module(easyeffects
         effects_base.hpp
         equalizer.hpp
         exciter.hpp
+        midside_equalizer.hpp
         expander.hpp
         filter.hpp
         gate.hpp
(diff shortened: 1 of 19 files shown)


wwmm:
I did some tests and it seems fine. The only thing that seems wrong is the `Link Mid/Side` button. When I enable it it actually disables the link between mid and side in the plugin native window. The boolean binding in this button needs to be inverted. But that is something I can do after the merge.

wwmm:
Or maybe just changing the label is enough

wwmm:
I changed the label to `Split mid/side`. Thanks for the PR!

PR #5248, opened 2026-08-22
Only hide window on failure

### LINKED ISSUE #5243: easyeffects --last-loaded-preset hide window
### EasyEffects Version

8.2.8

Arch (easyeffects)

### Distribution

Archlinux

### Describe the bug

`easyeffects --last-loaded-preset input` / `easyeffects --last-loaded-preset output` hide the easyeffect window by default.

Checking the code, this is because of a `Q_EMIT onHideWindow();` in https://github.com/wwmm/easyeffects/blob/97605d28264f5877e6674a194bc5228b2c23bfbf/src/command_line_parser.cpp#L130C4-L130C27

`easyeffects --last-loaded-presets` does not have this issue.

Is there a reason for hiding the window when using the `--last-loaded-preset` flag ?
I can create a PR to fix this if the `Q_EMIT onHideWindow();` is not needed.

### Expected Behavior

Calling `easyeffects --last-loaded-preset input` / `easyeffects --last-loaded-preset output` should not close the easyeffect window if open.

BarbUk:
With no `Q_EMIT onHideWindow();`:
- window is not closed or opened when calling `easyeffects --last-loaded-preset` with a correct parameters or without parameter.
- window is opened or focused if already opened when calling `easyeffects --last-loaded-preset` with a wrong parameter: `easyeffects --last-loaded-preset test` for exemple.

This change propose to only use `Q_EMIT onHideWindow();` on failure.

Fixes: https://github.com/wwmm/easyeffects/issues/5243

### DIFF
--- a/src/command_line_parser.cpp
+++ b/src/command_line_parser.cpp
@@ -127,11 +127,10 @@ void CommandLineParser::process_events() {
       std::cout << i18n("Must specify preset type: input/output.").toStdString() << '\n';
     }
 
-    Q_EMIT onHideWindow();
-
     if (ok) {
       QCoreApplication::exit(EXIT_SUCCESS);
     } else {
+      Q_EMIT onHideWindow();
       QCoreApplication::exit(EXIT_FAILURE);
     }
   }


wwmm:
I see. In this case it should be fine to do this.

PR #5249, opened 2026-08-22
feat: add explicit mnemonics to all QML action texts

### LINKED ISSUE #4504: [Feature request] Explicit mnemonics in action texts
As was [posted](https://github.com/wwmm/easyeffects/discussions/4500) in discussions, I'd like to suggest adding explicit mnemonics in action texts.

It seems that the controls without explicit mnemonics are [automatically assigned one](https://blog.broulik.de/2025/01/mnemonics-mnemonics-everywhere/#:~:text=KDE%20applications%2C%20both%20written%20in%20Qt%20Widgets%20and%20Qt%20Quick%2C%20automatically%20assign%20mnemonics%20for%20most%20controls%20that%20don%E2%80%99t%20have%20one%20explicitly%20set%2E) in Qt Widgets and Qt Quick, which is nice to have, but also leaves a hole for translators to fill. 

Another problem is that sometimes mnemonics change based on contexts, which is bothering to me.

For example, in the main UI "O" and "I" are typically assigned to the "Output" and "Input".
O for Output and I for Input

However, if I toggle the right bottom corner menu of Equalizer and press Alt, "I" would be re-assigned to the "Import" here, albeit the menu would be immediately closed.
I for Import?

As the result, the "Input" in the main UI are stripped of its mnemonic.
No I for Input
And "O" for "Output" are substituted too, by whom I don't know.

Generally speaking, explicit mnemonics are stable thus better for ergonomics in my opinion, and are recommended in the [KDE documentations](https://develop.kde.org/docs/getting-started/kxmlgui/using_actions/#setting-qaction-properties) too.

Plus, they can also be the handle for translators to put up localized mnemonics without guessing.

Broader inputs are welcomed. And we could come up with a draft of mnemonics for the most used controls.

kairosci:
Adds explicit `&` mnemonics to all user-facing action, button, and menu item texts in the QML UI so Qt stops auto-assigning conflicting shortcuts per context. Covers 42 files across main toolbar, tray icon menu, per-effect toolbars, preset dialogs, and blocklist dialogs.

Fixes #4504

### DIFF
--- a/src/contents/ui/Main.qml
+++ b/src/contents/ui/Main.qml
@@ -375,6 +375,7 @@ Kirigami.ApplicationWindow {
                     checked: DbMain.visiblePage === 0 && DbStreamOutputs.visiblePage === 1
                     display: segmentedButton.display
                     visible: DbMain.visiblePage !== 2
+                    Kirigami.MnemonicData.controlType: Kirigami.MnemonicData.DialogButton
                     onClicked: {
                         DbMain.visiblePage = 0;
                         DbStreamOutputs.visiblePage = 1;
@@ -390,6 +391,7 @@ Kirigami.ApplicationWindow {
                     checked: DbMain.visiblePage === 1 && DbStreamInputs.visiblePage === 1
                     display: segmentedButton.display
                     visible: DbMain.visiblePage !== 2
+                    Kirigami.MnemonicData.controlType: Kirigami.MnemonicData.DialogButton
                     onClicked: {
                         DbMain.visiblePage = 1;
                         DbStreamInputs.visiblePage = 1;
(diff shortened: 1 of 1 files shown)


wwmm:
I am doing some tests and the explicit mnemonics actually make things worse. I wonder if we are facing a KDE framework limitation. This is as things are right now
image

And this is how it is with the patch
image

The main visible difference is that more than one letter is highlighted by the `Alt` key. But with the current patch some buttons can't be triggered. For example the `Excluded apps` and `Share audio`. As things are right now they work with mnemonics. But with the patch it does not seem possible. Or it may be just that I do not know how to use mnemonics when more than one letter is highlighted.

kairosci:
I revert hardcoded mnemonics and prioritize main actions (Output, Input, Players, Recorders) in Main.qml using Kirigami.MnemonicData.DialogButton to prevent shortcut conflicts with submenus.

wwmm:
> I revert hardcoded mnemonics and prioritize main actions (Output, Input, Players, Recorders) in Main.qml using Kirigami.MnemonicData.DialogButton to prevent shortcut conflicts with submenus.

Hum... It does not seem to make much difference or cause side effects. But it seems strange to set `Kirigami.MnemonicData.DialogButton` to something that isn't a dialog

PR #5283, opened 2026-09-04
Monitoring volume control

Obsessee:
Adds a volume slider for input monitoring via a dropdown arrow.

There's duplication between setNodeVolume and setNodeMonitorVolume. I considered merging them, but I decided to leave it up for review.

Preview of the button:

Preview when clicked:


wwmm:
I did some tests and it seems fine. Qml linter complaied about the x and y Popup coordinates being set when the Popup is managed by a Layout bu this is something that I can take a look at after merging this PR. Thanks!

Issue #5120, opened 2026-06-02
[BUG] Segfault Crash


ArgentVASIMR:
### EasyEffects Version

8.2.4

### What package are you using?

Flatpak (Flathub)

### Distribution

Linux Mint 22 "Wilma"

### Describe the bug

Program randomly crashes at unknown time. Running in terminal with `--debug` yielded nothing of interest, except for a `[SIGSEGV]` error signal displayed on my Fish terminal, clearly showing that this is a segmentation fault.

The crash has occurred arbitrarily from various unrelated actions, including:
- Toggling all effects with the button in the top left.
- Toggling input monitoring.
- Rearranging effects.
- Switching to a different tab.
- Opening a plugin's native window.
- Stopping a recording in OBS, where "EasyEffects Source" was in use.

The cause is unknown. Reproducing the issue is difficult. I reproduced it by rapidly toggling all effects with the top left button. Sometimes a segfault happened in a matter of seconds, sometimes it took several minutes. The only clear pattern is that _some action_ occurs, then the segfault occurs.

### Debug Log

<details><summary>Debug Log (full backtrace of segmentation fault from GDB)</summary><br/>

```
#0  QV4::Heap::ArrayData::vtable (this=<optimized out>) at /run/build-runtime/qt6-qtdeclarative/src/qml/memory/qv4heap_p.h:46
#1  QV4::Heap::ArrayData::length (this=0x1b8fa1e0ff3) at /run/build-runtime/qt6-qtdeclarative/src/qml/jsruntime/qv4arraydata_p.h:88
#2  QQmlVMEResolvedList::size (this=0x7ffe0c727c50) at /run/build-runtime/qt6-qtdeclarative/src/qml/qml/qqmlvmemetaobject_p.h:58
#3  list_count (prop=<optimized out>) at /run/build-runtime/qt6-qtdeclarative/src/qml/qml/qqmlvmemetaobject.cpp:123
#4  0x0000710d485c9fc4 in QQmlListProperty<QObject>::toList_impl<QList<QObject*> > (this=0x710d2c147c78) at /run/build-runtime/qt6-qtdeclarative/src/qml/qml/qqmllist.h:104
        size = <optimized out>
        result = {<QListSpecialMethods<QObject*>> = {<QListSpecialMethodsBase<QObject*>> = {<No data fields>}, <No data fields>}, d = {d = <optimized out>, ptr = <optimized out>, size = <optimized out>}}
[...]

PR #5306, opened 2026-09-13
Fix application crash in Convolver view when multiple irs files are updated in short sequence

antoinecellerier:
fix ListModel::update emitting row changes inside a model reset

Fixes the QML crash in #5120 (`QmlListWrapper::toVariant` under `QQmlComponentPrivate::setInitialProperties` inside the incubator) by making `ListModel::update()` a plain sequence of row operations instead of row operations nested inside a model reset.

ListModel::update() wrapped calls to append() and remove() in beginResetModel()/endResetModel(). Qt forbids emitting other change signals during a reset, and the QSortFilterProxyModel plus the QML views on top of it (the Convolver page's impulse combo boxes) were left with row bookkeeping that no longer matched the model. Two directory events a few milliseconds apart run update() twice back to back: the second reset arrives while the views are still incubating delegates for the first, and the incubator dereferences a stale list (QmlListWrapper::toVariant under setInitialProperties), the crash reported in #5120.
    
append() also announced the new row at size()-1 instead of size(), and every insert/remove followed up with a whole-range dataChanged() that is redundant with rowsInserted/rowsRemoved and invalid once the list is empty. update() now removes then appends with plain row operations, append() announces the right row, remove(path) ignores a path that is not in the list, and update() ends with one dataChanged() over the surviving rows so a file replaced under the same name still refreshes.

## Repro

With the window open on the Convolver page (any impulse loaded or not):

```sh
cd ~/.local/share/easyeffects/irs && touch a.irs; touch b.irs
```

On 8.2.9-61 (735d551f6, Debug build, Qt 6.10.2, Kirigami 6.28, Debian) this segfaults within a few seconds, three runs out of three. Two separate commands matter: the directory watch fires once per `touch`, a couple of milliseconds apart, so `update()` runs twice back to back. A single `touch` never crashed it. With this applied the same two commands run clean.

Backtrace of the crash (top frames, same as #5120):

```
#0  ?? () from libQt6Qml.so.6
#1  QV4::QmlListWrapper::toVariant() const
#3  QV4::ExecutionEngine::toVariant(QV4::Value const&, QMetaType, bool)
#4  QV4::QObjectWrapper::setProperty(...)
#5  QV4::QObjectWrapper::setQmlProperty(...)
#6  QV4::QObjectWrapper::virtualPut(...)
#7  QQmlComponentPrivate::setInitialProperties(...)
#9  QQmlIncubatorPrivate::incubate(QQmlInstantiationInterrupt&)
#10 QQmlIncubationController::incubateFor(int)
#13 QQuickWindow::event(QEvent*)
#16 QPlatformWindow::deliverUpdateRequest()
#28 main (main.cpp:481)
```

How it was found: a converter that writes a few impulse files and presets in one go (temp file + rename, so two events per file) crashed EasyEffects for a user every time the Convolver page was showing. https://github.com/antoinecellerier/speaker-tuning-to-easyeffects/issues/95

## Use of AI

The crash and commit were originally built using Claude Code, and some clean-ups by me. I have verified claims in Qt documentation, as well as confirmed the repro. Note I am unfamiliar with the Qt programming model so might have missed subtleties. From past PRs I understand that to be ok in this repo but do let me know if this is an issue. Feedback is welcome.


wwmm:
>Qt forbids emitting other change signals during a reset

I wonder if that is maybe a Qt bug. The emit call was not done inside the begin/end reset calls. Qt should wait for the proper moment internally. Strange...

It was hard to reproduce this bug here on my computer but I did see it. This PR does seem to fix the problem. And I don't have a better idea right now. So let's merge it.

antoinecellerier:
In the update() method the emit calls from the append and remove methods happened between the two being/end markers.
Hopefully this works for the reporters on #5120.
Thanks for the quick merge!

PR #5308, opened 2026-09-14
Import community presets: add support for SOFA addons

Digitalone1:
Testing some newer community presets I found that presets using SOFA impulse responses for Convolver cannot be imported.

That's because we use the filename without extension into the JSON preset and the code for the import stage was only handling `.irs` extensions for Convolver. When support for SOFA files was added, the import of community presets was not refactored.

To be honest, I don't remember exactly why we chose to save the irs filename without the extension. I barely remember it was to move away from the full path which was exposing the local paths of users, but if we chose to hide also the extension there should be a good reason which I don't get at the moment.

Anyway, the CP import code has been changed to store the vector with the original contents of `kernel-name` (before we were appending the extensions). Then we append the extensions.

For the RNNoise, it's quite the same as before. For the Convolver, I added another for loop to search `irs` and `sofa` extension. The search succeed if at least one file is found (irs or sofa; or both).

I tested it importing a SOFA community presets and it's working. Unfortunately I didn't have time to test a RNNoise community preset, so I will do it tomorrow or in the next days. I'll leave the PR here since it has to be reviewed anyway.

@wwmm If you can test a RNNoise community preset, please do it with these changes.


Digitalone1:
Okay, maybe I'm starting to remember why we made that choice. That's because the Convolver supports also `wav` files, so in the Convolver initialization we search for irs and wav files when a filename is given.

Then, when SOFA support was implemented, we added the search for sofa extension also.

But, now that I'm thinking about it, since we can read also wav files, we have to search also for wav extensions in the community presets import stage. I forgot it... 😕

Digitalone1:
Tested the import of RNNoise addons. It works. 

> But, now that I'm thinking about it, since we can read also wav files, we have to search also for wav extensions in the community presets import stage. I forgot it... 😕

I gave a look the the presets irs manager code and found out that wav files imported by the user are rename with the `irs` extension, so we don't need to lookup for `wav` files in the import stage of community presets.

I think this can be merged.

PR #5310, opened 2026-09-16
Increase ceiling for input monitoring volume

Obsessee:
I have a use case where going above 100% input monitoring volume is useful. Having this included in the main project would save me from maintaining a separate build, though I completely understand if you’d prefer not to include it.


wwmm:
It should be fine as long as the user is careful.

PR #5315, opened 2026-09-17
Add CLI controls for microphone monitoring and audio sharing

Freezermod:
Adds the following commands using the existing local client/server protocol and global-bypass CLI architecture:

- `--microphone-monitoring <state>`
- `--microphone-monitoring-toggle`
- `--audio-sharing <state>`
- `--audio-sharing-toggle`

State arguments match `--bypass`: `1` enables, `2` disables, and `3` prints the current state. Commands update the running instance through the existing setters, without externally modifying configuration files.

The new controls preserve window visibility unless `--hide-window` is explicitly requested. This also fixes unwanted window hiding with `--bypass 1/2/3`; `--bypass-toggle` is unchanged. Includes translated help descriptions, protocol documentation, and integration tests.

Validation: build passed; tests cover live state changes, queries, toggles, invalid requests, PipeWire connectivity, and window visibility. Also tested interactively and alongside the separate routing fix.

**AI disclosure:** These changes and tests were developed with assistance from OpenAI Codex. I manually tested the builds and reviewed the PR descriptions.


wwmm:
I did some tests and it seems fine. Thanks.

PR #5316, opened 2026-09-17
Prevent audio sharing from duplicating playback through microphone monitoring

Freezermod:
Microphone monitoring currently listens to Easy Effects Source, which also receives desktop audio when sharing is enabled. This creates a second playback path to the speakers, reproducing an approximately +6 dB increase. When monitoring includes output effects, it creates a feedback path into the output pipeline.

This change monitors the processed microphone before desktop audio is mixed into the virtual source. A separate monitoring stage preserves independent monitoring volume without changing the microphone signal sent to recording applications. Both monitoring destinations remain supported, and monitoring keeps the input pipeline active without recording clients.

Validation: build passed; isolated audio tests verify unchanged desktop volume, independent microphone monitoring volume, both monitoring destinations, inactivity handling, and cleanup. Also verified interactively, independently and with the CLI changes.

This PR is independent of the CLI controls and can be reviewed separately.

**AI disclosure:** These changes and tests were developed with assistance from OpenAI Codex. I manually tested the builds and reviewed the PR descriptions.


wwmm:
I did some tests and it seems fine.

PR #5319, opened 2026-09-18
Silence QThreadStorage warnings on --help and --version

antoinecellerier:
QCommandLineParser::process() answers both options with ::exit() from inside main(), finalizing the shared libraries while QApplication is still alive. KConfig's QThreadStorage caches, filled in by KIconTheme::initTheme(), lose their destructors before the main thread releases its thread local storage, so Qt prints two "QThreadStorage: entry N destroyed before end of thread" lines.

Print both options ourselves and return from main() instead. Cosmetic: output and exit codes are unchanged, and everything else, including --help-all and malformed command lines, still goes through Qt.

Prior to the change, --version prints:
```
% easyeffects --version
easyeffects 8.2.8
QThreadStorage: entry 2 destroyed before end of thread 0x556bda3944e0
QThreadStorage: entry 1 destroyed before end of thread 0x556bda3944e0
```

With the change:
```
% ./build/src/easyeffects --version
easyeffects 8.2.9
```

This is mostly a cosmetic change. Please discard if you feel it's not worth it. Happy to amend with feedback as well.

Co-Authored-By: Claude Opus 5


wwmm:
>This is mostly a cosmetic change. Please discard if you feel it's not worth it. Happy to amend with feedback as well.

It is not a problem doing this. I used to see this annoying warning in the past indeed. But now that I am paying attention to it I see it does not happen anymore on my updated Arch Linux system. I am not sure if it was a Qt or a KDE Frameworks update. But the issue seems to have been fixed somewhere in the system libraries.

antoinecellerier:
Ah interesting. Maybe not worth pursuing then if it'll eventually disappear.

PR #5320, opened 2026-09-18
Fix Check QML format CI failures

antoinecellerier:
Ran ./util/qmlformat.sh in an attempt to fix CI which seems to have
started failing consistency since
https://github.com/wwmm/easyeffects/actions/runs/32455697344 was merged.

I do not know if this will work but it's cheap enough to try.


wwmm:
I think this might actually be a Qt bug. I do not remember this being a thing before its last release. And it is really odd it just silently fails without telling us the reason.

PR #5323, opened 2026-09-19
feat: exclude selected audio devices from automatic processing

chicco-carone:
This pr adds support in the settings to choose one or multiple audio sinks to skip completely when processing audio. When choosing this option for a sink pipewire will not connect it at all to the virtual easyeffects sink making it act like a regular unprocessed stream. This is useful in cases which even without applying any effects adding another sink could break audio like in surround setups that are not supported in easyeffects or with audio passthrough setups (Dolby Atmos, DTS etc). 

The user will have in the audio settings below `Process all streams` the option to open the dropdown and choose which sinks to skip
image


wwmm:
I did some tests and it seems fine. I probably will never have an use case for this but it seems useful for other people indeed. Thanks!

